PR #6, opened 2024-02-13
chore: adds gh-pages deploy support on coordinator releases

bucko13:


bucko13:
Going to merge and try and test this with a small tweak to caravan coordinator.

PR #7, opened 2024-02-13
Change coordinator base

bucko13:
With the base url now being `caravan-bitcoin` it felt a little redundant to keep the base for the coordinator app still be `caravan`. This PR changes the base path to be `coordinator/#`. This works out of the box if you try and run locally with `turbo run dev` and then can access from `http://localhost:5173/coordinator/#/`. Hopefully if the automation is working, then when this gets released, it'll trigger a deploy to https://caravan-bitcoin.github.io/coordinator/#/

I also changed a bunch of other old urls referencing the old repo and the old unchained website.

### DIFF
--- a/.changeset/thirty-camels-think.md
+++ b/.changeset/thirty-camels-think.md
@@ -1,0 +1,5 @@
+---
+"@caravan/coordinator": patch
+---
+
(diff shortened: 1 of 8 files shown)


nk1tz (approved):
All makes sense.

PR #14, opened 2024-02-19
feat(coordinator): implement changes to use @caravan/clients for network requests in coordinator

bucko13:
Re-implements [PR #365](https://github.com/unchained-capital/caravan/pull/365/files) from legacy web app repo. Uses new monorepo package instead of local module. 

Reproducing description below:

## Description
With a recent release adding mempool.space support (#347), a few issues came to light, which this PR seeks to address. 

#### Rate Limiting in mempool.space
The biggest one noticed in #364 was that recent rate limiting implemented on mempool.space's main production endpoint caused significant performance degradation. Throttling was already an issue with blockstream for which we added a delay to avoid hitting the rate limit, but this may not be enough. 

#### Hard to change client
Unfortunately the way to change the client had to come from unchained-bitcoin, which is really inconvenient and probably not the right long term place for this. 

#### Clumsy API
Every api call had to pass in the network and client info even though once the wallet is setup, this doesn't change. Would be nice to just get this from the store and be done.

#### Mempool not a perfect drop-in replacement
One of the api calls (getFeeEstimates) was not quite a drop-in b/w mempool and blockstream. So this might've been broken as well

#### Hard to make changes
If there are any changes in the future (immediate example: mempool is going to have a websocket option for using as a wallet backend), it's not easy to update the client without backwards incompatible changes. 

### Solution
- This PR first moves all the client relevant code into their own module (c90211ddee4de7fce9e7bcdcea5628cd31bccabe). 
- Then it introduces a new `BlockchainClient` class ([d3201c2](https://github.com/unchained-capital/caravan/pull/365/commits/d3201c2d1e5b3a17f0813d7e58d89282d4ac4f76)), which once initialized with the connection details shares an api no matter what client you're connecting with
- This class supports bitcoind, mempool, or blockstream. This moves the connection info from unchained-bitcoin. Longer term, I think this should exist in its own package to be maintained separately. An external package shouldn't dictate what the public client endpoint is, but it makes sense to have a unified (and well tested) package for communicating with the blockchain. The consumer (in this case the caravan coordinator) can decide which it wants to use when initialized and just go from there.
- the UI is still very tightly coupled with redux, so this PR adds some code [f591a0e](https://github.com/unchained-capital/caravan/pull/365/commits/f591a0e77d19ffedfdd2d8506be4d3a0ceba0af3) to the store and actions to set the client and get the client (based on other client settings in the store) for components to use as needed. 
- TypeScript and tests. TS isn't as useful as it could be without many TS consumers but it's a start. 
- The class also has built-in, configurable support for throttling if necessary to avoid rate limiting in a way that should be easy to turn on and off
- the update in `clientActions` shows how easy it should be to support multiple clients. It's one line now to change the default public b/w blockstream and mempool and eventually we should easily be able to support either and leave it up to the user.


apps/coordinator/src/actions/clientActions.ts
```diff
@@ -0,0 +1,50 @@
...
+import { BlockchainClient, ClientType } from "@caravan/clients";
+
+export const SET_CLIENT_TYPE = "SET_CLIENT_TYPE";
+export const SET_CLIENT_URL = "SET_CLIENT_URL";
+export const SET_CLIENT_USERNAME = "SET_CLIENT_USERNAME";
+export const SET_CLIENT_PASSWORD = "SET_CLIENT_PASSWORD";
+
+export const SET_CLIENT_URL_ERROR = "SET_CLIENT_URL_ERROR";
+export const SET_CLIENT_USERNAME_ERROR = "SET_CLIENT_USERNAME_ERROR";
+export const SET_CLIENT_PASSWORD_ERROR = "SET_CLIENT_PASSWORD_ERROR";
```
nk1tz, line 11: placeholders for future functionality?

nk1tz (dismissed):
I like it, clean.

Obv conflicts, but can re-approve after those are fixed if necessary.

dylan-thompson:
I'm seeing an error pulling fee estimates with public clients. I don't see the network request firing and this error is logged to the console:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toString')
    at OutputsForm.getFeeEstimate (OutputsForm.jsx:196:25)
    at async OutputsForm.initialOutputState (OutputsForm.jsx:225:5)
```

dylan-thompson:
Screenshot 2024-02-21 at 1 28 28 PM

bucko13:
@dylan-thompson good catch. Should be working now

dylan-thompson:
Looks like we have a similar situation with broadcast `There was an error broadcasting the transaction.: TypeError: client.broadcastTransaction is not a function`

bucko13:
@dylan-thompson thanks! Should be fixed now

bucko13:
And now it's fixed on the remote 🤦‍♂️  : [aebd85a](https://github.com/caravan-bitcoin/caravan/pull/14/commits/aebd85abfa306404ebe80cfaca6a90867872d32b)

bucko13:
> Broadcasting is now working, but I'm getting an error banner after broadcast. Returning to the wallet and refreshing the page refreshes the wallet state properly.

I think this is just due to caravan's over-reliance on the redux store as well as not persisting state very well. Basically a refresh happened (or maybe something like a hot reload) and so the client configs got reset so the client didn't know where to fetch from. I don't want to rely on defaults since this could result in inadvertently revealing user addresses to a public website. I actually wonder if this has been a problem previously because I ran into it once as well and found that the client data was just empty in the redux store so there wasn't anyway to query either way. 

All that to say that this might be something worthy of a separate and bigger fix.

dylan-thompson:
> > Broadcasting is now working, but I'm getting an error banner after broadcast. Returning to the wallet and refreshing the page refreshes the wallet state properly.
> 
> I think this is just due to caravan's over-reliance on the redux store as well as not persisting state very well. Basically a refresh happened (or maybe something like a hot reload) and so the client configs got reset so the client didn't know where to fetch from. I don't want to rely on defaults since this could result in inadvertently revealing user addresses to a public website. I actually wonder if this has been a problem previously because I ran into it once as well and found that the client data was just empty in the redux store so there wasn't anyway to query either way.
> 
> All that to say that this might be something worthy of a separate and bigger fix.

I didn't see the same error in the deployed version but I think a follow up is fine. No functionality is broken

dylan-thompson (dismissed):

bucko13:
> I didn't see the same error in the deployed version but I think a follow up is fine. No functionality is broken

Yeah, if I'm right I think this could be because of hot reloading clearing the redux store but the deployed version wouldn't have this hot reloading and the redux store is going to be persisted until refresh.

dylan-thompson (approved):

bucko13:
@dylan-thompson I found one more! [caf1ca3](https://github.com/caravan-bitcoin/caravan/pull/14/commits/caf1ca342b7370be0fce7e1288620b5a9ac3f07b)

dylan-thompson (approved):

PR #16, opened 2024-02-20
Caravan descriptors

bucko13:
Re-implements https://github.com/unchained-capital/caravan/pull/340 for the monorepo. 

~~I'm mostly happy with this as is. Only thing I'd like to tweak is the UI for importing via descriptor.~~

Total changes:
- small tweak to the generator to include prettier and other fixes
- Updates to ci workflow
- adds new @caravan/descriptor package
- ~~implements initial UIs for use in the coordinator~~ reverted in [571fd90](https://github.com/caravan-bitcoin/caravan/pull/16/commits/571fd90c5b0427fbc3cbb7dcc3d296145b0863c0). will add back in separate PR

I decided ultimately to split out the coordinator part and have @caravan/coordinator use a deployed version of the package rather than the local one. This will avoid developers having to support the rust toolchain unless they want to work on a package that needs it. 

~~I also need to update the build actions to support this toolchain so that it can support building and deploying.~~


packages/caravan-clients/src/client.test.ts
```diff
@@ -479,6 +479,7 @@ describe("BlockchainClient", () => {
       // Verify the returned result
       expect(result.utxos).toEqual(
         await Promise.all(
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
nk1tz, line 482: Can you not use your newly exported `UTXO` interface from `packages/caravan-clients/src/client.ts`?

packages/caravan-descriptors/caravan-rs/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+
+// #[wasm_bindgen]
+// extern "C" {
+//     fn alert(s: &str);
+// }
+
+// #[wasm_bindgen]
+// pub fn init() {
+//     alert("Initialized from Rust!");
+// }
```
nk1tz, line 40: keep or clean?

nk1tz (commented):
AFAICT tests are thorough.

Minor comments

packages/caravan-clients/src/client.test.ts
bucko13: I tried that and forget why I reversed it. It might’ve just been I was trying to clean up other lint issues in parallel and the scope was increasing. I’ll try again here to at least make this unnecessary. Thanks!

packages/caravan-descriptors/caravan-rs/src/lib.rs
bucko13: Oh yeah. Good catch. My adventures in rust and wasm documented 😅

bucko13: [994068e](https://github.com/caravan-bitcoin/caravan/pull/16/commits/994068e300a2404a9f8fd169b78d2973606af278)

packages/caravan-clients/src/client.test.ts
bucko13: [994068e](https://github.com/caravan-bitcoin/caravan/pull/16/commits/994068e300a2404a9f8fd169b78d2973606af278)

packages/caravan-descriptors/caravan-rs/src/config.rs
```diff
@@ -0,0 +1,1176 @@
...
+        .map_err(Error::Descriptor);
+
+        match result {
+            Ok((d, _, n)) => {
+                if n.contains(&network) {
+                    Ok(d)
+                } else {
+                    Err(Error::InvalidNetwork {
+                        requested: network,
+                        found: *n.iter().last().expect("network"),
```
tylerchambers, line 188: can this panic if there are no elements in `n`? do we care?

packages/caravan-descriptors/caravan-rs/src/config.rs
```diff
@@ -0,0 +1,1176 @@
...
+            network,
+            client,
+            quorum: external_quorum,
+            extended_public_keys,
+            starting_address_index: 0,
+        })
+    }
+
+    pub fn to_string_pretty(&self) -> String {
+        serde_json::to_string_pretty(self).unwrap()
```
tylerchambers, line 294: This assumes successful serialization. We might be OK with that.

packages/caravan-descriptors/caravan-rs/src/config.rs
```diff
@@ -0,0 +1,1176 @@
...
+}
+
+fn parse_key(pubkey: &DescriptorPublicKey) -> Result<CaravanExtendedPublicKey, Error> {
+    match pubkey {
+        DescriptorPublicKey::Single(_) => {
+            Err(Error::Generic("Unsupported single pub key.".to_string()))
+        }
+        DescriptorPublicKey::XPub(xpub) => {
+            let mut xfp = None;
+            let mut bip32_path = None;
```
tylerchambers, line 353: consider casting these runes instead
```rs
let xfp = xpub.origin.as_ref().map(|(s_xfp, _)| *s_xfp);
let bip32_path = xpub.origin.as_ref().map(|(_, s_bip32_path)| s_bip32_path.clone());
```

tylerchambers (approved):
looks good to me, nothing blocking just suggestions

packages/caravan-descriptors/caravan-rs/src/config.rs
bucko13: assuming `n` is network, I think this can be empty for the config. A descriptor doesn't require it/use it for example.

PR #19, opened 2024-02-23
add vercel specific deployment action

bucko13:


bucko13:
Going to merge as there are only build related changes here that get things working for deploys.

PR #29, opened 2024-02-28
feat: add latest tag for main branch builds

robertshuford:
1. Changes secret from `DOCKER_PASSWORD` to `DOCKER_TOKEN` because its actually an access token.
2. attempt to add the latest tag to `main` branch builds.


bucko13 (approved):
SEND IT!

PR #36, opened 2024-03-04
feat: amd64 and arm64 images

robertshuford:
Create amd64 and arm64 multiarch image.
Only create new images when coordinator tags are created.


bucko13 (approved):
Nice.

PR #39, opened 2024-03-06
fix: public client broadcast call

bucko13:
Should resolve an issue where transactions couldn't broadcast with mempool.space.

Unclear how the api request body got twisted like this. Before the latest change to these lines, it was the same for both clients but in the way mempool.space was before this change. So they are now both the same again, but in the way fixed previously for blockstream (I did have a memory of actually fixing this before and I guess that's why).


dylan-thompson (approved):
Broadcast is working with mempool and blockstream

PR #41, opened 2024-03-06
fix: TrezorConnect default import for some build environments

bucko13:


packages/caravan-wallets/src/trezor.ts
```diff
@@ -111,6 +129,7 @@ const TREZOR_DEV =
   env_variables.VITE_TREZOR_DEV;
 
 try {
+  console.log("TrezorConnect:", TrezorConnect);
```
sashafklein, line 132: Remove

packages/caravan-wallets/src/trezor.ts
```diff
@@ -804,6 +823,7 @@ export class TrezorSignMultisigTransaction extends TrezorInteraction {
   }
 }
 
+// eslint-disable-next-line multiline-comment-style
```
sashafklein, line 826: Seems these changes (this and the two commented imports) can be removed

sashafklein (changes requested):

sashafklein (dismissed):

packages/caravan-wallets/src/trezor.ts
bucko13: [8156e72](https://github.com/caravan-bitcoin/caravan/pull/41/commits/8156e7202c350205b2d5ed6b379054f26b45c60d)

packages/caravan-wallets/src/trezor.ts
bucko13: Unsure why eslint called just this comment out. Didn't really want to dig into it and potentially muck up the diff even more. Maybe something about spacing?

bucko13: cleaned up a superfluous rule: [bbc5543](https://github.com/caravan-bitcoin/caravan/pull/41/commits/bbc554319ff886c0f1946b1cb5ed9213c2dfe0f5)

PR #44, opened 2024-03-07
fix: descope coordinator app

robertshuford:
Caravan coordinator is an app, not an npm package, so the @scope/package naming pattern isn't needed.  This will enable us to quickly identify on the https://github.com/caravan-bitcoin/caravan/releases which release is a package release, and which is an app release.


bucko13 (approved):
Good call. Thanks for fixing!

PR #50, opened 2024-03-14
Use changeset to version and tag `coordinator`

robertshuford:
Use changeset to version and tag `coordinator` and outlined on:
https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md


bucko13 (approved):
Good find! Not sure how I missed that. Thanks for updating. I was going crazy with this.

PR #51, opened 2024-03-14
Update version in package-lock on release

robertshuford:
`changeset version` updates package.json, but not package-lock.json.

Changeset has an open issue for this, https://github.com/changesets/action/issues/203.  

A suggested workaround is to use `npm install`, https://github.com/changesets/action/issues/203#issuecomment-1460115073


bucko13:
Interesting 
https://github.com/changesets/changesets/issues/1139

Issue #53, opened 2024-03-31
Can't set private node


jbrauck-unchained:
It looks like no matter what you type in for your node's URL, it is still making the request to http:localhost:8332.
See two screenshots comparing old caravan to the new
<img width="1455" alt="Screenshot 2024-03-31 at 12 06 41 PM" src="https://github.com/caravan-bitcoin/caravan/assets/102981364/4c963cd3-90a8-4c2c-8b4f-32785e9a5ed2">
<img width="1409" alt="Screenshot 2024-03-31 at 12 05 56 PM" src="https://github.com/caravan-bitcoin/caravan/assets/102981364/ee8d0e4c-4988-4227-8f48-2c3c2467ee50">

jbrauck-unchained:
The above has been tested and merged. Closing this issue now.

PR #54, opened 2024-04-01
Fix client redux

jbrauck-unchained:
In working to get caravan running on my umbrel I discovered a bug with the redux being used on the client information section.
Fixes issue: https://github.com/caravan-bitcoin/caravan/issues/53

- I removed the outer async function in getBlockchainClientFromStore.
- Additionally, I changed `value` to `payload` in the dispatched action to follow the common convention in Redux.
- ~I imported `Dispatch` from redux rather than react~ I un-did this change as it ended up not being a part of the issues and actually broke one of the github workflows.
- Lastly, I removed some outdated comments bc we already have new clients --like mempool.space-- added 😎


bucko13 (approved):
Confirmed it works. Good catch on these (bugs which were introduced by yours truly 😓)

PR #57, opened 2024-04-01
refactor: split types among module categories creating room to grow

Shadouts:
This change should affect functionality in no way. This is purely a code organization refactor. PsbtV2 is given its own module directory to collect its related types into several categories by file:
- functions
- class psbtv2
- abstract psbtv2maps
- types
- values

### DIFF
--- a/packages/caravan-psbt/src/psbtv2/functions.ts
+++ b/packages/caravan-psbt/src/psbtv2/functions.ts
@@ -1,0 +1,169 @@
+import { BufferReader, BufferWriter } from "bufio";
+import {
+  validateHex,
+  validBase64,
+  validateBIP32Path,
+  PSBT_MAGIC_BYTES,
+} from "@caravan/bitcoin";
+
+import { Key, Value, NonUniqueKeyTypeValue, KeyType } from "./types";
+import {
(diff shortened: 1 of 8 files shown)


packages/caravan-psbt/src/psbtv2/functions.ts
bucko13, line 1: Should some of these be generalized utilities not just for psbtv2?

Shadouts: I'm not opposed to a change which would consider each function independently. It's important to recognize that the types here, while having generic sounding names, are specifically shaped for functionality within the psbtv2 module. I think `parseDerivationPathNodesToBytes` is the easiest for me to suggest as a candidate which is truly a generic utility, but iirc another version of that function exists somewhere else (maybe in caravan-bitcoin) and doesn't work the same way.

Even if genericizing some of them would be a benefit, I think it should be saved for another change PR.

bucko13: Yeah, that one and `bufferize` stuck out. Later is fine. Wouldn't change the API or anything anyway.

bucko13 (approved):

PR #73, opened 2024-04-17
PsbtV2.setProprietaryValue sets 0xFC keytypes for all maps

Shadouts:
This change adds `PsbtV2.setProprietaryValue()` which can set proprietary values (keytype `0xFC`) for all maps. Map selector and keydata values are specified in the method args, and the final key is formatted as `0xFC<compact uint identifier length><bytes identifier><bytes subkeytype><bytes subkeydata>`


packages/caravan-psbt/README.md
```diff
@@ -217,6 +218,20 @@ The Signer, when it creates a signature, must add the partial sig keypair to the
 
 Removes all sigs for an input unless a pubkey is specified. Validates that the input exists. When providing a pubkey, this validates that a sig for the pubkey exists.
 
+##### `public setProprietaryValue`
```
bucko13, line 221: Low priority todo list: it would be nice if we autogenerated docs

bucko13 (changes requested):
Looks good! Just needs a changeset

Shadouts:
Changeset added in 33f7ac20c0983dbacf580d775091534ebea770ec

bucko13 (approved):

PR #76, opened 2024-04-25
Update to client interactions to support wallet api with bitcoind

### LINKED ISSUE #75: Better compatibility with newer versions of bitcoind
The current bitcoind client api assumes an older rpc interface and bitcoind wallet infrastructure. Since the introduction of wallet descriptors there have been changes that cause issues with newer versions of bitcoind. 

Now that bitcoind nodes are deprecating old watch-only wallets that support importing addresses, a legacy wallet is needed with the current caravan clients api. Furthermore, it is likely that a special wallet that supports watch-only will be required to import descriptors to a wallet without private keys.

Some features that would be useful for caravan/clients and the coordinator to support:
- Accepting and saving a wallet name argument in the wallet config for private/bitcoind commands
- Passing the `-rpcwallet` argument to bitcoind client when available
- Support for creating and loading a named, watch-only wallet if it doesn't exist (Maybe? Not sure if this is desirable or not)
- Add command to import descriptors to bitcoind rather than addresses one at a time.

bucko13:
Several updates here:
* @caravan/clients
    * Adds new bitcoind client for making requests to _named_ wallets (which have a different url they make calls to) 
    * Adds new methods to blockchain client for importing descriptors (private client type only)
* Caravan Coordinator:
    * Typescript-ify some files
    * New optional field in private client config for a wallet name
    * Can import descriptors to bitcoind for syncing addresses
    * Other misc cleanup to how the blockchain client is accessed and passed around the app

TODO:
- [x] Review code for any additional cleanup
- [x] Add tests for client update

Fixes #75


apps/coordinator/src/components/Wallet/WalletGenerator.jsx
```diff
@@ -264,7 +264,11 @@ class WalletGenerator extends React.Component {
     try {
       const { getBlockchainClient } = this.props;
       const client = await getBlockchainClient();
-      await client.getFeeEstimate();
+      if (client.bitcoindParams.walletName) {
+        await client.getWalletInfo();
```
bucko13, line 268: If we have a wallet name then this is the way we need to check if the connection details are correct. Otherwise we fallback to checking fee estimate.

apps/coordinator/src/components/Wallet/WalletGenerator.jsx
```diff
@@ -278,7 +282,7 @@ class WalletGenerator extends React.Component {
         setPasswordError(
           "Unauthorized: Incorrect username and password combination",
         );
-      else setPasswordError(e.message);
+      else setPasswordError(e.response?.data?.error?.message || e.message);
```
bucko13, line 285: Allows us to pass error messages from bitcoind if that's the source.

apps/coordinator/src/hooks/client.ts
bucko13, line 1: the new hook that will set and return the blockchain client anytime any relevant settings are changed.

apps/coordinator/src/hooks/descriptors.ts
bucko13, line 1: Used in a couple components so extracted existing code into a reusable hook.

apps/coordinator/src/reducers/index.ts
```diff
@@ -53,3 +53,4 @@ const rootReducer = (state, action) => {
 };
 
 export default rootReducer;
+export type RootState = ReturnType<typeof rootReducer>;
```
bucko13, line 56: allows for typed reducers and action creators.

apps/coordinator/src/selectors/wallet.js
```diff
@@ -24,7 +24,8 @@ const getClientDetails = (state) => {
     return `{
     "type": "private",
     "url": "${state.client.url}",
-    "username": "${state.client.username}"
+    "username": "${state.client.username}",
+    "walletName": "${state.client.walletName}"
```
bucko13, line 28: adds wallet name to exported configs.

packages/caravan-clients/src/blockchain.js
```diff
@@ -13,7 +13,7 @@ import {
   bitcoindParams,
   bitcoindGetAddressStatus,
   isWalletAddressNotFoundError,
-} from "./bitcoind";
+} from "./wallet";
```
bucko13, line 16: new wallet-enabled client functions that support named wallets if available.

bucko13 (commented):
Leaving some annotations of the updates

dylan-thompson:
Looks good, nice work!

PR #77, opened 2024-04-27
fix: ensure safe handling of interaction.Messages to prevent `map is not a function` error

### LINKED ISSUE #72: Bug: Error when selecting Ledger or Trezor in address generator
**Current Behavior**
When we go to address generator and select method for public key as Ledger or Trezor error shows up
Screenshot from 2024-04-16 20-19-41

**Steps to Reproduce (for bugs)**
Go to Address
Select Ledger or Trezor as public key method
Note you will see error

**Environment**
Where are you running caravan: localhost
Operating system: Linux
Browser and version: Chrome

MashyBasker:
The error message `TypeError: interaction.messagesFor is not a function` suggests that `messagesFor` is not a function which can be called on the `interaction` object. 
To fix this we call `messagesFor` on the result of invoking `interaction` as a function.

### Output after the change
image

Fixes #72

### DIFF
--- a/apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
+++ b/apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
@@ -159,10 +159,9 @@ const HardwareWalletPublicKeyImporter = ({
             Import Public Key
           </Button>
         </Box>
-        <InteractionMessages
-          messages={interaction.messagesFor({ state: status })}
-          excludeCodes={["bip32"]}
-        />
+        {Array.isArray(InteractionMessages) && InteractionMessages.map(message => (
+          <div key={message.id}>{message.text}</div>
+        ))}
         <FormHelperText error>{publicKeyError}</FormHelperText>
       </Box>
     );


apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
```diff
@@ -159,10 +159,9 @@ const HardwareWalletPublicKeyImporter = ({
             Import Public Key
           </Button>
         </Box>
-        <InteractionMessages
-          messages={interaction.messagesFor({ state: status })}
-          excludeCodes={["bip32"]}
-        />
+        {Array.isArray(InteractionMessages) && InteractionMessages.map(message => (
```
bucko13, line 162: Does this actually work to end up showing anything? I have to test it out still but looking at the code, `InteractionMessages` is a component exported from `components/InteractionMessages` which I would assume means it's never going to be an array

bucko13:
Unless I'm missing something about how this fix was implemented, I believe the correct fix should look like this: https://github.com/caravan-bitcoin/caravan/pull/81

apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
MashyBasker: I was trying to fix this and my thought process was since everything is an object in JS, I can check it via an `isArray` check. The second part of the code comes from trial and error and coincidentally it fixed the bug on my end.
I now see how this is a bad idea and makes the code a lot harder to read. Closing this PR now.

PR #79, opened 2024-04-28
Fix error with odd length hex check

benthecarman:


bucko13:
Nice. Good catch. Can you add a [changeset](https://github.com/caravan-bitcoin/caravan/pull/79#issuecomment-2081280444) for a patch?

benthecarman:
added

bucko13 (approved):

PR #81, opened 2024-04-28
Fix error on address generator

### LINKED ISSUE #72: Bug: Error when selecting Ledger or Trezor in address generator

bucko13:
fix interactions call in HardwareWalletPublicKeyImporter.

Fixes #72

### DIFF
--- a/apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
+++ b/apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
@@ -160,7 +160,7 @@ const HardwareWalletPublicKeyImporter = ({
           </Button>
         </Box>
         <InteractionMessages
-          messages={interaction.messagesFor({ state: status })}
+          messages={interaction().messagesFor({ state: status })}
           excludeCodes={["bip32"]}
         />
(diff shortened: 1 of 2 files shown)


apps/coordinator/src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
```diff
@@ -160,7 +160,7 @@ const HardwareWalletPublicKeyImporter = ({
           </Button>
         </Box>
         <InteractionMessages
-          messages={interaction.messagesFor({ state: status })}
+          messages={interaction().messagesFor({ state: status })}
```
benthecarman, line 163: javascript was a mistake lmao

bucko13: You're right. Typescript is the solution! 🎉 

Getting there in caravan.

PR #88, opened 2024-05-02
Fix rescan

jbrauck-unchained:
With new bitcoindImportDescriptors functionality, rescan must be passed to the node via the timestamp value. See https://developer.bitcoin.org/reference/rpc/importdescriptors.html#argument-1-requests.

So now when rescan is set to true on an import addresses request, the timestamp value sent to the node is set to "0" prompting a rescan on that wallet.


bucko13 (approved):
Good stuff! Confirmed that a rescan was triggered on a local regtest setup.

bucko13 (approved):

PR #92, opened 2024-05-05
fix: docker publish if changeset published.

robertshuford:
Tags pushed from a GitHub action will not catch the `on.push.tags` trigger.
To fix this, publish to docker hub based on the output of the release step.


bucko13 (approved):
🤞

PR #98, opened 2024-06-10
fix: some build cleanup to avoid dependency

bucko13:
when trying to install one of the new packages (psbt or wallets) that depends on the internal package `@caravan/multisig`, I'm getting an error that the package can't be found in the registry. This shouldn't even be attempted. I'm not sure that this will fix it, but when looking at some turborepo examples, I did see that [internal packages](https://github.com/vercel/turbo/blob/main/examples/kitchen-sink/packages/logger/package.json) seem to have a version of `0.0.0` so maybe that will help. Included some other cleanup in the build processes that may or may not be related.

### DIFF
--- a/.changeset/funny-gorillas-destroy.md
+++ b/.changeset/funny-gorillas-destroy.md
@@ -1,0 +1,6 @@
+---
+"@caravan/psbt": patch
+"@caravan/wallets": patch
+---
+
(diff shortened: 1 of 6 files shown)


bucko13:
@robertshuford includes the changes from #99 as well

robertshuford (approved):

PR #104, opened 2024-06-11
upgrade more legacy psbt functions for s2tr

bucko13:
a couple other functions that needed to be using bitcoinjs-lib v6 for taproot output support. previously this would cause an error in trezor signing if the interaction had a psbt and key details, then the old translatePSBT would result in an output without an address. 

In caravan coordinator there should be no change in functionality (i added an upgrade so that it would deploy and reflect the other s2tr changes in the prior PR on this subject), but it can be verified by changing the DirectSignatureImporter to pass a PSBT and keyDetails to the signing interaction. This will cause the TrezorSignMultisigTransaction to hit the code path that calls translatePSBT. With the old translatePSBT from `@caravan/bitcoin`, trezor-connect throws an error.

### DIFF
--- a/packages/caravan-psbt/package.json
+++ b/packages/caravan-psbt/package.json
@@ -51,8 +51,8 @@
   "author": "unchained capital",
   "license": "ISC",
   "devDependencies": {
-    "@caravan/multisig": "*",
     "@caravan/eslint-config": "*",
+    "@caravan/multisig": "*",
     "@caravan/typescript-config": "*",
     "@inrupt/jest-jsdom-polyfills": "^3.2.1",
     "@jest/globals": "^29.7.0",
@@ -61,6 +61,7 @@
     "jest": "^29.4.1",
     "jsdom": "24.0.0",
     "jsdom-global": "3.0.2",
+    "lodash": "^4.17.21",
     "react-silence": "^1.0.4",
     "ts-jest": "^29.1.2",
     "ts-node": "^10.9.2",
--- a/packages/caravan-psbt/src/psbtv0/psbt.ts
+++ b/packages/caravan-psbt/src/psbtv0/psbt.ts
@@ -1,6 +1,7 @@
 import {
   ExtendedPublicKey,
   fingerprintToFixedLengthHex,
+  generateMultisigFromHex,
   isValidSignature,
   MultisigAddressType,
   multisigSignatureBuffer,
@@ -20,6 +21,8 @@ import * as ecc from "../../vendor/tiny-secp256k1-asmjs/lib/index.js";
 import * as bitcoin from "bitcoinjs-lib-v6";
 import { bufferize } from "src/functions";
 import BigNumber from "bignumber.js";
+import { PSBT_MAGIC_B64, PSBT_MAGIC_HEX } from "../constants";
+import { reverseBuffer } from "bitcoinjs-lib-v6/src/bufferutils.js";
 
 bitcoin.initEccLib(ecc);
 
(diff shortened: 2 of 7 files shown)


packages/caravan-psbt/package.json
```diff
@@ -61,6 +61,7 @@
     "jest": "^29.4.1",
     "jsdom": "24.0.0",
     "jsdom-global": "3.0.2",
+    "lodash": "^4.17.21",
```
Shadouts, line 64: I prefer to avoid the burden of lodash, but I suppose it's okay since it's just a dev dependency. It' might be preferable to define a `cloneDeep` function instead of bringing in this lib.

packages/caravan-psbt/src/psbtv0/psbt.ts
```diff
@@ -265,3 +268,133 @@ const getHashForSignature = (
...
 };
+
+/**
+ * Given a string, try to create a Psbt object based on MAGIC (hex or Base64)
+ */
+export function autoLoadPSBT(psbtFromFile, options?: any) {
+  if (typeof psbtFromFile !== "string") return null;
+  // Auto-detect and decode Base64 and Hex.
+  if (psbtFromFile.substring(0, 10) === PSBT_MAGIC_HEX) {
+    return Psbt.fromHex(psbtFromFile, options);
```
Shadouts, line 279: This is being done different than psbtv2 and checking only stringified magic bytes isn't good enough. The entire string must be checked for validity of its encoding and then check magic bytes as bytes. Why not try to unify? Below is sort-of how psbtv2 is doing it.

```
const psbtBuf = bufferize(psbtFromFile); // Validates hex or base64 and bufferizes string
const br = new BufferReader(buf);
if (!br.readBytes(PSBT_MAGIC_BYTES.length, true).equals(PSBT_MAGIC_BYTES)) {
  throw Error("Psbt magic bytes are incorrect.");
}

...

```

validateHex and validateBase64 are two functions which will tell you what you have. I like checking for encoding and then checking for magic bytes separately.

Shadouts, line 279: This is being done different than psbtv2 and checking only stringified magic bytes isn't good enough. The entire string must be checked for validity of its encoding and then check magic bytes as bytes. Why not try to unify? Below is sort-of how psbtv2 is doing it.

```
const psbtBuf = bufferize(psbtFromFile); // Validates hex or base64 and bufferizes string
const br = new BufferReader(buf);
if (!br.readBytes(PSBT_MAGIC_BYTES.length, true).equals(PSBT_MAGIC_BYTES)) {
  throw Error("Psbt magic bytes are incorrect.");
}

...

```

validateHex and validateBase64 are two functions which will tell you what you have. I like checking for encoding and then checking for magic bytes separately.

Shadouts (approved):
Some nits, but nothing holding it back.

packages/caravan-psbt/package.json
bucko13: Yeah, I wrestled with the same and tried to use alternatives. Apparently there's a native option in JS now: [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone). Unfortunately Jest doesn't support it.

packages/caravan-psbt/src/psbtv0/psbt.ts
bucko13: Ok, let me try and rewrite that a bit. I do prefer to unify but just copied this over from what we had in caravan/bitcoin.

bucko13: Fixed: [56ebc15](https://github.com/caravan-bitcoin/caravan/pull/104/commits/56ebc1572e677bb38a98998d85384bba44efd6df)

Shadouts (approved):

PR #108, opened 2024-06-13
specify direct import

bucko13:
Unclear to me why this wasn't picked up in the CI with #104, but this should fix an error when caravan/wallets or caravan/psbt are installed externally from the monorepo


dylan-thompson (approved):
Did a quick spot check against the vercel preview and everything looks good

PR #117, opened 2024-08-16
add support for the BitBox02 hardware wallet

benma:
Using the `bitbox-api` NPM package, which loads a WASM module.

Note about CJS:
Currently the `bitbox-api` package is an ESM package with a `module: ...` entrypoint, so it is not compatible with the `cjs` target of caravan-wallets. The only workaround that I could find where compilation succeeds and the package works in the browser is to mark bitbox-api as external in tsup.config.ts.

Note about signing tests:

- The BitBox02 requires the previous transaction of each input to be
present in the PSBT (`PSBT_IN_NON_WITNESS_UTXO`), so it can verify the
input amount and avoid fee attacks. The signing test fixtures are
missing these, so they fail.

- The BitBox02 uses the anti-klepto (anti-exfil) protocol to mitigate
covert nonce exfil attacks. This results in random signatures. The
unit test fixtures hardcode the expected signatures, assuming they are
always the same. As a result, also here the tests fail. To fix this,
the tests should rather verify the ECDSA signatures against the
transaction sighash for each input.

### DIFF
--- a/apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
+++ b/apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
@@ -7,6 +7,7 @@ import {
   convertExtendedPublicKey,
   validateExtendedPublicKey,
   Network,
+  P2SH,
 } from "@caravan/bitcoin";
 import { BITBOX, TREZOR, LEDGER, HERMIT, COLDCARD } from "@caravan/wallets";
 import {
@@ -68,7 +69,7 @@ class ExtendedPublicKeyImporter extends React.Component {
   };
 
   renderImport = () => {
-    const { extendedPublicKeyImporter, number } = this.props;
+    const { extendedPublicKeyImporter, number, addressType } = this.props;
     const { disableChangeMethod } = this.state;
     return (
       <div>
@@ -82,7 +83,7 @@ class ExtendedPublicKeyImporter extends React.Component {
             variant="standard"
             onChange={this.handleMethodChange}
           >
-            <MenuItem value={BITBOX}>BitBox</MenuItem>
+            {addressType != P2SH && <MenuItem value={BITBOX}>BitBox</MenuItem>}
             <MenuItem value={TREZOR}>Trezor</MenuItem>
             <MenuItem value={COLDCARD}>Coldcard</MenuItem>
             <MenuItem value={LEDGER}>Ledger</MenuItem>
--- a/apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
+++ b/apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
@@ -8,7 +8,7 @@ import {
   validateExtendedPublicKey,
   Network,
 } from "@caravan/bitcoin";
-import { TREZOR, LEDGER, HERMIT, COLDCARD } from "@caravan/wallets";
+import { BITBOX, TREZOR, LEDGER, HERMIT, COLDCARD } from "@caravan/wallets";
 import {
   Card,
   CardHeader,
@@ -82,6 +82,7 @@ class ExtendedPublicKeyImporter extends React.Component {
             variant="standard"
             onChange={this.handleMethodChange}
           >
+            <MenuItem value={BITBOX}>BitBox</MenuItem>
             <MenuItem value={TREZOR}>Trezor</MenuItem>
             <MenuItem value={COLDCARD}>Coldcard</MenuItem>
             <MenuItem value={LEDGER}>Ledger</MenuItem>
@@ -105,7 +106,7 @@ class ExtendedPublicKeyImporter extends React.Component {
     } = this.props;
     const { method } = extendedPublicKeyImporter;
 
-    if (method === TREZOR || method === LEDGER) {
+    if (method === BITBOX || method === TREZOR || method === LEDGER) {
       return (
         <DirectExtendedPublicKeyImporter
           extendedPublicKeyImporter={extendedPublicKeyImporter}
--- a/packages/caravan-wallets/src/bitbox.ts
+++ b/packages/caravan-wallets/src/bitbox.ts
@@ -36,6 +36,9 @@ import {
  */
 export const BITBOX = "bitbox";
 
+// Callback for showing the BitBox pairing code. It returns a function to hide it again.
+export type TShowPairingCode = (pairingCode: string) => (() => void) | null;
+
 function convertNetwork(network: BitcoinNetwork): BtcCoin {
   return network === 'mainnet' ? 'btc' : 'tbtc';
 }
@@ -103,11 +106,20 @@ async function convertMultisig(pairedBitBox: PairedBitBox, walletConfig: Multisi
  * const result = await interaction.run();
  * console.log(result); // whatever value `app.doSomething(...)` returns
  *
+ * The `showPairingCode` callback can be supplied to display and hide the BitBox pairing code.
+ * If not provided, the default is to open a browser popup showing the pairing code.
  */
 export class BitBoxInteraction extends DirectKeystoreInteraction {
   appVersion?: string;
 
   appName?: string;
+
+  showPairingCode?: TShowPairingCode;
+
+  constructor({ showPairingCode }: { showPairingCode?: TShowPairingCode }) {
+    super();
+    this.showPairingCode = showPairingCode;
+  }
 
   /**
    * Adds `pending` messages at the `info` level about ensuring the
@@ -132,19 +144,60 @@ export class BitBoxInteraction extends DirectKeystoreInteraction {
     return messages;
   }
 
+  showPairingCodePopup(pairingCode: string): (() => void) | null {
+    if (this.showPairingCode) {
+      return this.showPairingCode(pairingCode);
+    }
+    const htmlContent = `
+<!doctype html>
+<html>
+  <head>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta charset="utf-8">
+    <title>BitBox02 pairing</title>
+  </head>
+  <body>
+    <h1 style="text-align:center;">BitBox02 pairing code</h1>
+    <div id="code" style="white-space:pre;font-family:monospace,monospace;text-align:center;font-size:large;">${pairingCode}</div>
+  </body>
+</html>
+`;
+    const popup = window.open(
+      '',
+      'popupWindow',
+      'width=400,height=300',
+    );
+    if (popup) {
+      popup.document.write(htmlContent);
+      popup.document.close();
+      return () => {
+        popup.close();
+      };
+    }
+    return null;
+  }
+
   async withDevice<T>(f: (device: PairedBitBox) => Promise<T>): Promise<T> {
     const bitbox = await import('bitbox-api');
 
+    let hidePairingCode: (() => void) | null = null;
     try {
       const unpaired = await bitbox.bitbox02ConnectAuto(() => {
+        if (hidePairingCode) {
+          hidePairingCode();
+        }
       })
       const pairing = await unpaired.unlockAndPair()
       const pairingCode = pairing.getPairingCode()
       if (pairingCode) {
+        hidePairingCode = this.showPairingCodePopup(pairingCode);
         // TODO: display pairing code to the user.
         console.log('Pairing code:', pairingCode);
       }
       const pairedBitBox = await pairing.waitConfirm()
+      if (hidePairingCode) {
+        hidePairingCode();
+      }
       try {
         return await f(pairedBitBox)
       } finally {
@@ -156,7 +209,9 @@ export class BitBoxInteraction extends DirectKeystoreInteraction {
       const errorMessage = isErrorUnknown ? typedErr.err! : typedErr.message
       throw new Error(errorMessage);
     } finally {
-      // TODO hide pairing code if still shown.
+      if (hidePairingCode) {
+        hidePairingCode();
+      }
     }
   }
 
@@ -209,6 +264,10 @@ export class BitBoxInteraction extends DirectKeystoreInteraction {
  *
  */
 export class BitBoxGetMetadata extends BitBoxInteraction {
+  constructor({ showPairingCode }: { showPairingCode?: TShowPairingCode }) {
+    super({ showPairingCode });
+  }
+
   async run() {
     return this.withDevice(async (pairedBitBox) => {
       const product = pairedBitBox.product();
@@ -244,8 +303,13 @@ export class BitBoxExportPublicKey extends BitBoxInteraction {
    * @param {string} bip32Path path
    * @param {boolean} includeXFP - return xpub with root fingerprint concatenated
    */
-  constructor({ network, bip32Path, includeXFP }) {
-    super();
+  constructor({ showPairingCode, network, bip32Path, includeXFP }: {
+    showPairingCode?: TShowPairingCode;
+    network: BitcoinNetwork;
+    bip32Path: string;
+    includeXFP: boolean;
+  }) {
+    super({ showPairingCode });
     this.network = network;
     this.bip32Path = bip32Path;
     this.includeXFP = includeXFP;
@@ -290,8 +354,13 @@ export class BitBoxExportExtendedPublicKey extends BitBoxInteraction {
    * @param {string} network bitcoin network
    * @param {boolean} includeXFP - return xpub with root fingerprint concatenated
    */
-  constructor({ bip32Path, network, includeXFP }) {
-    super();
+  constructor({ showPairingCode, bip32Path, network, includeXFP }: {
+    showPairingCode?: TShowPairingCode;
+    network: BitcoinNetwork;
+    bip32Path: string;
+    includeXFP: boolean;
+  }) {
+    super({ showPairingCode });
     this.bip32Path = bip32Path;
     this.network = network;
     this.includeXFP = includeXFP;
@@ -329,9 +398,13 @@ export class BitBoxRegisterWalletPolicy extends BitBoxInteraction {
   walletConfig: MultisigWalletConfig;
 
   constructor({
+    showPairingCode,
     walletConfig
-  }: { walletConfig: MultisigWalletConfig}) {
-    super();
+  }: {
+    showPairingCode?: TShowPairingCode;
+    walletConfig: MultisigWalletConfig;
+  }) {
+    super({ showPairingCode });
     this.walletConfig = walletConfig;
   }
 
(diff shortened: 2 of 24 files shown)


packages/caravan-wallets/src/index.ts
```diff
@@ -557,6 +601,8 @@ export function ConfigAdapter({
   policyHmac?: string;
 }) {
   switch (KEYSTORE) {
+    case BITBOX:
+      throw new Error('TODO');
```
benma, line 605: Not really sure what to do here or if this is needed at all for BitBox.

packages/caravan-wallets/src/bitbox.ts
```diff
@@ -0,0 +1,467 @@
...
+    const bitbox = await import('bitbox-api');
+
+    try {
+      const unpaired = await bitbox.bitbox02ConnectAuto(() => {
+      })
+      const pairing = await unpaired.unlockAndPair()
+      const pairingCode = pairing.getPairingCode()
+      if (pairingCode) {
+        // TODO: display pairing code to the user.
+        console.log('Pairing code:', pairingCode);
```
benma, line 145: @bucko13 the BitBox02 asks the user to confirm a pairing code to establish the encrypted communication channel. It is known only when running a BitBox02 interaction, at this line. Any suggestions for how to go about showing the code to the user in caravan? Could the messages be used somehow?

packages/caravan-wallets/src/index.ts
bucko13: This is an interaction basically to do wallet registrations (but an older API). You already have that interaction covered so I think you can just remove this, or add to both to duplicate the support.

packages/caravan-wallets/src/bitbox.ts
bucko13: My first instinct is to use the `messages` api as you suggest. It's honestly not a part of the system I'm super familiar with so I'm not sure if you can run through them asynchronously like that, but I think you can. 

Maybe the way to do it is set up particular message type for pairing. Push the new message on when it's ready. At this point when you have the code you would add it to the messages queue with the appropriate code and type. The consumer of the library would have to know the API and make sure to make use of `messagesFor` correctly and wait for the expected message to be made available.

bucko13:
Note for when you're ready to take it out of draft that this will [need](https://github.com/caravan-bitcoin/caravan/pull/117#issuecomment-2294382271) a changeset in order to bump versions. Without having reviewed yet, I'm guessing minor version bump for @caravan/wallets.

benma:
> Note for when you're ready to take it out of draft that this will [need](https://github.com/caravan-bitcoin/caravan/pull/117#issuecomment-2294382271) a changeset in order to bump versions. Without having reviewed yet, I'm guessing minor version bump for @caravan/wallets.

Added the generated changeset file. Should I also bump the package.json version numbers manually?

packages/caravan-wallets/src/bitbox.ts
benma: I am not sure how to do that and trigger a re-render of the messages when the pairing code is available and hide it again when it is confirmed. Would it be possible someone on your team who is familiar with this to code the solution in the end?

In the meantime, this PR is still functional and could be merged, just the pairing code would be shown and confirmed on the BitBox02 while the host does not show it. This could be added later by your team in a way you think is proper.

packages/caravan-wallets/src/index.ts
benma: Added support in the same way as Ledger a few lines below.

benma:
Ready for review :smile:

packages/caravan-wallets/src/bitbox.ts
bucko13: If you have the interaction push the messages on the instance array (with the appropriate codes, messages, and tests), that should be sufficient. It can be up to the consumer of the library how they want to track that and update the UI. For a React app, if you track an object via state or properties for example, if the object changes then that should trigger a re-render (or with a useEffect you can trigger some other action to occur). But those are implementation details we can play with later.

bucko13:
> Added the generated changeset file. Should I also bump the package.json version numbers manually?


@benma nope! The automation will take care of that

packages/caravan-wallets/src/bitbox.ts
benma: It seems `messages()` is a function in the interaction, not an instance property/array. How would I trigger clients to call it when the pairing code becomes available? It seems this is not currently supported. 

I could add a "pairingCallback" param, but it would need to be passed to every interaction, since every interaction with the device could potentially ask to pair.

apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
```diff
@@ -82,6 +82,7 @@ class ExtendedPublicKeyImporter extends React.Component {
             variant="standard"
             onChange={this.handleMethodChange}
           >
+            <MenuItem value={BITBOX}>BitBox</MenuItem>
```
chadchapnick, line 85: Noticed we will need this option in the wallet spend—specifically `apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx` 

Suggesting:
```js
// Line 123
<MenuItem value={BITBOX}>BitBox</MenuItem>

// Line 158
if (method === TREZOR || method === LEDGER || method === BITBOX) {
  return (
    <DirectSignatureImporter
           ...
```

Screenshot 2024-08-31 at 20 54 34

packages/caravan-wallets/src/index.ts
```diff
@@ -335,6 +359,12 @@ export function SignMultisigTransaction({
   progressCallback,
 }: SignMultisigTransactionArgs) {
   switch (keystore) {
+    case BITBOX:
+      return new BitBoxSignMultisigTransaction({
+        walletConfig,
+        psbt,
```
chadchapnick, line 365: It looks like the `psbt` passed here is actually undefined. Short of us applying a proper fix, a quick hack is to do what's done below with ledger
```js
@@ -360,9 +360,18 @@ export function SignMultisigTransaction({
 }: SignMultisigTransactionArgs) {
   switch (keystore) {
     case BITBOX:
+      let _psbt = psbt;
+      if (!_psbt)
+        _psbt = getUnsignedMultisigPsbtV0({
+          network,
+          inputs: inputs ? inputs.map(convertLegacyInput) : [],
+          outputs: outputs ? outputs.map(convertLegacyOutput) : [],
+        }).toBase64();
       return new BitBoxSignMultisigTransaction({
         walletConfig,
-        psbt,
+        psbt: _psbt,
         returnSignatureArray,
       });
```

Tried this on my machine with a P2WSH PSBT and and got stuck at [Loading transaction...](https://github.com/BitBoxSwiss/bitbox02-firmware/blob/2c737961e830ba2dd1b2288c7b8211ef71ae4092/src/rust/bitbox02-rust/src/hww/api/bitcoin/signtx.rs#L591) when running [`pairedBitBox.btcSignPSBT`](https://github.com/BitBoxSwiss/bitbox-api-rs/blob/7abaa32aafbea466f4330a8859b318ade4f536c1/src/wasm/mod.rs#L342)

image
[samplepsbt.txt](https://github.com/user-attachments/files/16829714/samplepsbt.txt)

chadchapnick: Note this repo needs a fix for segwit PSBTs and this is my current hack
[cc_p2wsh_fix.patch](https://github.com/user-attachments/files/16829754/cc_p2wsh_fix.patch)

benma: @chadchapnick thanks for the review and for testing.

As for the issue you saw here, this is because the PSBT is missing the non_witness_utxo field in the inputs. See this comment from the commit message:

> - The BitBox02 requires the previous transaction of each input to be
present in the PSBT (`PSBT_IN_NON_WITNESS_UTXO`), so it can verify the
input amount and avoid fee attacks. The signing test fixtures are
missing these, so they fail.

@bucko13 mentioned to me that this is only a problem with the test fixtures, but that you do supply this field in production.

chadchapnick: @benma thank you for the clarification! Can confirm signing on device works very smoothly now for both P2WSH and P2SH-P2WSH. Fascinating to peer into the stack you folks have architected.

benma: I applied the workaround for the undefined psbt like Ledger as you suggested

apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
benma: I made the change, but I couldn't quickly figure out how to find this in the UI. Could you you test it?

chadchapnick: Looks great—thanks! It's accessible under the spend tab after loading the wallet + building a transaction
Screenshot 2024-09-02 at 22 47 07

packages/caravan-wallets/src/index.ts
chadchapnick: We're going to fix this upstream so this hack can actually be dropped, and we can use what you had before—thank you for entertaining it 
```js
BitBoxSignMultisigTransaction({
         walletConfig,
         psbt,
         returnSignatureArray,
       });
```

packages/caravan-wallets/src/bitbox.ts
```diff
@@ -0,0 +1,467 @@
...
+      const address = await pairedBitBox.btcAddress(
+        convertNetwork(this.network),
+        this.bip32Path,
+        scriptConfig,
+        display,
+      );
+      return {
+        address,
+        serializedPath: this.bip32Path,
+      };
```
chadchapnick, line 386: Looks like this is mostly working—noticing we get back a testnet address when on regtest
image

benma: The BitBox02 itself does not actually support regtest right now. Adding support for it would be easy, and in fact I made a PR for it now (https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1302). It would take a little while until it is released and can be used here though.

Another solution here would be to convert the address to the regtest format inside Caravan after the BitBox02 returns it in the testnet format.

What do you think is the best course of action?

chadchapnick: In favor for keeping it _elegant_ without the regtest hack for P2WSH! Nested segwit works perfectly and that's likely sufficient for now. If your firmware PR makes it in, it will puzzle in nicely.

Thank you sharing and for suggesting the address conversion, it does work well
```
import { decodeSegwitAddress, encodeSegwitAddress } from "./vendor/bech32";
if (this.walletConfig.network == "regtest") {
  if (address.startsWith("tb1q")) {
    try {
      const decoded = decodeSegwitAddress("tb", address);
      if (!decoded) {
        throw new Error("Could not decode address");
      }
      const resp = encodeSegwitAddress(
        "bcrt",
        decoded.version,
        decoded.program
      );
      return {
        address: resp,
        serializedPath: this.bip32Path,
      };
    } catch (err) {
      console.log("error is", err);
    }
  }
}
```

apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
chadchapnick: An additional consideration is we may want to hide BitBox in the various menus when using P2SH—attached a quick patch to assist with this
[hide_on_p2sh.patch](https://github.com/user-attachments/files/16933830/hide_on_p2sh.patch)

packages/caravan-wallets/src/bitbox.ts
bucko13: Sorry for the delay @benma. 

That's right, it's a function so that the underlying messages can be updated based on different states. You would add the pairing messages dictating a specific state that it should be relevant for. I think it should be fine to build in some UX so that the user clicks a button that prompts user to "confirm pairing code". Clicking that would call the "messagesFor" to get the pairing code. If it's not available, tell the user and re-enable the button. 

Can we at least include in this PR adding the message? I think the rest of the above should also work, but I haven't tried obviously so I might be wrong.

bucko13:
I'm surprised you were able to get this to run. I had to add the following to the coordinator's vite config:


```javascript
export default defineConfig({
    ...
    build: {
        target: "esnext", //browsers can handle the latest ES features
         ...
    },
}
```

Without that, I get these failures:

```
caravan-coordinator:build: error during build:
caravan-coordinator:build: Error: Transform failed with 1 error:
caravan-coordinator:build: assets/bitbox_api-!~{004}~.js:1371:27: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 3 overrides)
```

There are also a couple of linting [errors](https://github.com/caravan-bitcoin/caravan/actions/runs/10664318474/job/30492222714?pr=117#step:5:342) that need to be fixed.

benma:
> I'm surprised you were able to get this to run. I had to add the following to the coordinator's vite config:
> 
> ```js
> export default defineConfig({
>     ...
>     build: {
>         target: "esnext", //browsers can handle the latest ES features
>          ...
>     },
> }
> ```

I only developed using `turbo run dev`, where this error never appeared. I can see it however when running `turbo run build`. I added the above line.

> There are also a couple of linting [errors](https://github.com/caravan-bitcoin/caravan/actions/runs/10664318474/job/30492222714?pr=117#step:5:342) that need to be fixed.

Fixed :+1:

apps/coordinator/src/components/Wallet/ExtendedPublicKeyImporter.jsx
benma: Thanks, I applied the patch :+1:

packages/caravan-wallets/src/bitbox.ts
benma: I am afraid it won't work this way. The library client (e.g. the coordinator) does not have a way of knowing when the pairing code is being shown, which it needs to know in order to ask the user to confirm. 

I added a commit with a potential solution, please check it out. We can scrap it again if it is not good, it's just a proposal with a working demo:

- All interactions have a `showPairingCode` callback that can be supplied to show/hide the code when the BitBox asks for it
- If the callback is not provided (not implemented), by default a popup is opened to show the code, so at least there a way of checking the code even if the client did not implement a specific way of showing the pairing code.

What do you think?

bucko13: Ok, I think you've convinced me :). Thanks for your patience. Can we clean up the console log now that you've got the callback?

packages/caravan-wallets/src/bitbox.ts
```diff
@@ -0,0 +1,551 @@
...
+    case 'P2WSH':
+      return 'p2wsh';
+    case 'P2SH-P2WSH':
+      return 'p2wshP2sh';
+    default:
+      throw new Error(`BitBox does not support ${addressType} multisig.`);
+  }
+}
+
+async function convertMultisig(pairedBitBox: PairedBitBox, walletConfig: MultisigWalletConfig): Promise<{ scriptConfig: BtcScriptConfig; keypathAccount: string; }> {
```
bucko13, line 57: It feels like this function is doing two things and it might cleaner to separate them. Correct me if I'm wrong but what it's doing is:

1. verifying that the devices root fingerprint is in a given wallet config
2. converting the caravan style wallet config to bitbox style 

What if the verification was a method on the interaction base class where you know you'll have a paired bitbox you can interact with and then this was kept as a simple conversion utility function? 

Having the verification on the interaction would allow it to be used in other interactions as well. could even retrieve the root fingerprint presumably on startup and store it as a property on the class or something and then use it as needed. Not sure if that's strictly correct or necessary, but I did find this a little hard to grok the way it is.

packages/caravan-wallets/src/bitbox.ts
bucko13: @benma any sense of when the updated firmware might be released?

packages/caravan-wallets/src/bitbox.ts
```diff
@@ -0,0 +1,551 @@
...
+      throw new Error(errorMessage);
+    } finally {
+      if (hidePairingCode) {
+        hidePairingCode();
+      }
+    }
+  }
+
+  async maybeRegisterMultisig(pairedBitBox: PairedBitBox, walletConfig: MultisigWalletConfig): Promise<{ scriptConfig: BtcScriptConfig, keypathAccount: string; }> {
+    const { scriptConfig, keypathAccount } = await convertMultisig(pairedBitBox, walletConfig);
```
bucko13, line 219: Could do something like: 
```suggestion
    const rootFingerprint = await pairedBitBox.rootFingerprint()
    this.verifyConfig(rootFingerprint, walletConfig)
    // or this.verifyConfig(pairedBitBox, walletConfig) which could do the fingerprint retrieval
    const { scriptConfig, keypathAccount } = convertMultisig(walletConfig);
```

packages/caravan-wallets/src/bitbox.ts
```diff
@@ -0,0 +1,551 @@
...
+    const isRegistered = await pairedBitBox.btcIsScriptConfigRegistered(
+      convertNetwork(walletConfig.network),
+      scriptConfig,
+      keypathAccount,
+    );
+    // No name means the user inputs it on the device.
+    // eslint-disable-next-line no-undefined
+    const name = undefined;
```
bucko13, line 227: There is a name that comes in from the wallet config actually so maybe it should just use that (same is done for ledger and coldcard)

bucko13 (changes requested):
This looks great! Just a couple of small things that maybe aren't even blockers. 

I'd like us to add unit tests for the `@caravan/wallets` code, but since we do have the test suite at least, we don't need to block this going in. 

It would also be nice to have the regtest firmware update in so we can do more rigorous testing, but otherwise I'm excited to have this merged in soon!

packages/caravan-wallets/src/bitbox.ts
benma: It was released a few weeks ago - if you update your BitBox02 to v9.21.0 it will show the proper regtest address prefix. 

You can update to this firmware most easily via BitBoxApp v4.45+.

I also added a commit here to make use of it.

packages/caravan-wallets/src/bitbox.ts
benma: There might be trouble as the BitBox02 has some requirements on the name - max 30 chars, only printalble ASCII chars (including space), etc. If entered on the device it cannot fail.

packages/caravan-wallets/src/bitbox.ts
benma: > It feels like this function is doing two things and it might cleaner to separate them. Correct me if I'm wrong but what it's doing is:

>    1. verifying that the devices root fingerprint is in a given wallet config
>    2. converting the caravan style wallet config to bitbox style

It is not doing the first one. It is merely trying to identify which of the xpubs belongs to the device, as  that is a required parameter in the script config passed to the BitBox (`ourXpubIndex`).

benma:
> This looks great! Just a couple of small things that maybe aren't even blockers.
> 
> I'd like us to add unit tests for the `@caravan/wallets` code, but since we do have the test suite at least, we don't need to block this going in.
> 
> It would also be nice to have the regtest firmware update in so we can do more rigorous testing, but otherwise I'm excited to have this merged in soon!

Thanks! The firmware with regtest has been released and I integrated the change in a new commit, please check it out. See https://github.com/caravan-bitcoin/caravan/pull/117#discussion_r1824077840

bucko13 (approved):

PR #112, opened 2024-07-03
Caravan Health Package

Harshil-Jani:
TODO : 
- Discuss if we really want a standalone app or can we integrate it into the existing coordinator.
- Complete all the bitcoind or block explorer client related configurations. (I would need to use lots of caravan-bitcoin functionalities)
- Finalize the frontend design
- Take a decision of showing Z-Score.
- Writing a blog and linking it at top of the package so people don't get confused on math.

PR #118, opened 2024-08-19
Advising on UTXO Dusting based on feeRate (Frontend)

Harshil-Jani:
Will rebase this branch once the `@caravan/health` is merged 
#112 

image

I am not a fan of the current naming, But it's a simple change so let me know what could be appropriate naming convention in all three cases. Also the table heading "Spend Status". 

Green - Hold
Yellow - Risky
Red - Dust

### DIFF
--- a/apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
+++ b/apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
@@ -327,6 +327,8 @@ UTXOSet.propTypes = {
   existingTransactionInputs: PropTypes.arrayOf(PropTypes.shape({})),
   setSpendCheckbox: PropTypes.func,
   autoSpend: PropTypes.bool.isRequired,
+  feeRate: PropTypes.string,
+  walletConfig: PropTypes.object,
 };
 
 UTXOSet.defaultProps = {
--- a/apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
+++ b/apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
@@ -14,14 +14,12 @@ import {
   TableCell,
   Typography,
   Checkbox,
-  Chip,
 } from "@mui/material";
 import { OpenInNew } from "@mui/icons-material";
 import BigNumber from "bignumber.js";
 import { externalLink } from "utils/ExternalLink";
 import Copyable from "../Copyable";
-import { WasteMetrics } from "@caravan/health";
-import { getWalletConfig } from "../../selectors/wallet";
+import DustChip from "./DustChip";
 
 // Actions
 import { setInputs as setInputsAction } from "../../actions/transactionActions";
@@ -173,49 +171,13 @@ class UTXOSet extends React.Component {
     }
   };
 
-  getChipProperties(amountSats, feeRate, walletConfig) {
-    const wasteMetrics = new WasteMetrics();
-    const scriptType = walletConfig.addressType;
-    const config = {
-      requiredSignerCount: walletConfig.quorum.requiredSigners,
-      totalSignerCount: walletConfig.quorum.totalSigners,
-    };
-    const { lowerLimit, upperLimit } = wasteMetrics.calculateDustLimits(
-      feeRate,
-      scriptType,
-      config,
-    );
-    if (amountSats <= lowerLimit) {
-      return {
-        color: "error",
-        label: "dust",
-      };
-    } else if (amountSats > lowerLimit && amountSats <= upperLimit) {
-      return {
-        color: "warning",
-        label: "warning",
-      };
-    } else {
-      return {
-        color: "success",
-        label: "economical",
-      };
-    }
-  }
-
   renderInputs = () => {
-    const { network, showSelection, finalizedOutputs, walletConfig, feeRate } =
-      this.props;
+    const { network, showSelection, finalizedOutputs, feeRate } = this.props;
     const { localInputs } = this.state;
     return localInputs.map((input, inputIndex) => {
       const confirmedStyle = `${styles.utxoTxid}${
         input.confirmed ? "" : ` ${styles.unconfirmed}`
       }`;
-      const { color, label } = this.getChipProperties(
-        input.amountSats,
-        feeRate,
-        walletConfig,
-      );
       const confirmedTitle = input.confirmed ? "confirmed" : "unconfirmed";
       return (
         <TableRow hover key={input.txid}>
@@ -243,7 +205,7 @@ class UTXOSet extends React.Component {
             <Copyable text={satoshisToBitcoins(input.amountSats)} />
           </TableCell>
           <TableCell>
-            <Chip color={color} label={label}></Chip>
+            <DustChip amountSats={input.amountSats} feeRate={feeRate} />
           </TableCell>
           <TableCell>
             {externalLink(
@@ -328,7 +290,6 @@ UTXOSet.propTypes = {
   setSpendCheckbox: PropTypes.func,
   autoSpend: PropTypes.bool.isRequired,
   feeRate: PropTypes.string,
-  walletConfig: PropTypes.object,
 };
 
 UTXOSet.defaultProps = {
@@ -348,7 +309,6 @@ function mapStateToProps(state) {
     autoSpend: state.spend.transaction.autoSpend,
     finalizedOutputs: state.spend.transaction.finalizedOutputs,
     existingTransactionInputs: state.spend.transaction.inputs,
-    walletConfig: getWalletConfig(state),
   };
 }
 
--- a/apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
+++ b/apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
@@ -14,11 +14,14 @@ import {
   TableCell,
   Typography,
   Checkbox,
+  Chip,
 } from "@mui/material";
 import { OpenInNew } from "@mui/icons-material";
 import BigNumber from "bignumber.js";
 import { externalLink } from "utils/ExternalLink";
 import Copyable from "../Copyable";
+import { WasteMetrics } from "@caravan/health";
+import { getWalletConfig } from "../../selectors/wallet";
 
 // Actions
 import { setInputs as setInputsAction } from "../../actions/transactionActions";
@@ -170,13 +173,49 @@ class UTXOSet extends React.Component {
     }
   };
 
+  getChipProperties(amountSats, feeRate, walletConfig) {
+    const wasteMetrics = new WasteMetrics();
+    const scriptType = walletConfig.addressType;
+    const config = {
+      requiredSignerCount: walletConfig.quorum.requiredSigners,
+      totalSignerCount: walletConfig.quorum.totalSigners,
+    };
+    const { lowerLimit, upperLimit } = wasteMetrics.calculateDustLimits(
+      feeRate,
+      scriptType,
+      config,
+    );
+    if (amountSats <= lowerLimit) {
+      return {
+        color: "error",
+        label: "dust",
+      };
+    } else if (amountSats > lowerLimit && amountSats <= upperLimit) {
+      return {
+        color: "warning",
+        label: "warning",
+      };
+    } else {
+      return {
+        color: "success",
+        label: "economical",
+      };
+    }
+  }
+
   renderInputs = () => {
-    const { network, showSelection, finalizedOutputs } = this.props;
+    const { network, showSelection, finalizedOutputs, walletConfig, feeRate } =
+      this.props;
     const { localInputs } = this.state;
     return localInputs.map((input, inputIndex) => {
       const confirmedStyle = `${styles.utxoTxid}${
         input.confirmed ? "" : ` ${styles.unconfirmed}`
       }`;
+      const { color, label } = this.getChipProperties(
+        input.amountSats,
+        feeRate,
+        walletConfig,
+      );
       const confirmedTitle = input.confirmed ? "confirmed" : "unconfirmed";
       return (
         <TableRow hover key={input.txid}>
@@ -202,6 +241,9 @@ class UTXOSet extends React.Component {
           </TableCell>
           <TableCell>
             <Copyable text={satoshisToBitcoins(input.amountSats)} />
+          </TableCell>
+          <TableCell>
+            <Chip color={color} label={label}></Chip>
           </TableCell>
           <TableCell>
             {externalLink(
@@ -247,6 +289,7 @@ class UTXOSet extends React.Component {
               <TableCell>TXID</TableCell>
               <TableCell>Index</TableCell>
               <TableCell>Amount (BTC)</TableCell>
+              <TableCell>Dust Status</TableCell>
               <TableCell>View</TableCell>
             </TableRow>
           </TableHead>
@@ -303,6 +346,7 @@ function mapStateToProps(state) {
     autoSpend: state.spend.transaction.autoSpend,
     finalizedOutputs: state.spend.transaction.finalizedOutputs,
     existingTransactionInputs: state.spend.transaction.inputs,
+    walletConfig: getWalletConfig(state),
   };
 }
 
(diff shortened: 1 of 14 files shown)


apps/coordinator/src/components/Wallet/WalletSpend.jsx
```diff
@@ -62,6 +64,12 @@ class WalletSpend extends React.Component {
     }
   };
 
+  handleFeeEstimate = (feeEstimate) => {
+    const wasteMetrics = new WasteMetrics();
+    const dustLimits = wasteMetrics.calculateDustLimits(feeEstimate);
```
bucko13, line 69: Based on the description in the code and readme, I would've expected "spend waste metric" to be useful here:

```
     Spend Waste Score (S.W.S)
    Definition : 
      A score that indicates whether it is economical to spend a particular output now 
      or wait to consolidate it later when fees could be low.
```

Harshil-Jani: This was more about the transaction which we are forming and not for single UTXO. As commented here https://github.com/caravan-bitcoin/caravan/pull/112#discussion_r1726336091

We can manually select more than one UTXOs to make a transactions so I was not sure if this could be the correct metric to tell if a single UTXO should be present in wallet or should be spent or consolidated etc. This could fit more into the preview transaction tab where we have all the details right in place. And there we can suggest, If making this transaction now is going to save some sats or is going to affect badly in long term.

bucko13: I see! Yeah that makes sense. Thanks for the explanation!

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
```diff
@@ -173,11 +174,32 @@ class UTXOSet extends React.Component {
...
     const { localInputs } = this.state;
+    const dustLimits = this.props.dustLimits;
+
+    function getChipProperties(amountSats) {
+      const lowerLimit = dustLimits.lowerLimit;
+      const upperLimit = dustLimits.upperLimit;
+      let chipColor;
+      let chipLabel;
+      if (amountSats <= lowerLimit) {
+        chipColor = "#f55656"; // Indicating critical state
```
bucko13, line 185: Please use the MUI theme for defining colors so these can be updated programmatically and also fit with the rest of the theme: https://mui.com/material-ui/customization/palette/#default-colors

probably using `success`, `warning`, and `error`

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
```diff
@@ -173,11 +174,32 @@ class UTXOSet extends React.Component {
...
+        chipLabel = "Dust"; // Immediate action required
+      } else if (amountSats > lowerLimit && amountSats <= upperLimit) {
+        chipColor = "#ebeb49"; // Indicating caution
+        chipLabel = "Risky"; // Monitor closely
+      } else {
+        chipColor = "#53f559"; // Indicating safe state
+        chipLabel = "Hold"; // No immediate action needed
+      }
+      return { color: chipColor, label: chipLabel };
+    }
```
bucko13, line 195: What if fee rate was provided here (maybe via redux, context, or props)? and then you could do something like this (with a couple other cleanups):

```suggestion
    function getChipProperties(amountSats, feeRate) {
     const wasteMetrics = new WasteMetrics();
     const {lowerLimit, upperLimmit} = wasteMetrics.calculateDustLimits(feeRate);
     // need to use the theme context for this of course
      const { warning, error, success } = theme.vars.palette 
      if (amountSats <= lowerLimit) {
        return {
          color: error
          label: "dust"
        }
      } else if (amountSats > lowerLimit && amountSats <= upperLimit) {
        return {
          color: warning,
          label: "warning"
      } else {
        return { 
          color: success,
          label: "economical"
        }
      }
    }
```
(the above could also probably be done as a switch statement) 

When/if this were converted to a functional component, this could could even be converted to a hook such that you can call like:

```
const { color, label } = useUtxoHealthChipProps(amountSats, feeRate)
```

bucko13 (changes requested):
A couple suggestions for cleanup

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
Harshil-Jani: We could use, WasteMetrics() instance directly here too. My thought process was that this component will load for each UTXO so using it directly over here means that if someone has 1000 UTXOs (Bitcoin Whale) then it will just create 1000 instance of WasteMetrics(), which could be a performance overhead. Taking it from state will ensure only one initialization call on top parent component of the page. Am I correct ?

For the next comments on colour scheme and MUI compatibility, I will update it as needed.

bucko13:
Could you rebase ontop of `main` now that the health package has been merged in. It would help to cleanup the diff and resolve the merge conflicts.

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
bucko13: Honestly, I'd be surprised if it added too much overhead just instantiating the class especially since you're not loading it up with a wallet in anyway. It wouldn't be too different from if you had exposed this simply as a utility function.

Ideally we reduce the reliance on the redux store as much as possible. The original implementation of the wallet code relied on redux far too much anyway and the react ecosystem has moved in a direction where redux is kind of an anti-pattern. Reusable hooks and context are the preferred way to handle state management. 

If you wanted to reduce the instantiation of the wallet metrics, you could make it (and privacy metrics) into a singleton that is shared across the app. You could even do this in a set of hooks, something like `useWasteMetrics` and `usePrivacyMetrics` that could get the wallet info from the store, instantiate an instance of the class if it doesn't already exist in the global namespace or use the existing one if it does, and make sure they update if the wallet updates. This would also have the downside though of not being useable in the older class based components.

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -165,6 +166,20 @@ class TransactionPreview extends React.Component {
       unsignedPSBT,
     } = this.props;
 
+    const wasteMetrics = new WasteMetrics();
+    const weight = fee / satoshisToBitcoins(feeRate);
+    // Doubt-1: How do I know what is the amount that I want to actually spend and what is change amount?
+    const spendAmount = 10000;
+    // Doubt-2: In caravan context, How do we want to plan L value?
+    const L = 100; // longTermFeeEstimate
```
Harshil-Jani, line 174: For now, Just to try out the functionality, I have hardcoded random values. But how would we populate here the actual values ?

bucko13: For the first doubt- all of that information is there on the transaction preview screen. You can see the change output has a little `(change)` marked there. Since this is a transaction that is created by the user, change is known because it has to be decided by the client/application ahead of time. 

"amount to spend" are the other outputs

Screenshot 2024-09-07 at 8 31 27 PM

For `L` this is a good question. You could come up with some formula to estimate and hardcode with explanation. If you're going to do that, then I'd suggest having that in the package to start with and using that as a default. 

I think an overall better UX would be to default to the current "fast" fee rate (maybe with some percentage above as a buffer) and then in addition have a slider (something like [this](https://mui.com/material-ui/react-slider/) to use as the component. an example of a similar implementation is [Live Wallet](https://github.com/Jwyman328/LiveWallet)) that allows the user to set it themselves and see how that impacts the rating of their transaction.

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
```diff
@@ -170,13 +173,49 @@ class UTXOSet extends React.Component {
     }
   };
 
+  getChipProperties(amountSats, feeRate, walletConfig) {
+    const wasteMetrics = new WasteMetrics();
+    const scriptType = walletConfig.addressType;
```
bucko13, line 178: if you're doing this in the component, there's no need to pass as function arguments since these are available on the props.

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
```diff
@@ -203,6 +242,9 @@ class UTXOSet extends React.Component {
           <TableCell>
             <Copyable text={satoshisToBitcoins(input.amountSats)} />
           </TableCell>
+          <TableCell>
+            <Chip color={color} label={label}></Chip>
```
bucko13, line 246: Just a nit, but a "cleaner" way to do this (which we don't do well in the older code of this repo unfortunately) would be as its own component:

```javascript
const DustChip = ( { amountSats, feeRate, walletConfig }) => {
    const wasteMetrics = new WasteMetrics();
    const scriptType = walletConfig.addressType;
    const config = {
      requiredSignerCount: walletConfig.quorum.requiredSigners,
      totalSignerCount: walletConfig.quorum.totalSigners,
    };
    const { lowerLimit, upperLimit } = wasteMetrics.calculateDustLimits(
      feeRate,
      scriptType,
      config,
    );
    let chipArgs = {
        color: "success",
        label: "economical",
      };

    if (amountSats <= lowerLimit) {
      chipArgs = {
        color: "error",
        label: "dust",
      };
    } else if (amountSats > lowerLimit && amountSats <= upperLimit) {
      chipArgs = {
        color: "warning",
        label: "warning",
      };
    } 

  return <Chip {...chipArgs} />
}
```

And then you can do 
```jsx
<DustChip  amountSats={input.amountSats} feeRate={feeRate} walletConfig={walletConfig} />
```

you could even probably pull the `walletConfig` from the redux state using `useSelector` and remove that as a prop that's passed around: 

```javascript
  // you can search around the codebase for this as it's used in a few places
  const walletConfig = useSelector(getWalletConfig);
```

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -186,6 +231,23 @@ class TransactionPreview extends React.Component {
...
+            <Slider
+              value={longTermFeeEstimate}
+              min={1}
+              max={3000}
+              step={1}
+              onChange={this.handleLongTermFeeEstimateChange}
+              aria-labelledby="long-term-fee-estimate-slider"
+            />
+            <Typography id="long-term-fee-estimate-slider" gutterBottom>
+              Long Term Fee Estimate (L): {longTermFeeEstimate} sats/vB
```
bucko13, line 248: Can we add some info tooltips maybe to explain what the waste analysis, the SWA and the L are?

bucko13 (commented):
Looking good! Some minor comments/suggestions

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
Harshil-Jani: Sure

apps/coordinator/src/components/ScriptExplorer/UTXOSet.jsx
Harshil-Jani: Updated

bucko13:
Got some build failures:

```
caravan-coordinator:build:    6:21  error  'amountSats' is missing in props validation  react/prop-types
caravan-coordinator:build:    6:33  error  'feeRate' is missing in props validation     react/prop-types
caravan-coordinator:build:   36:10  error  'React' must be in scope when using JSX      react/react-in-jsx-scope
```

bucko13 (changes requested):
Couple small code comments and then some feedback/questions about implementation.

- I don't think we need the decimal point for the SWA
- Not pictured in the screenshot, but it looks like the component initializes to 0.00 SWA and then as soon as the slider moves it goes to 80m
- Maybe I'm still a little confused about SWA. But the tooltip says "Postive SWA means that it would be efficient to spend this transaction later when the feerate decreases". In the screenshot below, I have a "normal" tx, spending .2 btc with a fee rate of 1 sat/vbyte. But based on the SWA, it's telling me to wait until fees are lower before spending this transaction? Besides the fact that they can't be lower than 1, can this be right? This intuitively feels wrong as now seems like a fine time to spend a tx that creates utxos of this kind at that fee rate (this tx is in regtest). 
- It's also still not clear what this number actually means. L at 1 sat/vbyte means the tx has a score of 80m sats? Does it cost me 80m sats to spend now? to spend later? Something else? If it's just a score than it probably should have a unit of satoshis right? if it means something more concrete about cost, then the units make sense but the tooltip needs different details. 
- If the SWA is going to be that high, then can we add commas? `80,000,000` instead of `80000000`
- I think the L range of 0-3000 is probably larger than it needs to be. I'd say a max of 500 is probably fine as we've never really seen sustained fee rates that high. It should be easy to raise if we need to in the future right? You could even do it as a function of current fees (either like a multiple of ~50 or some logarithmic function). Not necessary for now though, just a thought. 

Screenshot 2024-09-10 at 8 56 49 AM

packages/caravan-health/src/waste.ts
```diff
@@ -54,7 +54,7 @@ export class WasteMetrics extends WalletMetrics {
       We can compare this ratio against the fiat charges for cross-border transactions.
       Mastercard charges 0.6% cross-border fee for international transactions in US dollars, 
       but if the transaction is in any other currency the fee goes up to 1%. 
-      Source : https://www.clearlypayments.com/blog/what-are-cross-border-fees-in-credit-card-payments/
+      Source : https://www.clearlypayments.com/blog/what-are-cross-um);border-fees-in-credit-card-payments/
```
bucko13, line 57: typo?

packages/caravan-health/src/waste.ts
```diff
@@ -102,10 +102,10 @@ export class WasteMetrics extends WalletMetrics {
           The amount by which we exceed our selection target when creating a changeless transaction, 
           mutually exclusive with cost of change. It is extra fees paid if we don't make a change output 
           and instead add the difference to the fees.
-      - Input Amount :
-          Sum of amount for each coin in input of the transaction
-      - Spend Amount :
-          Exact amount wanted to be spent in the transaction.
+      - Scripy Type:
```
bucko13, line 105: > Script Type

packages/caravan-health/src/waste.ts
```diff
@@ -102,10 +102,10 @@ export class WasteMetrics extends WalletMetrics {
...
+          The script type of the wallet. The change will always be of the same script type as the wallet.
```
bucko13, line 106: Maybe for clarification: 

> This is for calculating size of change output which will always be the same script type as the wallet

packages/caravan-health/src/waste.ts
```diff
@@ -102,10 +102,10 @@ export class WasteMetrics extends WalletMetrics {
...
+      - Config:
```
bucko13, line 107: Maybe just include the script type here? elsewhere in caravan we talk about a "wallet config" which includes this info (theoretically it could even be a descriptor). 

The actual config json unfortunately nests `m` and `n` in a `quorum` object. You could use the `MultisigWalletConfig` type that is exported from `@caravan/multisig` but if you didn't want to dig into the nested object, either keeping the related args together in the config arg or flattening all into top level args OR all into the same level into a single option object, e.g. 
```javascript
spendWasteAmount({
  weight, 
  feeRate,
  scriptType, 
  requiredSignerCount,
  totalSignerCount,
})
```
Could be cleaner.

packages/caravan-health/src/waste.ts
```diff
@@ -120,11 +120,14 @@ export class WasteMetrics extends WalletMetrics {
...
-    spendAmount: number, // Exact amount wanted to be spent in the transaction
+    scriptType: MultisigAddressType,
+    config: {
+      requiredSignerCount: number;
+      totalSignerCount: number;
+    },
     estimatedLongTermFeeRate: number, // Long term estimated fee-rate
   ): number {
-    const costOfTx: number = Math.abs(spendAmount - inputAmountSum);
+    const costOfTx: number = getInputWeight(scriptType, config) * feeRate;
```
bucko13, line 130: nice cleanup!

packages/caravan-health/src/spendType.ts
```diff
@@ -75,3 +76,43 @@ export function getSpendTypeScore(
       throw new Error("Invalid spend type");
   }
 }
+
+export function getInputWeight(
+  scriptType: MultisigAddressType,
+  config: { requiredSignerCount: number; totalSignerCount: number },
+): number {
```
bucko13, line 83: I think this is a bit cleaner. 
```suggestion
  { 
      requiredSignerCount,
      totalSignerCount
      scriptType
  }: { 
    requiredSignerCount: number; 
    totalSignerCount: number; 
    scriptType: MultisigAddressType 
}): number {
```

Legend101Zz:
@bucko13  can we close this in favour of the new SOB wallet health project ? 
We can use it as a reference ...

Issue #115, opened 2024-07-25
Migrate from jest to vitest


bucko13:
Jest as a project seems to be [under-resourced](https://github.com/jest-community/jest-extended/issues/205). It has [issues](https://archive.jestjs.io/docs/en/next/ecmascript-modules) with newer technologies like ES Modules. Vitest seems to be better maintained, more modern and has a better DX/UX>

Legend101Zz:
Hi Buck .... 
So do you want to rewrite all the tests for the packages and coordinator to vitest ?
I'll be more than happy to contribute once I am done with integrating the fees package to the main coordinator  :)

bucko13:
That’s the idea yeah! And that’d be amazing. Vitest did try and keep its API as close to the same as jest as possible so hopefully it’s an easy transition 🤞

Legend101Zz:
Ah Great !! So can we start with packages first ... Because I have a got a solid grasp of bitcoin , psbt and client packages so rewriting tests for them would be a great  starting point for me I guess ... also learning something new is always great :)

DhairyaMajmudar:
Hi @bucko13 @Legend101Zz this issues seems to be interesting, would love to collaborate, found a very good blog which can be helpful

https://medium.com/wiiisdom-labs/how-to-migrate-from-jest-to-vitest-without-headaches-ee200510738e

Legend101Zz:
Hi @DhairyaMajmudar thanks for help, would you like to contribute with me to start working on this issue .... we were planning on starting with the packages first ... so maybe we can decide on how to divide them and start by creating a new branch to which both then can contribute and create a draft PR for it then ...

DhairyaMajmudar:
Yes definitely @Legend101Zz  let's collaborate : )

Legend101Zz:
Great , what would be your preferred channel of communication @DhairyaMajmudar , if you prefer we can keep the conversations here itself or else anywhere that you suggest .. :)

DhairyaMajmudar:
> Great , what would be your preferred channel of communication @DhairyaMajmudar , if you prefer we can keep the conversations here itself or else anywhere that you suggest .. :)

Is there a Slack/Discord of caravan where we can connect ?

Legend101Zz:
I don't know of any such existing channels .... there is a discord subchannel of caravan in SOB's discord but I think that might not be the right place  , but maybe @bucko13 can help us with this ....  :)

Legend101Zz:
[...]

PR #119, opened 2024-08-27
[Test]: Jest to Vitest migration `caravan-client` package

DhairyaMajmudar:
### Screenshot

image

cc: @bucko13 @Legend101Zz


DhairyaMajmudar:
Needed one suggestion: 

How's about keeping test files in `__test__` folder this will help test and main files to keep separated and also gives below looking cool folder structure : )

image

bucko13:
Very nice! Reorganizing to a __tests__ directory sounds good to me as well. 

Something also to consider as part of the larger #115 issue is to update the template generator and tooling packages to have and use a shared vitest config as well as update the shared tsconfig to include the vitest

bucko13:
CI seems to be having trouble with this for some reason: https://github.com/caravan-bitcoin/caravan/actions/runs/10584531073/job/29340663941?pr=119

Have you tried running `npm run ci` locally to see if it's reproducible? I can't see anything obvious in the diff.

DhairyaMajmudar:
> Very nice! Reorganizing to a **tests** or **test** directory sounds good to me as well.
> 
> Something also to consider as part of the larger #115 issue is to update the template generator and tooling packages to have and use a shared vitest config as well as update the shared tsconfig to include the vitest

Thank you! yeah keeping  a common vitest config file will be a great idea leveraging the power of monorepos : )

DhairyaMajmudar:
> CI seems to be having trouble with this for some reason: https://github.com/caravan-bitcoin/caravan/actions/runs/10584531073/job/29340663941?pr=119
> 
> Have you tried running `npm run ci` locally to see if it's reproducible? I can't see anything obvious in the diff.

I've run the command in local and while running so I've observed that the process is not getting exit after completion, this might be  the reason Github actions got cancelled after a certain time

I think it will be good if we separate the linting, testing and building pipelines, and that too in both package and app folders (If files are changed and package folder/s than CI will run only for that not for app folder/s) this will reduce the GHA execution time. 

What are your thoughts on this @bucko13 ?

bucko13:
> this might be the reason Github actions got cancelled after a certain time

it actually never canceled! I had to manually run it. 

> I think it will be good if we separate the linting, testing and building pipelines,

You can try that for sure, but just want to note that the CI is passing in other PRs. It's just this one where it gets stuck.

DhairyaMajmudar:
> > this might be the reason Github actions got cancelled after a certain time
> 
> it actually never canceled! I had to manually run it.
> 
> > I think it will be good if we separate the linting, testing and building pipelines,
> 
> You can try that for sure, but just want to note that the CI is passing in other PRs. It's just this one where it gets stuck.

okayyy! I've made minor changes in the test scripts and it seems like now ci is passing.

vitest.config.ts
bucko13, line 1: we should have vitest configs on a per pacakge basis but that use a shared config that's kept as its own package. You can see how we do this for tsconfig and eslint. [Here](https://turbo.build/repo/docs/guides/tools/eslint#setting-up-a-lint-task) are some docs from turborepo on this too, and [how to add to generators](https://turbo.build/repo/docs/guides/generating-code) for making this the default for new packages.

packages/caravan-clients/src/__tests__/client.test.ts
```diff
@@ -5,21 +5,32 @@ import {
...
   BlockchainClientError,
-} from "./client";
-import * as bitcoind from "./bitcoind";
-import * as wallet from "./wallet";
+} from "../client";
+import * as bitcoind from "../bitcoind";
+import * as wallet from "../wallet";
 import BigNumber from "bignumber.js";
 
+import {
```
bucko13, line 13: this shouldn't be needed with the globals. You might need to set something up in eslint as well?

bucko13 (changes requested):
Nice! Glad it's passing now. Just a couple other small things (which should be reusable for the next rounds/packages!

vitest.config.ts
DhairyaMajmudar: Thank you for the suggestions @bucko13, I've created `vitest-config` package : )

packages/caravan-clients/src/__tests__/client.test.ts
DhairyaMajmudar: Yep, I've configured eslint a bit to improve the import orders

bucko13 (changes requested):
So I started to look at some turborepo examples to reference because I was a little confused since you're not actually using the config anywhere as far as I can see (it's not extended or imported for example). And it looks like maybe it's overkill to do it in this way. 

[Here](https://github.com/agarun/turborepo-vite-starter/blob/main/vitest.config.ts) is an example from turborepo. 

What I was thinking about though you can see in [this](https://github.com/vercel/turborepo/blob/main/examples/with-prisma/apps/web/.eslintrc.js) example from turborepo for eslint where there is a base package and then it's imported and extended in sub packages as they're needed. 

Perhaps that's not needed though and everything can share the same base vitest config at the root. 

Sorry for the back and forth on this! Just wanting to make sure we get it right while we're in the process of refactoring.

bucko13:
Also, because you applied the import order in the base eslint, then all other packages inherit that rule, and now the lint step is [failing](https://github.com/caravan-bitcoin/caravan/actions/runs/10628039530/job/29704473243?pr=119). Either we should do those one at a time in the individual child configs and then can move that to the parent when they're all done, or don't add that rule yet and do that in a separate PR separate from the testing changes. 

I do like that rule and would like to add it, but I think it increases the scope of these efforts to upgrade testing (many more files in the diff!)

DhairyaMajmudar:
Got your point here @bucko13  https://github.com/caravan-bitcoin/caravan/pull/119#pullrequestreview-2281759260 let me revert back the changes from the last commit and get back the vitest config file in the root itself.

As of the comment here https://github.com/caravan-bitcoin/caravan/pull/119#issuecomment-2330512515 yeah, it will be a good decision to open a separate PR for correcting the import orders from the codebase (Let me open one once this PR is merged)

packages/vitest-config/vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "vite";
+import react from "@vitejs/plugin-react";
+
+export default defineConfig({
+  plugins: [react()],
+  test: {
+    include: ["./**/*.test.ts"],
```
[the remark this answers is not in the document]
DhairyaMajmudar: Nope, it will not be required since there's a wild card `**` which will look into all subdirectories with files having extension `.test.ts`

bucko13 (approved):

PR #91, opened 2024-05-03
Spend to Taproot

bucko13:
Includes a few fixes:

- don't polyfill process env for trezor environment variables in @caravan/wallets
- cleanup some dependencies in @caravan/wallets
- remove unexported PSBTv0 modules from @caravan/psbt
- work to support bitcoinjs-lib v6 in @caravan/psbt
- Adding support to @caravan/psbt for taproot address outputs

TODO:

- [x] use new tx creation from @caravan/psbt to support paying tr addresses
- [x] new tooling for validateMultisigSignature which extracts the matching public key (?)
- [x] Fix failing tests
- [ ] More changesets documenting the changes 
- [x] Coordinator should have regtest support
- [x] Might need more robust address validation

NOTE: the diff looks a little bigger than it is because the tiny-secp256k1 library was vendor-ized. See updated readme in @caravan/psbt for more details

PR #120, opened 2024-08-28
fix: multisig support for indirect keystore interactions

chadchapnick:
Fixes a regression in which PSBTs could not be signed in indirect keystore interactions due to a missing redeem script. 

[Introduced](https://github.com/caravan-bitcoin/caravan/commit/0d81717fade918ec337093e3dc4c3862662d20c3#diff-12369fc70a2ad8c599734fefe86246c867a701f3b5bda04eb964e56b822c2a25R137) in #91

### DIFF
--- a/.changeset/thirty-dolls-type.md
+++ b/.changeset/thirty-dolls-type.md
@@ -1,0 +1,5 @@
+---
+"@caravan/psbt": patch
+---
+
+Restore multisig support for indirect keystore interactions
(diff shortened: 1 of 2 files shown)


bucko13:
Changes look good. needs a [changeset](https://github.com/caravan-bitcoin/caravan/pull/120#issuecomment-2314279514) to indicate a patch update (I think that's sufficient for this fix)

dylan-thompson:
Nice catch! I verified that this fixes signing on coldcard devices

.changeset/twenty-spies-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@caravan/psbt": major
```
bucko13, line 2: I think we can make this a patch. It's not backwards incompatible in any way and isn't a new feature. Just an (important) bug fix.

bucko13 (approved):

PR #124, opened 2024-09-02
Tweaks to SegWit PSBT Inputs/Outputs

chadchapnick:
Fixes some segwit psbt formatting:
- Include non witness utxo for inputs
- include witness script for outputs


packages/caravan-psbt/src/psbtv0/utils.ts
```diff
@@ -143,7 +141,7 @@ export const convertLegacyOutput = (output: LegacyOutput): PsbtOutput => {
     value: new BigNumber(output.amountSats).toNumber(),
     bip32Derivation:
       output.bip32Derivation || getBip32Derivation(output.multisig),
-    witnessScript: output.witnessScript || output.multisig?.witness?.output,
+    witnessScript: output.witnessScript || output.multisig?.witness?.[0],
```
chadchapnick, line 144: Bit of a hack—`bitcoinjs-lib` payments api outputs an array of witness buffers. Just extracting the first element

packages/caravan-bitcoin/src/multisig.ts
```diff
@@ -115,10 +115,16 @@ export function generateMultisigFromRaw(addressType, multisig) {
...
+        redeem: payments.p2wsh({
+          redeem: multisig,
+          witness: [multisig.output],
+        }),
       });
     case P2WSH:
-      return payments.p2wsh({ redeem: multisig });
+      return payments.p2wsh({
+        redeem: multisig,
+        witness: [multisig.output],
```
chadchapnick, line 126: This resulted in witness script being added to UTXO, but perhaps there's a more appropriate package in which to do this than in `caravan-bitcoin`?

packages/caravan-psbt/src/psbtv0/psbt.ts
```diff
@@ -106,10 +106,11 @@ const psbtInputFormatter = (
 ) => {
   const tx = Transaction.fromHex(input.transactionHex);
   const inputData: any = { ...input };
+  const nonWitnessUtxo = tx.toBuffer();
```
bucko13, line 109: good catch!

packages/caravan-bitcoin/src/multisig.ts
bucko13: Interesting. Yeah would be nice to keep this fix for the newer code/packages. This means that either we're still using a legacy caravan/bitcoin function for the psbt generation (or its constituent parts), OR one of the newer functions is still using some old code that still has the bug. 

Tracking down where this is used, `generateMultisigFromRaw` is called from `generateMultisigFromPublicKeys` and `generateMultisigFromHex`. The latter is used in a newer version of `getUnchainedInputsFromPSBT` which is used in `translatePSBT` which I only see being used in the legacy ledger and trezor interactions and `generateMultisigFromPublicKeys` doesn't seem to be really used anywhere relevant either... 

it's also used in `multisigScriptSig` which is used in `signedMultisigTransaction` which... isn't used anywhere. 

So I'm curious why this change was even necessary to have an impact.

packages/caravan-psbt/src/psbtv0/utils.ts
bucko13: This one has me puzzled. Partly just because of all our legacy code and how this data gets passed around. Where is this object being created by bitcoinjs-lib in the first place? 

Second question is about the type. Was the type wrong before with a `witness.output`? And then for this buffer selection, is this saying that the witness could have multiple buffers? I want to make sure we don't accidentally end up just taking the first byte off of a single buffer. If this is actually coming from bitcoinjs-lib maybe we can type LegacyMultisig more explicitly using bitcoinjs-lib types to make this a little safer.

packages/caravan-psbt/src/psbtv0/utils.ts
```diff
@@ -143,8 +143,9 @@ export const convertLegacyOutput = (output: LegacyOutput): PsbtOutput => {
     value: new BigNumber(output.amountSats).toNumber(),
     bip32Derivation:
       output.bip32Derivation || getBip32Derivation(output.multisig),
-    witnessScript: output.witnessScript || output.multisig?.witness?.output,
-    redeemScript: output.redeemScript || output.multisig?.redeem?.output,
+    witnessScript: output.witnessScript,
+    redeemScript: output.redeemScript,
```
bucko13, line 147: I think these are not necessary anymore right? they should come from `psbtMultisigLock`.

packages/caravan-psbt/src/psbtv0/utils.ts
```diff
@@ -143,8 +143,9 @@ export const convertLegacyOutput = (output: LegacyOutput): PsbtOutput => {
...
+    ...psbtMultisigLock(output.multisig),
```
bucko13, line 148: We'll need to make sure to update this function for basically the same chnage you did for `psbtInputFormatter`.

bucko13 (commented):
Great! So it looks like the fixes are: 
- update `psbtInputFormatter` to include nonWitnessUtxo on segwit inputs
- update `convertLegacyOutput` to use an updated legacy utility `psbtMultisigLock` which will make sure to get the redeem and witness scripts from the output correctly. (which incidentally is what `convertLegacyInput` does.
- Update to set the psbt in the state so they can be passed to signing interactions in the coordinator. 

Also need to update the psbt fixtures since our expectations around what these will look like will have changed.

packages/caravan-psbt/src/psbtv0/utils.ts
chadchapnick: If that's safe to do then yes! We can extract them from the `multisig` property—can update the fixtures such that this is handled smoothly

packages/caravan-psbt/src/psbtv0/utils.ts
chadchapnick: A bit confused on this—does this suggest we should have the following in `psbtMultisigLock`:
```js
    case P2WSH:
    case P2SH_P2WSH: // need both
      multisigLock.witnessScript = multisigWitnessScript(multisig).output;
      multisigLock.redeemScript = multisigRedeemScript(multisig)?.output;
      break;
```

Noting the optional `redeem` output, since `multisigRedeemScript` returns null for p2wsh

bucko13: Oh yeah, you know what. I think you're right. The description of the function is:

```
/**
 * Return the locking script for the given `Multisig` object in a PSBT consumable format
 */
```
and the only locking script that matters for P2WSH is the witnessScript, so I think you're right. This is fine.

chadchapnick: Thanks for verifying—understanding now that this is part of the BIP and not just an implementation detail!

packages/caravan-bitcoin/src/fixtures.ts
```diff
@@ -1162,9 +1162,9 @@ const MULTISIGS_BASE = [
...
...98Uq4iBgMotXwvZcmO183kvKVMw6E6+k1HEX/Z2uBmY6QWngXvhhwAAAAFMAAAgAAAAIBkAACAAQAAgAAAAAAAAAAAIgYDQpl/b81/pKPH4pDIhnFImS5hlHQhIJhcZk2eIURhr3wc9X7GXTAAAIAAAACAZAAAgAEAAIAAAAAAAAAAAAABAPcCAAAAAAEB1bu9SV5b/QDfMtnimjtFZmVfQfBwx609+WvkNuBkv1sAAAAAFxYAFA4CHBwzE++ZGs0MXZ9jyhLOEQAF/v///wKr1gEAAAAAABepFF0n21jxPUhRJgF02dyerN9iB0pOhxAnAAAAAAAAF6kUGrz4zqMhyodN5L60+XUHf+hkpUqHAkcwRAIgd7Tw8tFIBEP4sp7Wiim2Brg6PrvaQ0m/gTwzMwt/SqkCICBhgj/SUzC2wK22ivzkVGraGT62OYLyinVivGBeprDlASECkwp7im/VGozjYDm6zxYCOc+328ctwaBRqqwzO8f455JOSQkAAQEgECcAAAAAAAAXqRQavPjOoyHKh03kvrT5dQd/6GSlSocBBCIAIASdbpRQdFJbA+BId1k2j/Zj8Qu4iXYBe92dPM6EkIXlAQVHUiEDKLV8L2XJjtfN5LylTMOhOvpNRxF/2drgZmOkFp4F74YhA0KZf2/Nf6Sjx+KQyIZxSJkuYZR0ISCYXGZNniFEYa98Uq4iBgMotXwvZcmO183kvKVMw6E6+k1HEX/Z2uBmY6QWngXvhhwAAAAFMAAAgAAAAIBkAACAAQAAgAAAAAAAAAAAIgYDQpl/b81/pKPH4pDIhnFImS5hlHQhIJhcZk2eIURhr3wc9X7GXTAAAIAAAACAZAAAgAEAAIAAAAAAAAAAAAAAAQAiACBYHpgGYQrByCSdw0RuAfg2Xs/FBkYjedaUi6oNEoxRmgEBR1IhAnKDT2UkRJEOHMFXuZ9S/WBSgmoR+QmYeaVUkPa4GQvnIQKZl/nHCUiB0kGxOxTYVnze3KzNQRhExV/gmwuMobJUsVKuIgICcoNPZSREkQ4cwVe5n1L9YFKCahH5CZh5pVSQ9rgZC+ccAAAABTAAAIAAAACAZAAAgAEAAIABAAAAAAAAACICApmX+ccJSIHSQbE7FNhWfN7crM1BGETFX+CbC4yhslSxHPV+xl0wAACAAAAAgGQAAIABAACAAQAAAAAAAAAA",
```
bucko13, line 1165: That's a big 'un!
Screenshot 2024-09-08 at 1 00 37 PM

Screenshot 2024-09-08 at 1 00 50 PM

packages/caravan-bitcoin/src/fixtures.ts
```diff
@@ -1162,9 +1162,9 @@ const MULTISIGS_BASE = [
...
...KuIgYDKLV8L2XJjtfN5LylTMOhOvpNRxF/2drgZmOkFp4F74YcAAAABTAAAIAAAACAZAAAgAEAAIAAAAAAAAAAACIGA0KZf2/Nf6Sjx+KQyIZxSJkuYZR0ISCYXGZNniFEYa98HPV+xl0wAACAAAAAgGQAAIABAACAAAAAAAAAAAAAAQD3AgAAAAABAdW7vUleW/0A3zLZ4po7RWZlX0HwcMetPflr5DbgZL9bAAAAABcWABQOAhwcMxPvmRrNDF2fY8oSzhEABf7///8Cq9YBAAAAAAAXqRRdJ9tY8T1IUSYBdNncnqzfYgdKTocQJwAAAAAAABepFBq8+M6jIcqHTeS+tPl1B3/oZKVKhwJHMEQCIHe08PLRSARD+LKe1ooptga4Oj672kNJv4E8MzMLf0qpAiAgYYI/0lMwtsCttor85FRq2hk+tjmC8op1YrxgXqaw5QEhApMKe4pv1RqM42A5us8WAjnPt9vHLcGgUaqsMzvH+OeSTkkJAAEBIBAnAAAAAAAAF6kUGrz4zqMhyodN5L60+XUHf+hkpUqHAQQiACAEnW6UUHRSWwPgSHdZNo/2Y/ELuIl2AXvdnTzOhJCF5QEFR1IhAyi1fC9lyY7XzeS8pUzDoTr6TUcRf9na4GZjpBaeBe+GIQNCmX9vzX+ko8fikMiGcUiZLmGUdCEgmFxmTZ4hRGGvfFKuIgYDKLV8L2XJjtfN5LylTMOhOvpNRxF/2drgZmOkFp4F74YcAAAABTAAAIAAAACAZAAAgAEAAIAAAAAAAAAAACIGA0KZf2/Nf6Sjx+KQyIZxSJkuYZR0ISCYXGZNniFEYa98HPV+xl0wAACAAAAAgGQAAIABAACAAAAAAAAAAAAAAAEAIgAgWB6YBmEKwcgkncNEbgH4Nl7PxQZGI3nWlIuqDRKMUZoBAUdSIQJyg09lJESRDhzBV7mfUv1gUoJqEfkJmHmlVJD2uBkL5yECmZf5xwlIgdJBsTsU2FZ83tyszUEYRMVf4JsLjKGyVLFSriICAnKDT2UkRJEOHMFXuZ9S/WBSgmoR+QmYeaVUkPa4GQvnHAAAAAUwAACAAAAAgGQAAIABAACAAQAAAAAAAAAiAgKZl/nHCUiB0kGxOxTYVnze3KzNQRhExV/gmwuMobJUsRz1fsZdMAAAgAAAAIBkAACAAQAAgAEAAAAAAAAAAA==",
```
bucko13, line 1167: Screenshot 2024-09-08 at 1 01 54 PM
Screenshot 2024-09-08 at 1 02 19 PM

packages/caravan-bitcoin/src/fixtures.ts
```diff
@@ -1272,9 +1272,9 @@ const MULTISIGS_BASE = [
...
...jltEC/mwze3UJ2JtgfMmBbIQNp50/JVDVbb3rPm77FuGHBhoUrdZqF+SVY5CCgICBH9FKuIgYC4htzGM+9SCvbtmRBQguQGOW0QL+bDN7dQnYm2B8yYFscAAAABjAAAIAAAACAZAAAgAIAAIAAAAAAAAAAACIGA2nnT8lUNVtves+bvsW4YcGGhSt1moX5JVjkIKAgIEf0HPV+xl0wAACAAAAAgGQAAIACAACAAAAAAAAAAAAAAQD9AQECAAAAAAEBMXu/sYgUp9sf0V5Zb73IsiC3byq95XwFUCQz4kjl0MUAAAAAFxYAFH6I+yvHQNXv/++Lb1PV62i4Mxe2/v///wIQJwAAAAAAACIAIEl7Amw9NUejDm2ABuOF4DZq9eyitbRV2Hg4dZQeXH+p2KUBAAAAAAAWABRN3+39gj7l3JU/R5KoVbJiqF7XfAJHMEQCIC/INHqIAfu9E9tb0mlM2rssNCi2z6bt2gL/FrxsrLa4AiBUYE0HUQ7eUAGX3u/p05q+J3+5eoN65A6oFbaqOPAylwEhAkc0jAP1lrUnMEKeqJ30s9dbIZ/Twwy8tm2uBHeeDebSNkkJAAEBKxAnAAAAAAAAIgAgSXsCbD01R6MObYAG44XgNmr17KK1tFXYeDh1lB5cf6kBBUdSIQLiG3MYz71IK9u2ZEFCC5AY5bRAv5sM3t1CdibYHzJgWyEDaedPyVQ1W296z5u+xbhhwYaFK3WahfklWOQgoCAgR/RSriIGAuIbcxjPvUgr27ZkQUILkBjltEC/mwze3UJ2JtgfMmBbHAAAAAYwAACAAAAAgGQAAIACAACAAAAAAAAAAAAiBgNp50/JVDVbb3rPm77FuGHBhoUrdZqF+SVY5CCgICBH9Bz1fsZdMAAAgAAAAIBkAACAAgAAgAAAAAAAAAAAAAABAUdSIQMYpgr8mqfjGR+eSQUccHZPdojZpWezJ8CV5aP3X4hH5yEDi14pXxT2qEj1W31i793vRnncNh7NpfbNB0zCSuIXRRBSriICAximCvyap+MZH55JBRxwdk92iNmlZ7MnwJXlo/dfiEfnHAAAAAYwAACAAAAAgGQAAIACAACAAQAAAAAAAAAiAgOLXilfFPaoSPVbfWLv3e9Gedw2Hs2l9s0HTMJK4hdFEBz1fsZdMAAAgAAAAIBkAACAAgAAgAEAAAAAAAAAAA==",
```
bucko13, line 1275: Screenshot 2024-09-08 at 1 05 18 PM
Screenshot 2024-09-08 at 1 05 46 PM

packages/caravan-bitcoin/src/fixtures.ts
```diff
@@ -1272,9 +1272,9 @@ const MULTISIGS_BASE = [
...
...AY5bRAv5sM3t1CdibYHzJgWyEDaedPyVQ1W296z5u+xbhhwYaFK3WahfklWOQgoCAgR/RSriIGAuIbcxjPvUgr27ZkQUILkBjltEC/mwze3UJ2JtgfMmBbHAAAAAYwAACAAAAAgGQAAIACAACAAAAAAAAAAAAiBgNp50/JVDVbb3rPm77FuGHBhoUrdZqF+SVY5CCgICBH9Bz1fsZdMAAAgAAAAIBkAACAAgAAgAAAAAAAAAAAAAEA/QEBAgAAAAABATF7v7GIFKfbH9FeWW+9yLIgt28qveV8BVAkM+JI5dDFAAAAABcWABR+iPsrx0DV7//vi29T1etouDMXtv7///8CECcAAAAAAAAiACBJewJsPTVHow5tgAbjheA2avXsorW0Vdh4OHWUHlx/qdilAQAAAAAAFgAUTd/t/YI+5dyVP0eSqFWyYqhe13wCRzBEAiAvyDR6iAH7vRPbW9JpTNq7LDQots+m7doC/xa8bKy2uAIgVGBNB1EO3lABl97v6dOavid/uXqDeuQOqBW2qjjwMpcBIQJHNIwD9Za1JzBCnqid9LPXWyGf08MMvLZtrgR3ng3m0jZJCQABASsQJwAAAAAAACIAIEl7Amw9NUejDm2ABuOF4DZq9eyitbRV2Hg4dZQeXH+pAQVHUiEC4htzGM+9SCvbtmRBQguQGOW0QL+bDN7dQnYm2B8yYFshA2nnT8lUNVtves+bvsW4YcGGhSt1moX5JVjkIKAgIEf0Uq4iBgLiG3MYz71IK9u2ZEFCC5AY5bRAv5sM3t1CdibYHzJgWxwAAAAGMAAAgAAAAIBkAACAAgAAgAAAAAAAAAAAIgYDaedPyVQ1W296z5u+xbhhwYaFK3WahfklWOQgoCAgR/Qc9X7GXTAAAIAAAACAZAAAgAIAAIAAAAAAAAAAAAAAAQFHUiEDGKYK/Jqn4xkfnkkFHHB2T3aI2aVnsyfAleWj91+IR+chA4teKV8U9qhI9Vt9Yu/d70Z53DYezaX2zQdMwkriF0UQUq4iAgMYpgr8mqfjGR+eSQUccHZPdojZpWezJ8CV5aP3X4hH5xwAAAAGMAAAgAAAAIBkAACAAgAAgAEAAAAAAAAAIgIDi14pXxT2qEj1W31i793vRnncNh7NpfbNB0zCSuIXRRAc9X7GXTAAAIAAAACAZAAAgAIAAIABAAAAAAAAAAA=",
```
bucko13, line 1277: Screenshot 2024-09-08 at 1 06 18 PM
Screenshot 2024-09-08 at 1 06 42 PM

packages/caravan-bitcoin/src/fixtures.ts
```diff
@@ -1312,6 +1312,20 @@ const MULTISIGS = MULTISIGS_BASE.map((test) => {
           outputs: test.transaction.outputs.map((output) => ({
             ...output,
             ...{ amountSats: new BigNumber(output.amountSats).toString() },
+            ...{
+              multisig: (output.redeemScript || output.witnessScript) && {
```
bucko13, line 1316: nice

bucko13 (approved):
very nice. just did a quick sanity check on the psbts through bip14.org

packages/caravan-bitcoin/src/fixtures.ts
chadchapnick: What bip174.org does not reveal is the global xpubs!

bucko13: True! It also doesn't support psbt v2 :(. 

Something I've been thinking about for a while is it would be cool to do a "caravan-native" psbt explorer that was just using our code and plug some of these holes 😁

PR #132, opened 2024-09-10
unsignedPsbt in redux store should include global xpubs

bucko13:
We were storing the psbt in the state so that interactions could use if available (which was the right thing to do) but we were saving a version of the psbt without the global xpubs which is the wrong thing to do since some signers need this info.


chadchapnick (approved):
LGTM—[relevant](https://github.com/caravan-bitcoin/caravan/pull/124/files#diff-98dcda4c9fa2f5e6b7abcf392af80ca252711f72c61e237f9bf1aef06ebc6ba5R309) recent change

humanumbrella (approved):
approved

PR #133, opened 2024-09-12
[Test]: Jest to Vitest migration `caravan-psbt` package

DhairyaMajmudar:
### Screenshot

image

cc: @bucko13 @Legend101Zz


DhairyaMajmudar:
Out of 4 test 3 tests are passing but 1 test from `psbt.test.ts` file is failing below are the error logs

<details>
<summary>Error Code text</summary>
<br>

```bash

src/__tests__/psbtv0/psbt.test.ts [ src/__tests__/psbtv0/psbt.test.ts ]
Error: ecc library invalid
   assert ../../node_modules/bitcoinjs-lib-v6/src/ecc_lib.js:70:20
    verifyEcc ../../node_modules/bitcoinjs-lib-v6/src/ecc_lib.js:27:3
    Proxy.initEccLib ../../node_modules/bitcoinjs-lib-v6/src/ecc_lib.js:11:5
    src/psbtv0/psbt.ts:27:9
     25| import { autoLoadPSBT } from "./utils";
     26| 
     27| bitcoin.initEccLib(ecc);
       |         ^
     28| 
     29| export interface PsbtInput {
   src/__tests__/psbtv0/psbt.test.ts:2:31```
</details>


![image](https://github.com/user-attachments/assets/a92d0373-1b44-4fa9-972f-7082e742ea8b)

**Comment by DhairyaMajmudar** [comment]
While finding the fix for this I found a similar kind of issue here https://github.com/bitcoinjs/tiny-secp256k1/issues/136
@bucko13 can you pls. help in  this

**Comment by bucko13** [comment]
> Out of 4 test 3, tests are passing but 1 test from `psbt.test.ts` file is failing below are the error logs
> 
> Screenshot from 2024-09-12 22-37-53
> 
> Error Code text

I think I documented some of this including changes I had to make for jest in the readme of the package itself. See if anything there helps, otherwise I can take a further look.

**Comment by DhairyaMajmudar** [comment]
Thank you @bucko13, I've found the fix from the readme
Now all tests are passing : )

File: vitest.config.ts [hunk 1]
```diff
@@ -6,6 +6,6 @@ export default defineConfig({
   test: {
     include: ["./**/*.test.ts"],
     globals: true,
-    environment: "jsdom",
+    environment: "node",
   },
```
bucko13, line 10: Ah. This is where I wonder if we’ll run into problems not having different vitest configs per sub-project

DhairyaMajmudar: As of now its all fine I've checked `caravan-client` and its running correctly with same configurations 
image

DhairyaMajmudar: But can cause a problem in `caravan-wallet` package since there in some places `window` object is used which is not supported in node environment

DhairyaMajmudar: Got a solution for this : )
https://vitest.dev/guide/environment#environments-for-specific-files

DhairyaMajmudar: I've found a fix for this from vitest docs, we can define environment for specific files as well  :)
https://vitest.dev/guide/environment#environments-for-specific-files

bucko13: Oh nice. Maybe we can add this in some documentation somewhere so we don't lose track of it for the future? Maybe a "testing" section in the main readme?

packages/caravan-psbt/package.json
bucko13, line 1: ❤️

packages/caravan-psbt/src/__tests__/functions.test.ts
```diff
@@ -1,4 +1,5 @@
-import { bufferize } from "./functions";
+import { bufferize } from "../functions";
+import { describe, it, expect } from "vitest";
```
bucko13, line 2: Shouldn't this not be necessary if the globals are setup properly?

packages/caravan-psbt/package.json
```diff
@@ -54,18 +54,15 @@
...
-    "@inrupt/jest-jsdom-polyfills": "^3.2.1",
-    "@jest/globals": "^29.7.0",
-    "@types/jest": "^29.5.12",
+    "@vitejs/plugin-react": "^4.3.1",
     "eslint-plugin-prettier": "^5.1.3",
-    "jest": "^29.4.1",
     "jsdom": "24.0.0",
     "jsdom-global": "3.0.2",
     "lodash": "^4.17.21",
     "react-silence": "^1.0.4",
```
bucko13, line 62: I think you replaced the uses of this right? so can probably uninstall.

packages/caravan-psbt/tsconfig.json
```diff
@@ -2,14 +2,9 @@
   "extends": "@caravan/typescript-config/base.json",
   "compilerOptions": {
     "baseUrl": ".",
-    "types": [
-      "node",
-      "jest"
-    ],
+    "types": ["node", "vitest/globals"],
```
bucko13, line 5: Shouldn't this make the imports unnecessary?

bucko13 (changes requested):
Mostly looks good with some good dependency and configuration cleanup!

I have a question about the globals and if we can avoid the need for the test tooling imports and also adding some documentation for how to handle different js runtime environments.

packages/caravan-psbt/package.json
DhairyaMajmudar: Yeah, I haven't uninstalled them completely since those modules are shared in other packages, if uninstalled the test suite there will fail. 

I'll remove those unwanted modules once every package gets migrated to vitest : )

packages/caravan-psbt/src/__tests__/functions.test.ts
DhairyaMajmudar: Yeah correct, thank you suggestions I've removed those unnecessary imports

packages/caravan-psbt/tsconfig.json
DhairyaMajmudar: I've removed the imports

vitest.config.ts
DhairyaMajmudar: sounds cool, let me edit the readme file : )

DhairyaMajmudar:
> Mostly looks good with some good dependency and configuration cleanup!

That's amazing thank you @bucko13 

> I have a question about the globals and if we can avoid the need for the test tooling imports 

Yeah, I've removed the unwanted  imports from both `caravan-psbt` as well as `caravan-client`

> adding some documentation for how to handle different js runtime environments.

I've modified the readme section by adding a new testing section there

packages/caravan-psbt/package.json
bucko13: Huh. That shouldn’t be the case with the monorepo since they should be isolated. Any package that uses this should have it in its own package.json

DhairyaMajmudar: Yeah, every package has its own package.json but the node_modules and package-lock.json files are shared among them

bucko13: Right so you should be able to remove it from THIS package.json, re-run npm install to update the package lock accordingly and then everything else should still worn

DhairyaMajmudar: I've re-run the `npm i` command and pushed the changes in `package-lock.json` file in my latest commit. I've also checked if the tests are running fine and they are.

bucko13: Right I don’t imagine anything has changed. It looks like react-silence is still in this packages package.json. I think you can remove it and then re-run everything and it should be fine. If it’s not working then that means another package needs to be explicit about the react silence dev dependency.

DhairyaMajmudar: I've removed react-silence from package.json of psbt but after doing so the tests in caravan-bitcoin were failing so I've installed the same in its own package.json file

packages/caravan-psbt/package.json
```diff
@@ -54,18 +54,14 @@
     "@caravan/eslint-config": "*",
     "@caravan/multisig": "*",
     "@caravan/typescript-config": "*",
-    "@inrupt/jest-jsdom-polyfills": "^3.2.1",
-    "@jest/globals": "^29.7.0",
-    "@types/jest": "^29.5.12",
+    "@vitejs/plugin-react": "^4.3.1",
```
bucko13, line 57: why was this necessary?

DhairyaMajmudar: I had moved that dependency to the root package.json file as of which I've removed it from here

PR #116, opened 2024-08-06
Implement RBF and CPFP with new fees package and add pending transactions view in @caravan/coordinator

Legend101Zz:
This commit lays foundation of introduction of significant enhancements to be done to the transaction management capabilities of caravan/coordinator, focusing on Replace-By-Fee (RBF) and Child-Pays-For-Parent (CPFP) functionalities, along with a new pending transactions view.

Key changes Expected:

1. Pending Transactions View:
   - Add new component 'PendingTransactionsView' in src/components/Wallet/
   - Implement basic layout showing transaction ID, amount, recipient, and timestamp
   - Integrate view into main Wallet dashboard

2. Enhanced Transaction Information:
   - Add 'TransactionDetails' component to display comprehensive tx info
   - Implement time-pending calculation and display
   - Add current fee rate vs market rate comparison
   - Include estimated blocks until confirmation based on current mempool state

3. Fee Bumping Possibility Indicators:
   - Add RBF status indicator to transaction details
   - Implement change output detection and display
   - Calculate and show estimated cost for potential fee bump

4. UI Support for Manual RBF:
   - Modify existing transaction creation flow to support RBF
   - Add RBF toggle in transaction creation form (OutputsForm.jsx)
   - Implement logic to replace existing transaction when RBF is selected

5. Fee Bumping Interfaces:
   - Add 'CancelTransaction' button and associated logic
   - Implement 'AccelerateTransaction' functionality
   - Create 'DownloadPSBT' option for RBF/CPFP transactions
   - Modify transaction authoring page to support pre-filled RBF/CPFP data

6. Integration with new fees package:
   - Import and utilize functions from @caravan/fees for fee calculations
   - Implement FeeManager class to handle fee-related operations

7. State Management Updates:
   - Add new reducers for managing pending transactions and fee bump operations
   - Implement new actions for RBF and CPFP operations
   - Update selectors to support new state structure

PR #134, opened 2024-09-12
feat: Add getTransaction method to BlockchainClient

Legend101Zz:
This commit adds a getTransaction method to the BlockchainClient class, which provides a consistent structure for fetching transaction details across different data providers (Bitcoin Core, Blockstream, Mempool).

Changes:
 - Add `TransactionDetails` interface for unified transaction data
 - Implement `getTransaction` method
 - Add `normalizeTransactionData` to standardize response


Legend101Zz:
#116  This method was added to help add fee-bumping methods RBF/CPFP to caravan-coordinator

Legend101Zz:
#116  This method was added to help support adding RBF and CPFP fee-bumping methods to caravan/coordinator

Legend101Zz (commented):
Does this fn looks good to you ?
Because I noticed that `getTransactionHex` does not have typed return type

packages/caravan-clients/src/client.ts
```diff
@@ -454,6 +479,33 @@ export class BlockchainClient extends ClientBase {
...
+      let txData: any;
+
+      if (this.type === ClientType.PRIVATE) {
+        const response = await callBitcoind(
+          this.bitcoindParams.url,
+          this.bitcoindParams.auth,
+          "getrawtransaction",
+          [txid, true],
+        );
+        txData = response.result;
```
bucko13, line 493: Move this into a function in the bitcoind file (or wallet.ts? it depends on if you want this to require tx indexing to be on). Also, note that there already is a `bitcoindRawTxData` function exported from there. I think that does what you need, except it looks like it's probably broken. I think it was moved from a class and still references `this`. and that file is still js and not ts so not typed. Maybe you could clean those up and use that as a part of this PR?

bucko13 (changes requested):
A couple requests. Overall looks good though. Thanks and sorry for the delay in review!

packages/caravan-clients/src/client.ts
```diff
@@ -34,6 +34,31 @@ export enum ClientType {
   MEMPOOL = "mempool",
 }
 
+interface TransactionDetails {
```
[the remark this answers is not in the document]
Legend101Zz: got it ... initally I thought that other interfaces and types , also did not have a particular type definition file .. so restricted myself from creating one .... 
Should I clean up other files too by putting all the types in a single file ?

packages/caravan-clients/src/client.ts
Legend101Zz: do you want me to convert the whole bitcoind.js file to ts ??

packages/caravan-clients/src/client.ts
```diff
@@ -486,4 +538,34 @@ export class BlockchainClient extends ClientBase {
 
     return await bitcoindWalletInfo({ ...this.bitcoindParams });
   }
+
+  private normalizeTransactionData(txData: any): TransactionDetails {
```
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/d193089442116b617960ff5a49cb089da48aac8b#diff-4fd3f89188bb2c009b1c8451e8ccd94db7f894cbc6456d4e1c73a83327b42287R522)

packages/caravan-clients/src/client.ts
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/d193089442116b617960ff5a49cb089da48aac8b#diff-36fedc7d8a3d224383353cd1053c2fb3626ffaacc71564dd703a059dc617e54d)

Please let me know if you want to have all the types added to the type definition file from the whole package and add some doc-strings to them .. :)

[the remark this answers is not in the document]
bucko13: Definitely would be nice but not a requirement for this PR imo!

packages/caravan-clients/src/client.ts
bucko13: More ts refactoring is always desirable! It's not a blocker imo for this PR but would be more than welcome if you wanted to take that on.

packages/caravan-clients/src/client.ts
```diff
@@ -486,4 +518,36 @@ export class BlockchainClient extends ClientBase {
 
     return await bitcoindWalletInfo({ ...this.bitcoindParams });
   }
+
+  private normalizeTransactionData(
```
bucko13, line 522: I'd recommend that this should exist on its own rather than as a class method on the client class. This is basically a utility method that we don't need an instantiated client in order to use. Moving it out would also mean it could be separately tested and then easily mocked in tests for the client class.

packages/caravan-clients/src/types.ts
```diff
@@ -43,3 +43,70 @@ export interface FeeRatePercentile {
...
+  locktime: number;
+  vin: Array<{
+    txid: string;
+    vout: number;
+    sequence: number;
+  }>;
+  vout: Array<{
+    value: number;
+    scriptpubkey: string;
+    scriptpubkey_address?: string;
```
bucko13, line 101: Since this is a value that we output, can we reformat the key values for these? snake case like this is because of the rpc api right? but we have an opportunity to have these conform to a more js friendly formatting unless I'm missing something.

packages/caravan-clients/src/bitcoind.js
```diff
@@ -144,16 +144,14 @@ export function bitcoindImportMulti({ url, auth, addresses, label, rescan }) {
...
-    return await callBitcoind(
-      this.bitcoindParams.url,
-      this.bitcoindParams.auth,
-      "decoderawtransaction",
-      [txid],
-    );
-  }
-  catch(e){
-    return e;
+export async function bitcoindRawTxData(url, auth, txid) {
```
bucko13, line 147: NOTE: This function is used in other methods. Can you make sure that the function there is called correctly?

bucko13 (changes requested):
A couple of relatively small changes otherwise looking good!

packages/caravan-clients/src/client.ts
Legend101Zz: same for this ... If we can wait a few days I redo this file

packages/caravan-clients/src/types.ts
Legend101Zz: Ah yes, good catch! Sorry been writing alot of rust  these days 😅

You're absolutely right - since this is our output interface we should follow JS conventions rather than the RPC snake_case. Will switch these to camelCase. 

The snake_case was leaking through from the RPC API but we can normalize that in our utility function.

packages/caravan-clients/src/client.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/402dcd07d298417886848663b2ea42708657f9d5)

packages/caravan-clients/src/types.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/da477750ea0a736390dde481a660585c585b1f83)

packages/caravan-clients/src/bitcoind.js
bucko13: Locally I'm seeing it called in getAddressTransactions [here](https://github.com/caravan-bitcoin/caravan/blob/main/packages/caravan-clients/src/client.ts#L181)

packages/caravan-clients/src/bitcoind.js
```diff
@@ -144,16 +144,14 @@ export function bitcoindImportMulti({ url, auth, addresses, label, rescan }) {
...
+  try {
+    const response = await callBitcoind(url, auth, "getrawtransaction", [
+      txid,
+      true,
+    ]);
+    return response.result;
```
bucko13, line 153: when i run this with the #168 PR I'm getting an error because callBitcoind doesn't return a result prop. if you change this, the error doesn't happen anymore.

packages/caravan-clients/src/client.ts
```diff
@@ -454,6 +501,32 @@ export class BlockchainClient extends ClientBase {
...
+    try {
+      let txData: RawTransactionData;
+
+      if (this.type === ClientType.PRIVATE) {
+        const response = await bitcoindRawTxData(
+          this.bitcoindParams.url,
+          this.bitcoindParams.auth,
+          txid,
+        );
+        txData = response.result;
```
bucko13, line 514: this also needs to just be 
```suggestion
        txData = response;
```

packages/caravan-clients/src/client.ts
```diff
@@ -38,6 +44,47 @@ const delay = () => {
...
+        clientType === ClientType.PRIVATE
+          ? output.value
+          : satoshisToBitcoins(output.value),
+      scriptPubkey: output.scriptpubkey,
+      scriptPubkeyAddress: output.scriptpubkey_address,
+    })),
+    size: txData.size,
+    weight: txData.weight,
+    fee: clientType === ClientType.PRIVATE ? txData.fee || 0 : txData.fee,
+    status: {
```
bucko13, line 79: When processing data from my local bitcoind I'm getting an error here because txData.status doesn't exist. This is what the object looks like for me:

```
{
    "txid": "d912ba4b63d0c9f3316a1dd1d8609c2b3a4e2bd70da4a4e8c3a29181e88c0d36",
    "hash": "d912ba4b63d0c9f3316a1dd1d8609c2b3a4e2bd70da4a4e8c3a29181e88c0d36",
    "version": 2,
    "size": 243,
    "vsize": 243,
    "weight": 972,
    "locktime": 140,
    "vin": [
        {
            "txid": "ff5db4428b88784f1811b80a3967dbc270b2a69e9d96a219fd0ead64d3a107c7",
            "vout": 0,
            "scriptSig": {
                "asm": "30440220462d827bcd0b1772ccf14998fb52c6552119ac758368bafcb3dcfb33bf3fad7d022042527895fa8f6076e2940d26c1270cfc2800ddbee8c39fd8bfb7176b48d4b2c4[ALL] 039d04265a33947a45f28fe1af4c77cae42ce3caebfc7fc1dbc6c599da892cc1ae",
                "hex": "4730440220462d827bcd0b1772ccf14998fb52c6552119ac758368bafcb3dcfb33bf3fad7d022042527895fa8f6076e2940d26c1270cfc2800ddbee8c39fd8bfb7176b48d4b2c40121039d04265a33947a45f28fe1af4c77cae42ce3caebfc7fc1dbc6c599da892cc1ae"
            },
            "sequence": 4294967294
        }
    ],
    "vout": [
        {
            "value": 7.99999757,
            "n": 0,
            "scriptPubKey": {
                "asm": "1 f32e4ae2ef4fff06f397a16ff0d5eade2016be8288ce5bb4ccd266e2118846f5",
                "desc": "rawtr(f32e4ae2ef4fff06f397a16ff0d5eade2016be8288ce5bb4ccd266e2118846f5)#qgxdpa8k",
                "hex": "5120f32e4ae2ef4fff06f397a16ff0d5eade2016be8288ce5bb4ccd266e2118846f5",
                "address": "bcrt1p7vhy4ch0fllsduuh59hlp402mcspd05z3r89hdxv6fnwyyvggm6stx9qpz",
                "type": "witness_v1_taproot"
            }
        },
        {
            "value": 42,
            "n": 1,
            "scriptPubKey": {
                "asm": "0 88caacd1da0ad119c9cb1b7748e779fa8b8a14134a46c9129228c9840db62467",
                "desc": "addr(bcrt1q3r92e5w6ptg3njwtrdm53emel29c59qnffrvjy5j9rycgrdky3nsma4q30)#p3w8vnrx",
                "hex": "002088caacd1da0ad119c9cb1b7748e779fa8b8a14134a46c9129228c9840db62467",
                "address": "bcrt1q3r92e5w6ptg3njwtrdm53emel29c59qnffrvjy5j9rycgrdky3nsma4q30",
                "type": "witness_v0_scripthash"
            }
        }
    ],
    "hex": "0200000001c707a1d364ad0efd19a2969d9ea6b270c2db67390ab811184f78888b42b45dff000000006a4730440220462d827bcd0b1772ccf14998fb52c6552119ac758368bafcb3dcfb33bf3fad7d022042527895fa8f6076e2940d26c1270cfc2800ddbee8c39fd8bfb7176b48d4b2c40121039d04265a33947a45f28fe1af4c77cae42ce3caebfc7fc1dbc6c599da892cc1aefeffffff020d07af2f00000000225120f32e4ae2ef4fff06f397a16ff0d5eade2016be8288ce5bb4ccd266e2118846f500ea56fa0000000022002088caacd1da0ad119c9cb1b7748e779fa8b8a14134a46c9129228c9840db624678c000000",
    "blockhash": "2a518c52fb2d4dc1207d9a6cac49d320fa8ee022496d13a103341ba2b956ab80",
    "confirmations": 6,
    "time": 1734635974,
    "blocktime": 1734635974
}
```

bucko13 (changes requested):
tx data processing from private client is not working for me. Left some comments that will hopefully help fix.

packages/caravan-clients/src/client.ts
Legend101Zz: refactored the whole bitcoind file to [ts](https://github.com/caravan-bitcoin/caravan/pull/134/commits/8f5439e780508036e365cbbefb4b500ecd528283)

packages/caravan-clients/src/bitcoind.js
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/7d4f233a77b8833aca96d444b993841430057c05)

packages/caravan-clients/src/client.ts
Legend101Zz: thanks for the fix :) 
so I change the bitcoin file to return the whole `response object` and not the `response.result object` as it made more sense and let the end consumer methods decide what they want to do with the whole object.

[here](https://github.com/caravan-bitcoin/caravan/pull/134/commits/8f5439e780508036e365cbbefb4b500ecd528283#diff-becbd4bbfbd061556d05bcab33227d25b2b48e6f34ce5cb03697b239772f3896R346)

Legend101Zz:
So I have majorly converted the whole `clients package` from js to ts ... maybe a further scope would include to clean the explicit any used to prevent the type errors ...

packages/caravan-clients/src/client.ts
bucko13: is anything except the result used though? I did a quick search around for `bitcoindRawTxData` and it looks like there are still places that are not accessing the result but before were expecting response.result. and I note in your [test](https://github.com/caravan-bitcoin/caravan/pull/134/files#diff-416e19bc7d09f1657b40c0ee25212cdb78cb020fa7f7038d639877e98bb38b9aR1259-R1261) you're only mocking out the `result` so it seems like that is actually all we care about so it might be worth switching it back.

packages/caravan-clients/src/client.ts
bucko13: makes sense to me!

packages/caravan-clients/src/block_explorer.ts
```diff
@@ -0,0 +1,176 @@
...
+ * @param network - The Bitcoin network to use (mainnet|testnet)
+ * @returns Promise resolving to an object containing the address usage status
+ * @throws Will throw if the API request fails
+ */
+export async function blockExplorerGetAddressStatus(
+  address: string,
+  network: Network,
+): Promise<AddressStatusResponse> {
+  try {
+    await throttleRequest();
```
bucko13, line 105: lost a code comment explaining the need for the delay. I think there was something lost in translation from the conversion. Previously this was a `delay` and the comment said "delay hack to prevent throttling". basically sometimes some of the explorers would start to throttle requests if they came in too quickly in rapid succession.

packages/caravan-clients/src/bitcoind.ts
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * Checks if an error from bitcoind indicates an address wasn't found in the wallet
+ *
+ * This is useful for client-side error handling, especially when dealing with
+ * watch-only wallets or addresses that might not be imported.
+ *
+ * @param error - The error object to check
+ * @returns true if the error indicates address not found in wallet
+ */
+export function isWalletAddressNotFoundError(error: unknown): boolean {
```
bucko13, line 72: this function was totally refactored from the original. not just a type change. I think it would be easier to review if there was one commit that just changed the file type, another commit that fixed type errors, and then a separate commit for refactoring changes. 

note the original:

```
export function isWalletAddressNotFoundError(e) {
  return (
    e.response &&
    e.response.data &&
    e.response.data.error &&
    e.response.data.error.code === -4
  );
}
```
which i find more readable actually

packages/caravan-clients/src/bitcoind.ts
```diff
@@ -0,0 +1,352 @@
...
+    return resp.result;
+  } catch (error) {
+    console.log("Error broadcasting transaction:", error);
+    if (error && typeof error === "object" && "response" in error) {
+      const response = error.response as {
+        data?: { error?: { message: string } };
+      };
+      throw response?.data?.error?.message || error;
+    }
+    throw error;
```
bucko13, line 267: this is weird and different from the original. I'm assuming some fix for typing because type of error is unknown? I think this can be done a little more concisely though.

packages/caravan-clients/src/bitcoind.ts
```diff
@@ -0,0 +1,352 @@
...
+  addresses: string[];
+  label?: string;
+  rescan: boolean;
+}): Promise<RPCResponse<ImportMultiResponse[]>> {
+  const imports: ImportDescriptor[] = addresses.map((address) => ({
+    scriptPubKey: {
+      address,
+    },
+    label,
+    timestamp: 0, // TODO: Consider making this configurable for better history management
```
bucko13, line 300: why did this TODO change from the previous version? 

```
TODO: better option to ensure address history is picked up?
```

bucko13 (commented):
Submitting some initial comments.

packages/caravan-clients/src/blockchain.ts
```diff
@@ -0,0 +1,201 @@
...
+  bitcoindParams,
+  isWalletAddressNotFoundError,
+} from "./bitcoind";
+
+import { BLOCK_EXPLORER, BITCOIND, ClientConfig, UTXOUpdates } from "./types";
+
+/**
+ * Type guard to check if client has required bitcoind parameters
+ */
+function isBitcoindClient(
```
bucko13, line 26: 👌

packages/caravan-clients/src/wallet.ts
```diff
@@ -32,6 +32,7 @@ export function callBitcoindWallet({
 
   if (walletName)
     url.pathname = url.pathname.replace(/\/$/, "") + `/wallet/${walletName}`;
+  //@ts-expect-error Will Fix this
```
bucko13, line 35: what's the story with these?

bucko13 (changes requested):
Looks great! just a couple of minor comments and cleanup (some optional)

.changeset/empty-ants-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@caravan/clients": minor
+---
+
+Add getTransaction method to BlockchainClient
```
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/9ff54b35ca7dc2c20a1ee7c4a751f2b064725824)

packages/caravan-clients/src/block_explorer.ts
Legend101Zz: Thanks for pointing it out! I’ve reverted all the files and followed the three-step approach: renaming commit, basic TypeScript fix commit, and proper TypeScript fix commits...

packages/caravan-clients/src/bitcoind.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/f2097e3ce1abace77efad1913f1a776ef93fee4b)

packages/caravan-clients/src/bitcoind.ts
Legend101Zz: Thanks for pointing it out! I’ve reverted all the files and followed the three-step approach: renaming commit, basic TypeScript fix commit, and proper TypeScript fix commits

packages/caravan-clients/src/bitcoind.ts
Legend101Zz: reverted this back ...

packages/caravan-clients/src/bitcoind.ts
```diff
@@ -0,0 +1,341 @@
...
+ * @param auth - Basic auth credentials
+ * @param method - The RPC method to call
+ * @param params - Parameters for the RPC method (optional)
+ * @returns Promise resolving to the RPC response
+ */
+export async function callBitcoind<T>(
+  url: string,
+  auth: AxiosBasicCredentials,
+  method: string,
+  params: unknown[] = [],
```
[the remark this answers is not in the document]
Legend101Zz: ahh ...  good call! `params` actually accepts more than just strings or numbers.

For example:

- In `bitcoindImportMulti`, it's an array with complex objects like `[imports, { rescan }]`.
- In `wallet.ts`, `bitcoindListUnspent` passes an object: `{ minconf: 0, maxconf: 9999999, addresses: addressParam `}.

Since Bitcoin RPC methods have pretty varied parameter structures, sticking with `unknown[]` makes sense. But to be even more precise, we could use:

```ts
params: unknown[] | Record<string, unknown> = []
```

packages/caravan-clients/src/wallet.ts
Legend101Zz: I described it above in detail :)

packages/caravan-clients/src/bitcoind.ts
```diff
@@ -0,0 +1,341 @@
...
+ */
+export function bitcoindImportMulti({
+  url,
+  auth,
+  addresses,
+  label,
+  rescan,
+}: {
+  url: string;
+  auth: AxiosBasicCredentials;
```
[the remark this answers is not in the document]
Legend101Zz: Thanks for the suggestion! That’s a great improvement—definitely makes the typings cleaner and more reusable. Changes [applied](https://github.com/caravan-bitcoin/caravan/pull/134/commits/07c9d544b0114f824f022bee7cde8a4ae472fcf4)! 🚀

packages/caravan-clients/src/bitcoind.ts
```diff
@@ -0,0 +1,341 @@
...
+ * @param url - Bitcoin node URL
+ * @param auth - Authentication credentials
+ * @param txid - Transaction ID to look up
+ * @returns Detailed transaction data
+ */
+export async function bitcoindRawTxData(
+  url: string,
+  auth: AxiosBasicCredentials,
+  txid: string,
+): Promise<any> {
```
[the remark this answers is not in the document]
Legend101Zz: Got it ... that makes each call in a standard format .... needed to change a couple of other files too as now `bitcoindRawTxData` expects a object ... 

changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/c4d5ad728e122885fd57180f2f2bf74962b37510)

packages/caravan-clients/src/block_explorer.ts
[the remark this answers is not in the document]
Legend101Zz: sure changes [done](https://github.com/caravan-bitcoin/caravan/pull/134/commits/0bd563de8f72be6fa725e54d12ec41ba03205168)

bucko13 (approved):

PR #137, opened 2024-09-24
Trezor Dev Settings

humanumbrella:
Adds a couple init settings if in dev mode


humanumbrella:
this isn't worth diving into any further

PR #156, opened 2024-11-18
fix: unknown network error for regtest

chadchapnick:
Checks for REGTEST network in Coldcard and Custom interactions rather than throwing "Unknown network"

### DIFF
--- a/.changeset/stupid-starfishes-taste.md
+++ b/.changeset/stupid-starfishes-taste.md
@@ -1,0 +1,5 @@
+---
+"@caravan/wallets": patch
+---
+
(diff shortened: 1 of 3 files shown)


bucko13 (approved):
I think there's an array somewhere in @caravan/bitcoin that has these (including signet) that we could just check for. but this is fine for now. Thanks!

chadchapnick:
I'm seeing at least one other location this same subset is [hardcoded](https://github.com/caravan-bitcoin/caravan//blob/main/packages/caravan-bitcoin/src/keys.ts#L133) in @carvan/bitcoin. The reusable array would be much easier to maintain!

PR #158, opened 2024-11-24
PsbtV2.toV0(): Convert a PsbtV2 class to its serialized PSBTv0 counterpart

Shadouts:
To gain the features of the `PsbtV2` class while not sacrificing the interoperability needed with systems requiring a PSBTv0, this method allows for the easy conversion of a PSBTv2 to a PSBTv0. Together with the already existing `PsbtV2.FromV0()`, the interoperability comes full circle.

The new method introduces a new PsbtV2Maps class with features that allow for creation of unsigned transactions needed for the `0x00` key type and for deleting of incompatible keys on the map.

This feature revealed a couple insufficiencies existing in current `PsbtV2.addInput` method.


packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -915,6 +915,7 @@ export class PsbtV2 extends PsbtV2Maps {
     redeemScript,
     witnessScript,
     bip32Derivation,
+    sighashType,
```
Shadouts, line 918: While not required on a psbt input, this field was required for constructing a complete, unsigned txn to match the test vectors unsigned raw txn.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -1326,6 +1333,8 @@ export class PsbtV2 extends PsbtV2Maps {
         .readInt32LE(0);
     }
 
+    psbtv2.PSBT_GLOBAL_FALLBACK_LOCKTIME = psbtv0.locktime;
```
Shadouts, line 1336: This was just missing, but resulting unsigned raw transactions constructed by `toV0` were not matching the text vector unsigned raw txn.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
```diff
@@ -1665,3 +1684,20 @@ describe("PsbtV2.setInputSequence", () => {
...
+    (vect) => {
+      const psbtv2 = PsbtV2.FromV0(vect.base64, true);
+      const psbtv0 = psbtv2.toV0("hex");
+      // At a minimum, the raw tx cannot start before string index 16 because
+      // the hex psbt must at least start with "70736274ff0100" and one byte for
+      // length.
+      expect(psbtv0.indexOf(vect.unsignedRawTx)).toBeGreaterThanOrEqual(16);
+    },
+  );
+});
```
Shadouts, line 1703: I started with a test that tried converting to psbtv2 and then back to psbtv0 and comparing the entire serialized psbt. I knew this would be brittle since the maps don't have to be in the same order. It worked mostly, but failed on a couple of the comparisons since mapped values were out of order.

After thinking about it, I came to the conclusion that is described in the first comment. However, a bug in this change, namely the lack of setting of locktime by `FromV0`, was revealed from the full psbt comparison tests.

Shadouts, line 1703: I started with a test that tried converting to psbtv2 and then back to psbtv0 and comparing the entire serialized psbt. I knew this would be brittle since the maps don't have to be in the same order. It worked mostly, but failed on a couple of the comparisons since mapped values were out of order.

After thinking about it, I came to the conclusion that is described in the first comment. However, a bug in this change, namely the lack of setting of locktime by FromV0, was revealed from the full psbt comparison tests.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
```diff
@@ -1665,3 +1684,20 @@ describe("PsbtV2.setInputSequence", () => {
...
+  // to check that the required PSBT_GLOBAL_UNSIGNED_TX was properly created.
+  test.each(BIP_174_VECTORS_VALID_PSBT)(
+    "Properly constructs PSBT_GLOBAL_UNSIGNED_TX. $case",
+    (vect) => {
+      const psbtv2 = PsbtV2.FromV0(vect.base64, true);
+      const psbtv0 = psbtv2.toV0("hex");
+      // At a minimum, the raw tx cannot start before string index 16 because
+      // the hex psbt must at least start with "70736274ff0100" and one byte for
+      // length.
+      expect(psbtv0.indexOf(vect.unsignedRawTx)).toBeGreaterThanOrEqual(16);
```
bucko13, line 1700: Would it be worthwhile to test that the psbtv0 serialized can then be ingested by the psbtv0 tooling without throwing errors?

bucko13 (commented):
This looks great. I just left a couple comments thinking about how maybe the testing could be a little more robust, but I'm not completely certain about it. So if you think that wouldn't add any valuable coverage, I think we can approve and merge.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
Shadouts: I added some more tests on 78a4380447db81eebf68bda711c2a33737b07daa which will check the unsigned txn coming from `autoLoadPsbt` which invokes bitcoinjs-lib's Psbt class.

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
```diff
@@ -110,3 +110,111 @@ export abstract class PsbtV2Maps {
...
+      this.v0delete(inputMap, KeyType.PSBT_IN_SEQUENCE);
+      this.v0delete(inputMap, KeyType.PSBT_IN_REQUIRED_TIME_LOCKTIME);
+      this.v0delete(inputMap, KeyType.PSBT_IN_REQUIRED_HEIGHT_LOCKTIME);
+    }
+
+    for (const outputMap of this.outputMaps) {
+      this.v0delete(outputMap, KeyType.PSBT_OUT_AMOUNT);
+      this.v0delete(outputMap, KeyType.PSBT_OUT_SCRIPT);
+    }
+  };
```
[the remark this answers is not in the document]
Shadouts: I don't think those tests are needed. I explain in a comment a psbt with unrecognized keytypes is still technically valid as incompatible implementations/versions would just ignore those keytypes. This is verified against bitcoinjs-lib with the new tests I've just added. That lib's `Psbt` class has a field named something like `unknownKeys`.

bucko13 (approved):

PR #159, opened 2024-12-12
Include addressType for accurate PSBT translation in SignMultisigTransaction

chadchapnick:
Adjust inherited `DirectKeystoreInteraction`s to accept an `addressType` prop for accurate PSBT translation when invoked in `SignMultisigTransaction`.


bucko13 (approved):
Good catch

Issue #161, opened 2024-12-13
Upgrade Trezor Connect to 9.4.0+


kimlwh:
Could Trezor Connect be upgraded to 9.4.7 (latest version as of today) or any version above 9.4.0?

Initializing newer versions of Trezor Connect with `TrezorConnect.init({coreMode: "auto", ...})` now allows Trezor devices to be used without any local bridge software necessary, like the other devices Caravan supports. This is especially useful on platforms and devices where only the web is available or local software isn't, this compatibility has always been Caravan's biggest selling point as a stateless coordinator.

Was there a compatibility reason as to why Trezor Connect is frozen at version 9.1.12? I believe this is a worthwhile upgrade, the only meaningful one in fact since the current version Caravan is currently using.

Thank you.

bucko13:
Glad the problem isn't that trezor connect is now broken!

This does seem like a meaningful upgrade as well! Do you have a link to more docs on this upgrade? I didn't see anything in the 9.4.7 changelog. 

> Was there a compatibility reason as to why Trezor Connect is frozen at version 9.1.12?

No reason for that version in particular. That was just the last one we explicitly tested. The reason that we pinned though is that if we used the default to use the `latest` then trezor could inadvertently brick users since you access connect JIT and whatever version is latest is what you use. This has happened before where something was broken and users couldn't sign. We pinned to an earlier version in a release and reported the bug to trezor which was quickly released, but it felt like a precarious position to maintain so pinning and choosing to upgrade (and testing first) seemed preferable. 

would you be willing to put up a PR that bumps the version? We can confirm everything still works and then upgrade.

kimlwh:
>Glad the problem isn't that trezor connect is now broken!

Yes I may have been confusing it with the development/hosted version that doesn't have cross-origin issues.

There doesn't seem to be much documentation other than the addition of the `coreMode` option which is documented under the [`init`](https://connect.trezor.io/9/methods/other/init/) method TrezorConnect and in the [changelog](https://connect.trezor.io/9/changelog/) under 9.4.0.

> would you be willing to put up a PR that bumps the version? We can confirm everything still works and then upgrade.

I will try, though I have no means of local development at the moment and little JavaScript knowledge. A couple of other changes are likely required such as removing the pinned `connectSrc`.

bucko13:
[...]

PR #163, opened 2024-12-13
Upgrade Trezor Connect to 9.4.7

kimlwh:


packages/caravan-wallets/src/trezor.ts
```diff
@@ -130,9 +130,8 @@ const TREZOR_DEV =
 
 try {
   TrezorConnect.init({
-    connectSrc: TREZOR_DEV
-      ? TREZOR_CONNECT_URL
-      : "https://connect.trezor.io/9.1.9/", // pinning to this connect version to avoid backwards incompatible changes
+    connectSrc: TREZOR_DEV ? TREZOR_CONNECT_URL : undefined,
+    coreMode: "auto", // default value still required to enable WebUSB support
```
bucko13, line 134: So what is the default behavior of this change? would it go back to using the latest? I'd rather if we could continue pinning to a particular version if possible. I'm also not familiar with the mechanics of the recent api change though.

kimlwh: I believe `connectSrc` is mainly intended for operating on different origins rather than pinning versions, at least I've never come across it being used that way, though I'm not aware of the reason why this was originally implemented. I would have assumed the package and `connectSrc` versions should generally remain in lockstep.

kimlwh:
I haven't had time to set up a full test device nor can I run the emulator but everything outside of signing seems to work. Co-signer extended keys are missing information when viewing addresses but I believe that's always been broken in Caravan due to not correctly sending the full extended key parameters in [trezorPublicKey(publicKey)](https://github.com/caravan-bitcoin/caravan/blob/35e944d24e662a8bfb4ad8379e9490cdd9254b10/packages/caravan-wallets/src/trezor.ts#L1107). Would be nice to fix but not a regression that I can tell in this case.

packages/caravan-wallets/src/trezor.ts
bucko13: > mainly intended for operating on different origins rather than pinning versions, at least I've never come across it being used that way,

Heh, well you've come across one instance of it being used that way here! The rationale was explained [here](https://github.com/caravan-bitcoin/caravan/issues/161#issuecomment-2541827797), but basically without pinning, backwards incompatible changes can be inadvertently be foisted upon caravan users in ways that (for example) break signing. This is actually pretty critical as this exact situation has happened before and we had to rush out a patch that did pin versions to avoid users being locked out of funds, so I would hate to lose this.

> I would have assumed the package and connectSrc versions should generally remain in lockstep.

This seems reasonable. Maybe pin the package version in the package.json (remove the `^` and run `npm install` again to update the package-lock) and have the versions match.

bucko13:
> I haven't had time to set up a full test device nor can I run the emulator but everything outside of signing seems to work. Co-signer extended keys are missing information when viewing addresses but I believe that's always been broken in Caravan due to not correctly sending the full extended key parameters in [trezorPublicKey(publicKey)](https://github.com/caravan-bitcoin/caravan/blob/35e944d24e662a8bfb4ad8379e9490cdd9254b10/packages/caravan-wallets/src/trezor.ts#L1107). Would be nice to fix but not a regression that I can tell in this case.

Ok, great, we can try and test the signing ourselves too.

dylan-thompson (approved):
Confirm address on device and signing works ✅

kimlwh:
Thanks @dylan-thompson and @bucko13. I've had no problems with using the updated version so far either for what it's worth.

PR #164, opened 2024-12-17
fix: turbo dev depends on build w/ esbuild noExternal

bucko13:
I noticed that after making the change in https://github.com/caravan-bitcoin/caravan/commit/5c3c456d78e8ab017cb960027d316cff98dc3ca2 turbo.json needed to define build as a dependency for dev tasks since the `noExternal` dep would need to be pre-built and bundled. I noticed this was breaking when trying to run `npm run dev` fresh.


chadchapnick (approved):
Confirmed this fixes the development server for the coordinator with a fresh install, and that the HEAD of main will currently error out if the `^build` dependency is not specified.

PR #168, opened 2025-01-05
feat(wallet): Add transaction history tab with pending/confirmed views

Legend101Zz:
Add a new **Transactions** tab to the wallet interface that provides users with a detailed view of their transaction history. The tab includes:

## Features
- **Split view** for pending and confirmed transactions.
- **Transaction tables** with key information:
  - Transaction ID with copy functionality.
  - Time/date of transaction.
  - Total fee.
  - Transaction size in vBytes.
  - Confirmation status.
  - Block height (for confirmed transactions).
- Ability to view transaction details in a block explorer.
- **Auto-refresh** capability for transaction status updates.
- Clear loading states and error handling for a better user experience.

## Technical Changes
- Added new transaction state management in **Redux**.
- Implemented transaction filtering by confirmation status.
- Added data fetching logic for transaction details.
- Included test coverage for:
  - Redux actions.
  - Reducers.
  - Component behavior.

## Why This Change?
This enhancement improves the wallet's usability by providing users with:
- Better visibility into their transaction history and statuses.
- Easy access to key details for pending and confirmed transactions.
- A more streamlined and informative transaction management experience.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new "Transactions" view in the wallet, allowing you to easily switch between confirmed and pending transactions.
	- The transactions screen features a sortable table with details such as fees, size, and time, plus refresh and error indicators.
	- Enhanced transaction data retrieval ensures more consistent and reliable transaction information display.


Legend101Zz:
Here’s a preview of the new **Transactions** tab, showing the split view for pending and confirmed transactions:

Screenshot 2025-01-05 at 23 38 04

Legend101Zz:
### Dependency Notice  

This PR depends on #134 as it uses a newly added transaction method we will need here. So we need to ensure #134 is reviewed and merged first.

bucko13:
> ### Dependency Notice
> This PR depends on #134 as it uses a newly added transaction method we will need here. So we need to ensure #134 is reviewed and merged first.

Submitted another review. Should be almost ready to go. I'll make sure to prioritize review and merge of #134 ASAP

apps/coordinator/src/components/Wallet/TransactionsTab.jsx
```diff
@@ -0,0 +1,361 @@
...
+    transactions && Array.isArray(transactions)
+      ? transactions.filter((tx) => !tx.status?.confirmed)
+      : [];
+  const confirmedTxs =
+    transactions && Array.isArray(transactions)
+      ? transactions.filter((tx) => tx.status?.confirmed)
+      : [];
+
+  // Render the transaction table
+  const renderTransactionTable = (txs) => (
```
bucko13, line 172: This is a pattern from our legacy code that we'd really like to move away from. It's kind of an anti-pattern for react. This should just be a component `<TransactionTable txs={txs} />`. This way you get react rendering optimizations, type safety, and a cleaner parent component because you can pull this out as a standalone component.

apps/coordinator/src/actions/transactionActions.js
bucko13, line 1: is there a reason this needs to be in redux? Are there any other components that need to use this outside of the component tree you've added? If you need to pass this data down through multiple children, you can also use [context](https://react.dev/reference/react/useContext) instead of redux. You'll still have access to what you need from the store use useSelector or useDispatch as needed. If we can avoid adding to our redux code that's preferable (when possible).

bucko13 (changes requested):
I haven't been able to test because the client method isn't available, but excited for this change! 

I think a lot of my requests for changes are due to you using old code as a reference for this change. Sorry for that confusion with the old code. Hopefully my comments give an idea of the preferred patterns though.

apps/coordinator/src/actions/transactionActions.js
Legend101Zz: Ahh ... yes I realise it now ... maybe I over-thought of having the use of redux for passing the states onto the CPFP/RBF components ... but now that you have mentioned that you'd refer not to use redux and rely on modern react practices... I revert back to the version I wrote without the use of redux :)

bucko13:
FYI: we'll need a changeset for the coordinator as well.

Legend101Zz:
Here are some screenshots to how the transaction tab looks in both private and non-private nodes ...

## `Private`
Screenshot 2025-02-21 at 19 06 48

Screenshot 2025-02-21 at 19 08 00


## `Normal`

Screenshot 2025-02-21 at 19 10 28

apps/coordinator/src/components/Wallet/TransactionsTab.jsx
```diff
@@ -0,0 +1,361 @@
...
+  setTransactionsLoading,
+  setTransactionsError,
+}) {
+  const [tabValue, setTabValue] = useState(0);
+  const [mounted, setMounted] = useState(true);
+
+  // Helper function to fetch individual transaction details
+  const fetchTransactionDetails = async (client, txid) => {
+    try {
+      const txDetails = await client.getTransaction(txid);
```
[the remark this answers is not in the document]
Legend101Zz: Thanks for the suggestion ... I have now rebased the current branch to use #134 as base branch

apps/coordinator/src/components/Wallet/TransactionsTab.jsx
```diff
@@ -0,0 +1,361 @@
...
+              No confirmed transactions
+            </Typography>
+          )}
+        </TabPanel>
+      </CardContent>
+    </StyledCard>
+  );
+}
+
+TransactionsTab.propTypes = {
```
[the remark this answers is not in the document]
Legend101Zz: refactoring [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/0a5c90f6bdaea3ee0b73198e3bb08139179a81c3)

apps/coordinator/src/components/Wallet/TransactionsTab.jsx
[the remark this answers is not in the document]
Legend101Zz: refactoring [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/0a5c90f6bdaea3ee0b73198e3bb08139179a81c3)

apps/coordinator/src/components/Wallet/TransactionsTab.jsx
```diff
@@ -0,0 +1,361 @@
...
+    error: state.spend.transaction.transactions.error,
+  };
+};
+
+const mapDispatchToProps = {
+  getBlockchainClient: updateBlockchainClient,
+  setTransactions,
+  setTransactionsLoading,
+  setTransactionsError,
+};
```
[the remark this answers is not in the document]
Legend101Zz: removed the unecessary redux [states](https://github.com/caravan-bitcoin/caravan/pull/168/commits/fa37dbc85d5f08120d3cf77805fa18a94a735536)

apps/coordinator/src/actions/transactionActions.js
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/fa37dbc85d5f08120d3cf77805fa18a94a735536)

apps/coordinator/src/components/Wallet/TransactionsTab.jsx
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/0a5c90f6bdaea3ee0b73198e3bb08139179a81c3)

jevidon:
Thanks for the updates here! Some feedback on the changes:

- It would be helpful to show a column with the bitcoin value of the UTXO
- The "Transaction ID" pill in the left column should just copy the ID to the clipboard instead of opening a block explorer
- The “open in preferred explorer” link at the right end seems to always open in mempool.space - is it possible to have that use whatever block explorer was selected by the user when loading the wallet?

Legend101Zz:
> Thanks for the updates here! Some feedback on the changes:
> 
> * It would be helpful to show a column with the bitcoin value of the UTXO
> * The "Transaction ID" pill in the left column should just copy the ID to the clipboard instead of opening a block explorer
> * The “open in preferred explorer” link at the right end seems to always open in mempool.space - is it possible to have that use whatever block explorer was selected by the user when loading the wallet?

Thanks for the reviews @jevidon I have pushed the required changes as per the reviews ... please let me know if they work :)

bucko13:
Screenshot 2025-03-14 at 4 07 27 PM

Unfortunately it doesn't look like the table is populating correctly now. This is a local regtest node so communicating via bitcoind rpc. The network calls are being made, but I am getting an error still:

```
Error fetching tx 8962df7edfe0e610af294ae1b6a738f325df96cebb9e3846dff420f4c404c34f: Error: Failed to get transaction: Cannot read properties of undefined (reading 'txid')
    at BlockchainClient.getTransaction (index-370fe92a.js:441:39165)
```

.changeset/empty-ants-destroy.md
bucko13, line 1: cleanup. This is probably from an old merge.

Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/c3d1e3b4cbbf4cf8512c99a813420da0775165e0#diff-4a1788cdad5570add1ee5b095b22fa1e919225b1bdb94973e920be11895b34b9)

Legend101Zz:
> FYI: we'll need a changeset for the coordinator as well.

I've added all the required changesets, and I had to include an extra one for the `clients package` too. This extra changeset covers a `minor patch` for `BlockchainClient.getTransaction`—it now properly accesses the result property that was already being returned by `bitcoindRawTxData`.

Sorry for not catching this in #134; I'll be more careful next time `:)

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+            explorerUrl = `https://blockstream.info/${network === "mainnet" ? "" : "testnet/"}tx/${txid}`;
+            break;
+
+          default:
+            // Fall back to the default explorer
+            explorerUrl = blockExplorerTransactionURL(txid, network);
+        }
+      } else {
+        // If no blockchain client specified, use default
+        explorerUrl = blockExplorerTransactionURL(txid, network);
```
bucko13, line 155: feels like this is duplicative of the default case above.

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+  }, [fetchTransactions]);
+
+  // Handle sorting
+  const handleSort = (property: string) => {
+    const isAsc = sortBy === property && sortDirection === "asc";
+    setSortDirection(isAsc ? "desc" : "asc");
+    setSortBy(property);
+  };
+
+  const handleTransactionClick = useCallback(
```
bucko13, line 120: `handleExplorerLinkClick` feels like a better function name.

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+}) => {
+  const [tabValue, setTabValue] = useState(0);
+  const [transactions, setTransactions] = useState<Transaction[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const [sortBy, setSortBy] = useState<string>("blockTime");
+  const [sortDirection, setSortDirection] = useState<"asc" | "desc">("desc");
+
+  // Fetch transactions
+  const fetchTransactions = useCallback(async () => {
```
bucko13, line 66: Do you think it would be worthwhile to pull these `useCallbacks` out into their own file to clean up the component? You can make them reusable hooks, create a `TransactionsTable` directory and then put them in `TransactionsTable/hooks.ts`. It might be nice to cleanup the component to focus just on the view elements.

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
bucko13: could also be a reusable hook that is in a `TransactionsTable/hooks.ts` file, something like `useHandleExplorerTxLinkClick` that returns a function to be called.

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+  deposits: state.wallet.deposits,
+  change: state.wallet.change,
+  client: state.client,
+});
+
+const mapDispatchToProps = {
+  getBlockchainClient: updateBlockchainClient,
+};
+
+export default connect(mapStateToProps, mapDispatchToProps)(TransactionsTab);
```
bucko13, line 273: shouldn't need connect or mapStateToProps anymore. There is a useDispatch and useSelector hook that you can use and not require the props at all.

apps/coordinator/src/components/Wallet/TransactionsTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+}) => {
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
+
+  // Handler for sorting
+  const createSortHandler = (property: keyof Transaction) => () => {
+    onSort(property);
+  };
+
+  // Render table header with sort labels
+  const renderTableHeader = () => (
```
bucko13, line 108: Confusing because I know this pattern is used elsewhere in the codebase, but that's largely legacy code and this is a pattern that should be avoided. You can just pull these out as their own standalone component `<TransactionTableHeader />` and use that int he component, or just put them directly in the main render section.

bucko13 (changes requested):
I did a little bit of testing. Seems to be working for the most part. I've left some code comments to address, a mix of aesthetic and more substantive. Getting really close though!

bucko13:
I merged and released the fix #196 and merged the changes back into this PR.

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+  Chip,
+} from "@mui/material";
+import { Refresh } from "@mui/icons-material";
+import { BlockchainClient } from "@caravan/clients";
+import { blockExplorerTransactionURL, Network } from "@caravan/bitcoin";
+import { updateBlockchainClient } from "../../actions/clientActions";
+import { TransactionTable } from "./TransactionsTable";
+
+// How are Transaction should look like
+interface Transaction {
```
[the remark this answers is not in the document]
Legend101Zz: Yes, it's available as the `TransactionDetails` type , but it looks like we aren’t exporting all the necessary types from the client package.
```ts
export { bitcoindImportDescriptors } from "./wallet";
export { BlockchainClient, ClientType } from "./client";
export type { UTXO, Transaction, FeeRatePercentile } from "./types";
```
Would it be better to create a separate PR to fix this first by exporting the missing types, and then update the code to use the proper types provided by the package?

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+        node.utxos.forEach((utxo: { txid: string }) => {
+          txids.add(utxo.txid);
+        });
+      });
+
+      // Fetch transaction details in parallel
+      const txDetails = await Promise.all(
+        Array.from(txids).map(async (txid) => {
+          try {
+            const tx = await blockchainClient.getTransaction(txid);
```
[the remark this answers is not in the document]
Legend101Zz: Understood ... 
`"fee" : n, (numeric) The amount of the fee in BTC. This is negative and only available for the
                                           'send' category of transactions.`
  So I tested out the `getWalletTransaction` by implementing it in the client's package :
  
  ```ts
  /**
 * Gets detailed information about a wallet transaction
 *
 * This function uses the "gettransaction" RPC call which includes fee information
 * for wallet transactions, unlike getrawtransaction which requires txindex=1 and
 * doesn't include fee data.
 *
 * @param options - Connection details and transaction ID
 * @returns Detailed transaction data with fee information
 */
export async function bitcoindGetWalletTransaction({
  url,
  auth,
  walletName,
  txid,
  includeWatchonly = true,
  verbose = true,
}: BaseBitcoindParams & {
  txid: string;
  includeWatchonly?: boolean;
  verbose?: boolean;
}): Promise<any> {
  try {
    const response = await callBitcoindWallet({
      baseUrl: url,
      walletName,
      auth,
      method: "gettransaction",
      params: [txid, includeWatchonly, verbose],
    });

    if (typeof response?.result === "undefined") {
      throw new BitcoindWalletClientError(
        `Error: invalid response from ${url} for transaction ${txid}`,
      );
    }

    return response.result;
  } catch (e) {
    const error = e as Error;
    // Handle specific error for transaction not found in wallet
    if ((e as any)?.response?.data?.error?.code === -5) {
      throw new BitcoindWalletClientError(
        `Transaction ${txid} not found in wallet. Either the transaction doesn't exist or it's not in this wallet.`,
      );
    }
    console.error("Error getting wallet transaction:", error.message);
    throw e;
  }
}
  ```
  
  
- Since the fee field is only available for sent transactions (and absent for received ones), we could use this to categorize transactions as sent or received.
- However, for public clients, where we need to determine the net sent/received amount, we'll need an explicit calculation.
                                           
and as discussed for :
> It would be helpful to show a column with the bitcoin value of the UTXO
That we'll need a way to figure out if the net vout's are to one of our addresses or not ...

So should we keep this change for that PR or should I push a new `client feat PR`  changing the `BlockchainClient.getTransaction` to use the `bitcoindGetWalletTransaction` instead of `bitcoindRawTxData`

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+            // Fall back to the default explorer
+            explorerUrl = blockExplorerTransactionURL(txid, network);
+        }
+      } else {
+        // If no blockchain client specified, use default
+        explorerUrl = blockExplorerTransactionURL(txid, network);
+      }
+
+      window.open(explorerUrl, "_blank");
+      window.open(explorerUrl, "_blank");
```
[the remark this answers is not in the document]
Legend101Zz: Ahh ... sorry that seems to be a typo

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+            break;
+
+          case "blockstream":
+            // Use blockstream.info explorer
+            explorerUrl = `https://blockstream.info/${network === "mainnet" ? "" : "testnet/"}tx/${txid}`;
+            break;
+
+          default:
+            // Fall back to the default explorer
+            explorerUrl = blockExplorerTransactionURL(txid, network);
```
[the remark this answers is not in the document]
Legend101Zz: Ahh got it .. so maybe a `PR` to add this feature into the `bitcoin package` ?

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const [sortBy, setSortBy] = useState<string>("blockTime");
+  const [sortDirection, setSortDirection] = useState<"asc" | "desc">("desc");
+
+  // Fetch transactions
+  const fetchTransactions = useCallback(async () => {
+    setIsLoading(true);
+    setError(null);
+
+    try {
+      const blockchainClient = await getBlockchainClient();
```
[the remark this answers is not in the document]
Legend101Zz: Did not explored it till now ... but now have updated the component to use that [hook](https://github.com/caravan-bitcoin/caravan/pull/168/commits/15969b3e955eb785c75b76a492c74adebb102636)

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
Legend101Zz: Thanks for pointing that out ... changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/6a0abda7e8da9610f05eb969ed144a7a689b1fbd)

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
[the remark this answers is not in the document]
Legend101Zz: also how does this [look](https://github.com/caravan-bitcoin/caravan/pull/168/commits/f795f11efbf325f2664769f08f41242b81692eb2)  ?

apps/coordinator/src/components/Wallet/TransactionsTable.tsx
Legend101Zz: Sorry for the miss .. I seem have not refactored these into new modern React styled components ...
Changes done [now](https://github.com/caravan-bitcoin/caravan/pull/168/commits/a057e88304d2b2cd8efb9d42a78cf9a79f1e98fd)

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
Legend101Zz: yes that sounds great ... It would increase the maintainability and testing of the component ...

I have made the changes [here](https://github.com/caravan-bitcoin/caravan/pull/168/commits/5b4a3c400b3d33966d707cf16c9d33da26ef0f1e)

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
Legend101Zz: I have made the changes [here](https://github.com/caravan-bitcoin/caravan/pull/168/commits/5b4a3c400b3d33966d707cf16c9d33da26ef0f1e)

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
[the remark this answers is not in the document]
bucko13: > So should we keep this change for that PR or should I push a new client feat PR changing the BlockchainClient.getTransaction to use the bitcoindGetWalletTransaction instead of bitcoindRawTxData

I think we could keep the existing method and just add a new method to the client so both are supported. That way it's also not a breaking change to the api. but a separate PR would be great and should be a quick approval and merge as well. 

Good catch about fees only being available for sent. I think that's a great idea that we can use that for identifying internal wallet transactions. I think for public clients like mempool, they'll always include the fee since they have much greater ability to do indexing, so it might be tricky to rely just on that. for dealing with incoming txs we want to bump (can only do CPFP for those though) we'll have to calculate the fees, or to start just leave it blank for private clients but if someone is waiting a long time they can go to a public client like mempool or just bump with cpfp if they're getting impatient.

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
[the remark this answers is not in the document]
bucko13: > Would it be better to create a separate PR to fix this first by exporting the missing types, and then update the code to use the proper types provided by the package?

Yeah, that'd be great!

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -0,0 +1,253 @@
+import { useState, useEffect, useCallback } from "react";
+import { useSelector } from "react-redux";
+import { blockExplorerTransactionURL } from "@caravan/bitcoin";
+import { Transaction, SortDirection, SortBy } from "./types";
+import { useGetClient } from "../../../hooks/client";
+
+/**
+ * Custom hook to manage transaction fetching and state
+ */
+export const useTransactions = () => {
```
bucko13, line 10: naming nit, i prefer having a verb in there as well where applicable like this. so `useFetchTransactions`

apps/coordinator/src/components/Wallet/TransactionsTab/types.ts
```diff
@@ -0,0 +1,46 @@
+// Types for the transaction data
+export interface Transaction {
```
bucko13, line 2: should we put a TODO here that this should come from @caravan/clients?

bucko13 (commented):
just a couple nits but a really nice cleanup!

apps/coordinator/src/components/Wallet/TransactionsTab.tsx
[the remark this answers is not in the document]
Legend101Zz: Yes, that makes sense ... I'll research further into checking how other wallets handle RBF/CPFP for private clients

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/a73c2e9fb02b7f5d6281a1030138eac693ef8e88#diff-511018b155f4393fa6f4bc530aff6da3ac23c2527ff6ce2af32404419e4cf4b6R10)

apps/coordinator/src/components/Wallet/TransactionsTab/TransactionsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+const columns = [
+  { id: "txid", label: "Transaction ID", sortable: false },
+  { id: "blockTime", label: "Time", sortable: true },
+  { id: "size", label: "Size (vBytes)", sortable: true },
+  { id: "fee", label: "Fee (sats)", sortable: true },
+  { id: "status", label: "Status", sortable: false },
+  { id: "actions", label: "", sortable: false },
+];
+
+// FeeDisplay component to display fees display in both sats and BTC
```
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/a73c2e9fb02b7f5d6281a1030138eac693ef8e88#diff-4a0b69244c694fd95d0106a3eaef6a945cd690f68507107d38d02cddc6bc5e62R1)

apps/coordinator/src/components/Wallet/TransactionsTab/types.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/168/commits/a73c2e9fb02b7f5d6281a1030138eac693ef8e88#diff-4a0b69244c694fd95d0106a3eaef6a945cd690f68507107d38d02cddc6bc5e62R1)

jevidon:
I checked the latest vercel preview, and it looks good to me as a starting point to build on. Happy to see this merged.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -0,0 +1,280 @@
...
+            // Address has active UTXOs
+            addressesWithActiveUTXOs.add(node.multisig.address);
+
+            // Add transaction IDs from active UTXOs
+            node.utxos.forEach((utxo: any) => {
+              if (utxo.txid) txids.add(utxo.txid);
+            });
+          } else if (node.addressUsed) {
+            // Address has been used but has no UTXOs (spent)
+            addressesWithSpentUTXOs.add(node.multisig.address);
```
bucko13, line 73: curious why this was necessary?

bucko13 (approved):

PR #174, opened 2025-01-24
fix(psbt): lock dependency of caravan-bitcoin

bucko13:
I tried to do the noExternal fix that we do in caravan/wallets but unfortunately trying to pre-bundle caravan/bitcoin caused all sorts of issues (as far as I can tell most of those revolve around the pre-bundling of bitcoinjs-lib). I think I had the same issue trying to pre-bundle caravan-bitcoin in caravan-wallets too. 

This should resolve the issue. It just means it makes testing and developing in caravan/psbt more complex if we need to make changes in in caravan-bitcoin.

### DIFF
--- a/apps/coordinator/package.json
+++ b/apps/coordinator/package.json
@@ -77,6 +77,7 @@
     "predeploy": "npm run build",
     "start": "HTTPS=true npm run preview",
     "test": "__GIT_SHA__=`git rev-parse --short HEAD` jest",
+    "test:watch": "__GIT_SHA__=`git rev-parse --short HEAD` jest --watch",
     "test:ci": "CI=true jest --coverage",
     "dev": "__GIT_SHA__=`git rev-parse --short HEAD` vite",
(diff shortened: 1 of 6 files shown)


apps/coordinator/package.json
```diff
@@ -77,6 +77,7 @@
     "predeploy": "npm run build",
     "start": "HTTPS=true npm run preview",
     "test": "__GIT_SHA__=`git rev-parse --short HEAD` jest",
+    "test:watch": "__GIT_SHA__=`git rev-parse --short HEAD` jest --watch",
```
Shadouts, line 80: :+1:

package-lock.json
Shadouts, line 1: Just noting patch version differences on `node_modules/undici` and `node_modules/vite` after diffing the following output on `main`:
```
cd packages/caravan/psbt && npm rm @caravan/bitcoin && npm i @caravan/bitcoin@0.3.1 && npm audit fix
```

Shadouts (approved):

PR #185, opened 2025-03-05
Fix/docker updates

robertshuford:
If we follow the https://docs.docker.com/build/building/multi-platform/#cross-compiling-a-go-application multi-platform build strategy, we will avoid the strange test failure we were seeing with arm builds.

Build failure: https://github.com/caravan-bitcoin/caravan/actions/runs/13575497175/job/38014051338
Test build I ran on another branch: https://github.com/caravan-bitcoin/caravan/actions/runs/13672772724/job/38226478331

### DIFF
--- a/.changeset/poor-eggs-hope.md
+++ b/.changeset/poor-eggs-hope.md
@@ -1,0 +1,5 @@
+---
+"caravan-coordinator": patch
+---
+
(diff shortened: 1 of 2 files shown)


robertshuford:
@coderabbitai full review

PR #187, opened 2025-03-10
Vitest migration

bucko13:


bucko13:
I'm going to close this PR in preference for a more piecemeal approach. Apologies for all those that were involved in putting this together. It was getting unwieldy and I think there were other design decisions we could've made. see #228 for a reference for moving forward.

PR #188, opened 2025-03-11
Allow a PsbtV2 to be constructed with a psbtv2 which has been forceably set to have PSBT_GLOBAL_TX_VERSION === 1

Shadouts:
This change is to help with compatibility when converting between psbt versions. `PsbtV2.FromV0()` has an option to create a PSBTv2 forceably setting the `PSBT_GLOBAL_TX_VERSION` field to 1. This is a bip370 non-compliant psbtv2, but it was presumably added for compatibility reasons when converting from v0.

A `PsbtV2` class created in such a way can be serialized with `PsbtV2.serialize()`, however, the resultant serialized psbt cannot be used to re-instantiate a `PsbtV2`. This change relaxes the Txnv2 requirement so that these exception case serialized psbts can be passed back into the constructor.

It should be carefully noted that this change might imply a dangerous use for a PSBTv2. The bip370 specifies that the txn v2 requirement was added to simplify the negotiation of validation rules for `OP_CHECKSEQUENCEVERIFY`. Forcing the txnv1 does not guarantee that consumers of the resultant psbtv2 will treat the calculated locktimes properly.

### DIFF
--- a/packages/caravan-psbt/src/psbtv2/psbtv2.ts
+++ b/packages/caravan-psbt/src/psbtv2/psbtv2.ts
@@ -24,19 +24,26 @@ import { bufferize } from "../functions";
  * BIP-defined keytypes. Very few setters and modifier methods exist. As they
  * are added, they should enforce implied and documented rules and limitations.
  *
- * Defining BIPs:
- * https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki
+ * allowTxnVersion1: A Note
+ * A psbtv2 must have its transaction version GTE 2 to be bip370 compliant. If
+ * this class is instantiated with allowTxnVersion1 set to `true`, then a psbtv2
+ * which has had its txn version forceably set to 1 (for example with
+ * PsbtV2.dangerouslySetGlobalTxVersion1) can be instantiated. This has,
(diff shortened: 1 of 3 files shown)


packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -766,7 +774,7 @@ export class PsbtV2 extends PsbtV2Maps {
       }
     }
     for (const amount of this.PSBT_OUT_AMOUNT) {
-      if (!amount) {
+      if (amount === undefined || amount === null) {
```
Shadouts, line 777: Tests revealed this to be a bug. It was failing when `amount === 0`, but a zero amount value was included in the bip174 vectors.

bucko13: Oh so we were missing this because only the v0 tests had an amount of 0? So it wasn't until you ran the v0 vectors through the v2 convertor that you found this bug?

bucko13 (approved):

bucko13:
Doesn't prove a whole lot, but as a smoke test, i confirmed ledger signing (which uses psbt v2) still works. 

Screenshot 2025-03-11 at 10 08 00 AM

PR #194, opened 2025-03-17
feat: remove any types from packages/caravan-clients

AvhiMaz:


bucko13:
> @bucko13 please. have a look, thank you!

will do! thanks. things are a bit backed up at the moment but definitely will try and get to this asap.

Legend101Zz:
Hey @AvhiMaz ! Thanks a ton for the PR—really appreciate the effort! 

Just dropping in some review comments for a few tweaks. Let me know what you think! "

packages/caravan-clients/src/blockchain.ts
```diff
@@ -109,9 +117,12 @@ export async function fetchAddressUTXOs(
 
   // calculate the total balance from all utxos
   const balanceSats = utxos
-    .map((utxo) => utxo.amountSats)
+    .map((utxo: { amountSats: BigNumber }) => utxo.amountSats)
```
Legend101Zz, line 120: It looks like the suggested change was likely intended to address this issue. However, these modifications are unnecessary. Instead, we can simply define `unsortedUTXOs` as `FormattedUTXO[]`, which aligns with the expected structure.  

Additionally, changing `unsortedUTXOs: { amountSats: BigNumber }[] | undefined;` breaks the following usage:  

```js
const utxos = sortInputs(unsortedUTXOs);
```  

This happens because `sortInputs` expects each UTXO to have a `txid`, as seen in its implementation:  

```js
export function sortInputs(inputs) {
  return inputs.sort((input1, input2) => {
    if (input1.txid > input2.txid) {
      return 1;
    } else if (input1.txid < input2.txid) {
      return -1;
    } else {
      return input1.index < input2.index ? -1 : 1;
    }
  });
}
```  

Since `sortInputs` relies on `txid` and `index`, the additional traversal logic seems unnecessary. Keeping `unsortedUTXOs` typed as `FormattedUTXO[]` should resolve the issue without modifying the function's structure.

Legend101Zz (changes requested):
Some initial reviews for the `bitcoind.ts` and `blockchain.ts` files

Legend101Zz:
@coderabbit re-review

packages/caravan-clients/src/client.ts
```diff
@@ -102,7 +107,7 @@ export class ClientBase {
     }
   }
 
-  private async Request(method: Method, path: string, data?: any) {
+  private async Request(method: Method, path: string, data?: object) {
```
Legend101Zz, line 110: Ok ... so this would break existing functionality...

```ts
public async Post(path: string, data?: object) {
  return this.Request("POST", path, data);
}
```

Changing the parameter type from any to object would prevent passing strings to the Post method, which would break this functionality. 

```ts
public async broadcastTransaction(rawTx: string): Promise<any> {
  try {
    if (this.type === ClientType.PRIVATE) {
      return bitcoindSendRawTransaction({
        hex: rawTx,
        ...this.bitcoindParams,
      });
    }
    return await this.Post(`/tx`, rawTx); // rawTx is a string, not an object
  } catch (error: any) {
    throw new Error(`Failed to broadcast transaction: ${error.message}`);
  }
}
```

One idea can be to use a union type instead ( But honestly I am not a fan of this , I guess we can do better):

```ts
public async Post(path: string, data?: string | object | FormData) {
  return this.Request("POST", path, data);
}
```

packages/caravan-clients/src/client.ts
```diff
@@ -113,16 +118,20 @@ export class ClientBase {
...
+      }
     }
   }
 
   public async Get(path: string) {
     return this.Request("GET", path);
   }
 
-  public async Post(path: string, data?: any) {
+  public async Post(path: string, data?: object) {
```
Legend101Zz, line 134: I left a comment above telling how can it potentially break some functionalities

packages/caravan-clients/src/client.ts
```diff
@@ -198,18 +207,29 @@ export class BlockchainClient extends ClientBase {
...
-  public async getAddressUtxos(address: string): Promise<any> {
+  public async getAddressUtxos(address: string): Promise<UTXO[]> {
     try {
       if (this.type === ClientType.PRIVATE) {
-        return bitcoindListUnspent({
+        const unspent = await bitcoindListUnspent({
           address,
           ...this.bitcoindParams,
         });
+        return unspent.map((utxo) => ({
```
Legend101Zz, line 217: looks good ... can you please check that `bitcoindListUnspent` returns `index` instead of vout so 
`vout: utxo.index` and looking at 
```ts
export interface UTXO {
  txid: string;
  vout: number;
  value: number;
  status: {
    confirmed: boolean;
    block_time: number;
  };
}
```

where are we getting the extra `block_height` and `block_hash` fields and also I see `bitcoindListUnspent` not returning these fields :

```ts
Promise<
  {
    txid: string;
    amount: string;
    amountSats: string;
    index: number;
    confirmed: boolean;
    transactionHex: string;
    time: string;
  }[]
  ```
  
  can you please check

File: packages/caravan-clients/src/client.ts [hunk 5]
```diff
@@ -198,18 +207,29 @@ export class BlockchainClient extends ClientBase {
...
+          value: Number(utxo.amount),
+          status: {
+            confirmed: utxo.confirmed,
+            block_height: utxo.block_height,
+            block_hash: utxo.block_hash,
+            block_time: Number(utxo.time),
+          },
+        }));
       }
       return await this.Get(`/address/${address}/utxo`);
```
Legend101Zz, line 229: If possible can also cross reference the [mempool](https://mempool.space/docs/api/rest#get-address-utxo) and [blockstream](https://github.com/Blockstream/esplora/blob/master/API.md#get-addressaddressutxo) API docs and check if what type you are expecting they are returning that only

packages/caravan-clients/src/client.ts
```diff
@@ -252,8 +272,8 @@ export class BlockchainClient extends ClientBase {
             }
             for (const output of rawTxData.vout) {
               transaction.vout.push({
-                scriptPubkeyHex: output.scriptPubKey.hex,
-                scriptPubkeyAddress: output.scriptPubKey.address,
+                scriptPubkeyHex: output.scriptpubkey,
```
Legend101Zz, line 275: Please try to limit changes strictly to type definitions to avoid any unintended side effects. Altering existing functionality, even slightly, could introduce breaking changes across the entire package.

The goal should be to enhance type safety while keeping modifications to the core logic as minimal as possible. If a functional change is absolutely necessary, it would be great to discuss it first to assess the potential impact.

packages/caravan-clients/src/client.ts
```diff
@@ -484,26 +514,31 @@ export class BlockchainClient extends ClientBase {
...
       );
     }
   }
 
-  public async getTransactionHex(txid: string): Promise<any> {
+  public async getTransactionHex(txid: string): Promise<string> {
     try {
       if (this.type === ClientType.PRIVATE) {
-        return await callBitcoind<TransactionResponse>(
+        const response = await callBitcoind<TransactionResponse>(
```
Legend101Zz, line 527: looks good

packages/caravan-clients/src/types.ts
```diff
@@ -279,6 +284,7 @@ export interface BlockExplorerAddressResponse {
  * Interface for formatted UTXO data
  */
 export interface FormattedUTXO {
+  vout: number;
```
Legend101Zz, line 287: please try to avoid changes like these as they would break the package in many places as discussed above

packages/caravan-clients/src/types.ts
```diff
@@ -230,6 +230,11 @@ export interface ListUnspentResponse {
 }
 
 export interface TransactionResponse {
+  size: number;
```
Legend101Zz, line 233: Why  was  it needed to add more fields here ??

packages/caravan-clients/src/wallet.ts
```diff
@@ -206,8 +209,19 @@ export async function bitcoindListUnspent({
         index: utxo.vout,
         amount: amount.toFixed(8),
         amountSats: bitcoinsToSatoshis(amount.toString()),
-        transactionHex: previousTransactions[mapindex].result.hex,
-        time: previousTransactions[mapindex].result.blocktime,
+        transactionHex: (
```
Legend101Zz, line 212: again this looks like a breaking change to me ... can you please verify if this won't break anything

Legend101Zz (changes requested):
Some more review comments ...

I'd like you to focus on two things :

1) Try to limit changes strictly to type definitions to avoid any unintended side effects as Altering existing functionality, even slightly, could introduce breaking changes across the entire package.

2) The catch block `any` comments I mentioned

bucko13:
@AvhiMaz if you have time to address the comments, i might have some availability to review and hopefully merge this week.

Legend101Zz:
Hi @AvhiMaz, just checking in to see if you’ve had a chance to go through the pending review comments on the PR. Let me know if you need any help or clarification :)

Legend101Zz:
Hey @AvhiMaz , just checking in — we haven’t seen a response on this for a while.
@bucko13 , if there’s no update soon, we can go ahead and close this PR.

bucko13:
Closing for inactivity.

PR #195, opened 2025-03-18
feat/docs caravan health privacy

AvhiMaz:


bucko13:
Hi @AvhiMaz. Thanks for the contributions! Could you please split up the commit history of this PR so it only has the relevant code? The title says docs for caravan health but it includes #194 code.

packages/caravan-health/docs/health-metrics.md
bucko13, line 1: I like having this documentation in jsdocs. What's the motivation for adding this file as well? I don't mind it but we don't have this anywhere else and my biggest concern is that it will be hard to maintain. Ideally we just have more jsodcs closer to the code like you've added and then we can start generating these docs automatically.

AvhiMaz: The motivation for adding this file was to provide a structured overview of these metrics in one place, making it easier to reference their definitions, calculations, and expected ranges. However, I understand the concern about maintainability. I’m happy to keep everything in JSDoc instead and work towards generating the docs automatically. Let me know if you'd like me to remove this file and focus on enhancing the inline documentation instead.

AvhiMaz:
@bucko13  please. have a look I've made suggested changes, Thank you!

packages/caravan-health/src/privacy.ts
```diff
@@ -4,25 +4,15 @@ import { WalletMetrics } from "./wallet";
...
-      influence the topology type of the transaction.
-
-    Calculation :
-      We have 5 categories of transaction type each with their own impact on privacy score
-      - Perfect Spend (1 input, 1 output)
-      - Simple Spend (1 input, 2 outputs)
-      - UTXO Fragmentation (1 input, more than 2 standard outputs)
-      - Consolidation (more than 1 input, 1 output)
-      - CoinJoin or Mixing (more than 1 input, more than 1 output)
-  */
```
bucko13, line 25: i still think it's worth keeping these descriptions as they are pretty useful. Once we have doc generation this can all be put in a single document (stripped of the code part for readability).

package.json
```diff
@@ -20,7 +20,8 @@
     "@changesets/changelog-github": "^0.5.0",
     "@changesets/cli": "^2.27.1",
     "prettier": "^3.1.1",
-    "turbo": "^2.0.3"
+    "turbo": "^2.0.3",
+    "typedoc": "^0.28.1"
```
bucko13, line 24: I don't think we need to add this yet until we're actually using it in CI. We can do that separately.

bucko13:
I'm not really sure what the point of this PR is. It looks like some very basic cleanup and now adding a package that's not even being used yet. If you wanted to experiment with generating documentation that we can build and hopefully host with CI I'd be open to work on that (especially something more at the top monorepo level).

AvhiMaz:
@bucko13 closing the pr, let me focus on other one, will be seeing it, implement it correctly and will raise a pr again. thank you.

Issue #192, opened 2025-03-14
Improve Type System for `Caravan-Clients` Package


Legend101Zz:
## Summary
After digging through the codebase, I've noticed our type system in the `caravan-clients` package has several inconsistencies that make development harder than it needs to be. I'd like to propose a series of improvements to make our types more consistent, reduce `any` usage, and better align with Bitcoin RPC responses.

## The Problems
Looking at the code, I've noticed a few issues:

- **Types are all over the place** - We have similar types defined in multiple files, and it's not clear which ones we should use.
- **Too many `any` types** - We're using `any` in a lot of places with `eslint-disable` comments and `@ts-expect-error` workarounds.
- **Inconsistent parameter handling** - Sometimes we pass arrays like `[imports, { rescan }]`, other times objects like `{ minconf: 0, maxconf: 9999999, addresses: addressParam }`.
- **Type assertions without validation** - We often just assume what the response structure will be rather than properly typing it.
- **Each backend returns different formats** - `bitcoind`, `blockstream`, and `mempool` all return different structures, making it hard to work with.

## Proposed Approach

I think we should tackle this in stages through several PRs:

### 1. Centralize Type Definitions 
- Create a logical organization structure for types.
- Move shared types from various files to `types.ts`.
- Create consistent naming conventions for types.
- Document types with JSDoc comments.
- Remove duplicate/redundant type definitions.

### 2. Improve RPC Parameter Handling 
- Create a flexible `RPCParams` type that can handle both array and object formats:
  ```ts
  export type RPCParam = string | number | boolean | null | undefined | RPCParam[] | { [key: string]: RPCParam };
  export type RPCParams = RPCParam[] | Record<string, RPCParam>;
  ```
- Update RPC call functions to use these types.
- Document parameter patterns for different RPC methods.
- Update `BitcoindWalletParams` interface to use better typing.

### 3. Create Response Type Mapping 
- Create a reference table of Bitcoin RPC methods to TypeScript interfaces:
  | RPC Method | Documentation URL | Response Type | Currently Used In |
  |------------|-------------------|--------------|------------------|
  | `getwalletinfo` | [Link](https://developer.bitcoin.org/reference/rpc/getwalletinfo.html) | `WalletInfoResponse` | `wallet.ts` |
  | `listunspent` | [Link](https://developer.bitcoin.org/reference/rpc/listunspent.html) | `ListUnspentResponse[]` | `wallet.ts` |
[...]

PR #200, opened 2025-03-24
issue #192 Improved types by shifted to types.ts for the caravan-clients package

ice-009:
I have shifted the interfaces from wallets.ts to the type.ts for better code readablity and also removed reduntant types defined.

### DIFF
--- a/.changeset/eight-singers-smoke.md
+++ b/.changeset/eight-singers-smoke.md
@@ -1,5 +1,0 @@
----
-"@caravan/psbt": minor
----
-
-Fixes PsbtV2.isReadyForConstructor by not requiring that PSBT_GLOBAL_FALLBACK_LOCKTIME be set. This field is not required on a psbtv2 per BIP370.
--- a/.changeset/new-squids-talk.md
+++ b/.changeset/new-squids-talk.md
@@ -1,5 +1,0 @@
----
-"@caravan/clients": patch
----
-
-Patch: align client.getTransaction method implementation with bitcoindRawTxData response
--- a/.changeset/pretty-chefs-explain.md
+++ b/.changeset/pretty-chefs-explain.md
@@ -1,5 +1,0 @@
----
-"caravan-coordinator": minor
----
-
-Add a new Transactions tab to the wallet interface that provides users with a detailed view of their transaction history.
--- a/.changeset/selfish-dodos-cheer.md
+++ b/.changeset/selfish-dodos-cheer.md
@@ -1,5 +1,0 @@
----
-"caravan-coordinator": patch
----
-
-UI cleanup: Remove redundant header and focus on pending transactions
--- a/apps/coordinator/CHANGELOG.md
+++ b/apps/coordinator/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## 1.6.0
+
+### Minor Changes
+
+- [#168](https://github.com/caravan-bitcoin/caravan/pull/168) [`accc093`](https://github.com/caravan-bitcoin/caravan/commit/accc0937159691e7be5b0b68c389ee6594423aa4) Thanks [@Legend101Zz](https://github.com/Legend101Zz)! - Add a new Transactions tab to the wallet interface that provides users with a detailed view of their transaction history.
+
+### Patch Changes
+
+- [#205](https://github.com/caravan-bitcoin/caravan/pull/205) [`19bff49`](https://github.com/caravan-bitcoin/caravan/commit/19bff491cdcc8ccb9e092c76c75d826dd33e3c87) Thanks [@Legend101Zz](https://github.com/Legend101Zz)! - UI cleanup: Remove redundant header and focus on pending transactions
+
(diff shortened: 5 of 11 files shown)


packages/caravan-clients/src/types.ts
```diff
@@ -143,18 +143,16 @@ export interface ListTransactionsItem {
   abandoned?: boolean; // Only available for 'send' category
 }
 
-export interface BitcoindWalletParams<T> {
-  baseUrl: string; // Base URL of the Bitcoin node
+export interface BitcoindWalletParams {
```
Legend101Zz, line 146: So .. here the issue is that the function `callBitcoindWallet` which uses the `BitcoindWalletParams` is not generic and due to which we have to have the `//@ts-expect-error Will Fix this`

https://github.com/caravan-bitcoin/caravan/blob/ca9c6e85d1f953758a1632391544d913faa637f8/packages/caravan-clients/src/wallet.ts#L33-L36

Now I'd we'd want the `callBitcoindWallet` to be generic which is also a core part of the issue #192 
> 4. Make Core Functions Generic

So ... as the scope of this PR is limited so we can Go with this change for now ... but I'd like to if possible think way of solving it .. :)

packages/caravan-clients/src/types.ts
```diff
@@ -315,3 +313,19 @@ export interface UTXOUpdates {
...
+export interface BaseBitcoindParams {
+  url: string;
+  auth: {
+    username: string;
+    password: string;
+  };
+  walletName?: string;
+}
+
+export interface ListUnspentResponse {
```
Legend101Zz, line 326: I can see that there are two different definitions of `ListUnspentResponse` in type.ts and the `ListUnspentResponse` that was already in type.ts was used in other places of package like `bitcoind.ts` so you would need to think of a way to export it to wallet.ts file ... maybe use the one we already have in `type.ts` has all fields needed for the `ListUnspentResponse` used in wallet.ts so maybe you can take a subset of it

packages/caravan-clients/src/types.ts
```diff
@@ -315,3 +313,19 @@ export interface UTXOUpdates {
   fetchUTXOsError: string;
   addressKnown?: boolean;
 }
+
+export interface BaseBitcoindParams {
```
Legend101Zz, line 317: again a conflict between `BaseBitcoindParams` and `BitcoindParams` of type.ts ...

Legend101Zz (approved):
Thank for the changes @ice-009 , I'd added few reviews ... feel free to ask anything :)

packages/caravan-clients/src/types.ts
ice-009: Will tackle the this issue in the next part/task of the issue. Need a bit more thinking about this issue.

ice-009:
@Legend101Zz I have made a several changes to it. Kindly review the changes.

packages/caravan-clients/src/types.ts
```diff
@@ -314,18 +314,4 @@ export interface UTXOUpdates {
...
-}
-
-export interface ListUnspentResponse {
-  txid: string;
-  amount: number;
-  confirmations: number;
-  vout: number;
-}
\ No newline at end of file
+export interface ListUnspentResponseSubset extends Pick<ListUnspentResponse, 'txid' | 'amount' | 'confirmations' | 'vout'> {}
```
Legend101Zz, line 317: Why create a separate type ?

Can we just use the `ListUnspentResponse` in `wallet.ts` like this :

```ts
result: Pick<ListUnspentResponse, 'txid' | 'amount' | 'confirmations' | 'vout'>[];
} = await callBitcoindWallet({ baseUrl: url, auth,
```

packages/caravan-clients/src/wallet.ts
```diff
@@ -31,7 +31,7 @@ export function bitcoindWalletInfo({
   url,
   auth,
   walletName,
-}: BaseBitcoindParams) {
+}: BitcoindParams) {
```
Legend101Zz, line 34: looks good , thanks

Legend101Zz (approved):
Thanks for the updates! Everything looks good now—solid PR! Clean, concise, and easy to review :)

Looping in @bucko13 for any additional comments or suggestions. 😊

packages/caravan-clients/src/types.ts
```diff
@@ -70,13 +91,31 @@ interface RawTxOutput {
...
+ * Interface representing raw transaction status
+ */
 interface RawTxStatus {
   confirmed: boolean;
   block_height?: number;
   block_hash?: string;
   block_time?: number;
 }
 
+export type RPCParam =
```
Legend101Zz, line 104: Nice changes .. :)
Not a blocking comment but I'd like you to just remember that `RPCParam` would become important as it's used by `RPCRequest` which is used by the main bitcoind fn `callBitcoind` :

```ts
export async function callBitcoind<T>(
  url: string,
  auth: AxiosBasicCredentials,
  method: string,
  params: unknown[] = [],
): Promise<RPCResponse<T>> {
  if (!params) params = [];

  // Build our RPC request object
  const rpcRequest: RPCRequest = {
    jsonrpc: "2.0",
    id: 0, // We use a static ID since we're not batching requests
    method: `${method}`,
    params,
  };
```

and as it still just assumes the params to of type `unknown` we want in future to update that by the new `RPCParams` and then handle type check-in in other parts of the codebase accordingly

Legend101Zz (approved):
Left another round of review comments :)

PR #206, opened 2025-03-26
Feat/add get wallet transaction

Legend101Zz:
# Add support for wallet transaction fee data in private nodes

## Problem
When using Caravan with a local regtest node (or any private node), transaction fees were not showing up in the UI. This is because the current implementation uses the `getrawtransaction` RPC call, which doesn't include fee information, rather than the wallet-specific `gettransaction` call.

## Solution
This PR adds a new method `getWalletTransaction` that uses the Bitcoin Core `gettransaction` RPC call to retrieve fee information for transactions tracked by the wallet. 

- Added `bitcoindGetWalletTransaction` function for direct RPC access
- Added `getWalletTransaction` method to the BlockchainClient class
- Added tests for above methods

### DIFF
--- a/packages/caravan-clients/src/client.ts
+++ b/packages/caravan-clients/src/client.ts
@@ -17,6 +17,7 @@ import {
   bitcoindImportDescriptors,
   bitcoindListUnspent,
   bitcoindWalletInfo,
+  bitcoindGetWalletTransaction,
 } from "./wallet";
 import BigNumber from "bignumber.js";
 import {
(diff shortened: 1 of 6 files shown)


packages/caravan-clients/src/client.ts
```diff
@@ -507,11 +573,72 @@ export class BlockchainClient extends ClientBase {
...
+    txid: string,
+    forceRawTx: boolean = false,
+  ): Promise<TransactionDetails> {
     try {
       let txData: RawTransactionData;
 
       if (this.type === ClientType.PRIVATE) {
+        // For private nodes, try wallet transaction first (which includes fee data)
+
+        // unless forceRawTx is true
```
bucko13, line 629: nice

bucko13 (approved):
Good stuff! Can confirm it works as expected!

PR #225, opened 2025-04-03
Transaction Value Column 

### LINKED ISSUE #221: # Value Column Integration for Transaction Table
# Value Column Integration for Transaction Table

## Overview
This issue addresses the implementation of a transaction value column in the Caravan transactions table. The value column will display the net effect of each transaction on the wallet balance, providing users with clear visibility into how transactions impact their funds.

## Background
As part of the [Fee Package Integration Roadmap #219 , we need to enhance the transaction display with more relevant information for users. The value column is a critical component that will help users understand the financial impact of their transactions.

## Current Limitations
Currently, the transaction table lacks a dedicated column showing the transaction's value to the wallet. This makes it difficult for users to quickly assess how a transaction affects their balance without manually calculating input and output values.

### Functionality
- [ ] Calculate the net value of transactions relative to the wallet (inputs - outputs - fees)
- [ ] Handle both incoming and outgoing transactions correctly
- [ ] Display positive values for incoming transactions and negative values for outgoing transactions
- [ ] Support proper calculation even with change outputs
- [ ] Implement proper handling for transactions where not all inputs/outputs belong to the wallet

## Related Issues
- Parent issue : #219 

## References
- #168

Legend101Zz:
# Overview  
This PR adds a new **"Value"** column to the transaction table that displays the net effect of each transaction on the wallet balance. The value is displayed as a signed amount (**positive for incoming, negative for outgoing**) with **green/red color-coding** for easy visual identification.  

This fixes #221 

Screenshot 2025-04-04 at 01 20 41

## Purpose  
Previously, users had to manually deduce how transactions affected their wallet by analyzing inputs and outputs. This enhancement provides immediate visibility into:  

- **How much was received** in incoming transactions  
- **How much was spent** (including fees) in outgoing transactions  
- **Net balance impact** of complex transactions (e.g., with change outputs)  

#### New Fields in Transaction Type  
- Added `valueToWallet` field to transaction interfaces  

#### Value Calculation Logic  
- Created modular helper functions for different transaction scenarios  
- Handles both **private and public client** data formats  
- Uses **wallet address detection** to calculate accurate values  
- Implements **fallback strategies** when complete input data isn't available  

#### Value Display Component  
- Shows value in both **satoshis and BTC**  
- **Color-codes** based on transaction direction  
- Includes **comprehensive tooltips** to help users understand the values  

#### Sorting Support  
- Added ability to **sort transactions by value amount**  

## Calculation Approach  

The implementation uses a **smart, multi-strategy** approach to calculate transaction values:  

### For Private Client Transactions with Details  
- Uses **bitcoind's detailed categorization** for the most accurate value  

### For Transactions with Complete Input/Output Data  
- Evaluates which **inputs/outputs** belong to the wallet  
- Calculates:  
  ```math
  (outputs to wallet) - (inputs from wallet)
  ```
  ### For Transactions with Limited Input Data  
- Uses **transaction direction flags (`isReceived`)**  
- **For incoming:** Sums all outputs to wallet addresses  
- **For outgoing:** Calculates spent amount including fees


apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -30,6 +36,200 @@ const hasCompleteInputData = (tx: any): boolean => {
...
+  let valueToWallet = 0;
+
+  // Sum all transaction details affecting the wallet
+  details.forEach((detail: any) => {
+    if (
+      detail.category === "receive" ||
+      detail.category === "generate" ||
+      detail.category === "immature"
+    ) {
+      valueToWallet += detail.amount * 100000000; // Convert BTC to satoshis
```
bucko13, line 53: there's a satoshiToBtc converter in @caravan/bitcoin you can use.

Legend101Zz: Ahh .. yes I always forget about it 😅

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -30,6 +36,200 @@ const hasCompleteInputData = (tx: any): boolean => {
...
+
+  // CASE 1: Private client with details array - most accurate calculation
+  if (tx.details && Array.isArray(tx.details)) {
+    return calculateValueFromDetails(tx.details);
+  }
+
+  // CASE 2: Public client or private client without details field
+  if (tx.vin && tx.vout) {
+    // Calculate sum of all outputs to wallet addresses
+    const walletOutputsSum = calculateWalletOutputsSum(tx, walletAddresses);
```
bucko13, line 212: for outputs we can call this `totalChange` or something like that. Outputs that go back to the wallet are known colloquially as change.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -30,6 +36,200 @@ const hasCompleteInputData = (tx: any): boolean => {
...
+  let walletOutputsSum = 0;
+
+  // Sum all inputs from our wallet
+  for (const input of tx.vin) {
+    if (
+      input.prevout &&
+      input.prevout.scriptpubkey_address &&
+      walletAddresses.includes(input.prevout.scriptpubkey_address)
+    ) {
+      walletInputsSum += outputValueToSatoshis(input.prevout.value);
```
bucko13, line 90: prevout.value will always exist?

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -30,6 +36,200 @@ const hasCompleteInputData = (tx: any): boolean => {
...
+
+  return typeof value === "string"
+    ? parseFloat(value) * 100000000
+    : value * 100000000;
+};
+
+/**
+ * Calculate value based on inputs and outputs when we have complete data
+ */
+const calculateValueFromCompleteData = (
```
bucko13, line 76: This is an aesthetic comment so not hard blocker, but I think you could make this function a little more concise by using some functional method chaining. using a `.filter` and a `.reduce` rather than a `for` + `if` `+=` might be cleaner.

bucko13: it's possible this way is more performant though, so I'm not totally sure. at least the `if` could be moved inside of the reduce even though i feel like filtering first is nicer to read.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -30,6 +36,200 @@ const hasCompleteInputData = (tx: any): boolean => {
...
 
+/**
+ * Calculate transaction value from transaction details (private client)
+ * This gives the most accurate calculation as bitcoind provides detailed category information and amount field
+ */
+const calculateValueFromDetails = (details: any[]): number => {
+  let valueToWallet = 0;
+
+  // Sum all transaction details affecting the wallet
+  details.forEach((detail: any) => {
```
bucko13, line 47: could use a reduce here as well.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -30,6 +36,200 @@ const hasCompleteInputData = (tx: any): boolean => {
...
+  tx: any,
+  walletAddresses: string[],
+): number => {
+  let walletOutputsSum = 0;
+
+  if (!tx?.vout || !Array.isArray(tx.vout)) return 0;
+
+  for (const output of tx.vout) {
+    const outputAddress = output.scriptPubkeyAddress;
+    if (outputAddress && walletAddresses.includes(outputAddress)) {
```
bucko13, line 156: if you turn this into a simple filter function, then you can reuse the function in a call to `.filter` each time you use this. again, might be slightly less performant but could be nice to have 
```
outputTotal = tx.output.filter((output) => belongsToWallet(output, walletAddresses)).reduce(...)
```
and much of this could be reused.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: Yes it should always exist for both public and private clients

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: Thanks for the suggestion .. I have done the [changes](https://github.com/caravan-bitcoin/caravan/pull/225/commits/591bed2a1be8acf3a9e194d86bb62f8d21c8eee0)

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: Ahh .. Understood ... learned something new ... changes [done](https://github.com/caravan-bitcoin/caravan/pull/225/commits/5fcbca73eab28579adaf447ca4710deca8a6726b)

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/225/commits/b5e49436e38dce6ede255e18ebf46a2c56315b3a#diff-511018b155f4393fa6f4bc530aff6da3ac23c2527ff6ce2af32404419e4cf4b6R90)

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: Thanks for the suggestion ... it's looks much better .. changes [done](https://github.com/caravan-bitcoin/caravan/pull/225/commits/b5e49436e38dce6ede255e18ebf46a2c56315b3a#diff-511018b155f4393fa6f4bc530aff6da3ac23c2527ff6ce2af32404419e4cf4b6R154)

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,211 @@ import { useGetClient } from "../../../hooks/client";
...
+ */
+const calculateValueFromDetails = (details: any[]): number => {
+  return details.reduce((valueToWallet, detail) => {
+    const amountInSats = Number(bitcoinsToSatoshis(detail.amount));
+
+    if (
+      detail.category === "receive" ||
+      detail.category === "generate" ||
+      detail.category === "immature" ||
+      detail.category === "send" // Amount is already negative for send
```
bucko13, line 65: total aesthetic nit, but another way to do this that is a little less verbose would be:

```suggestion
      ["receive", "generate", "immature", "send"].includes(detail.category)
```

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,211 @@ import { useGetClient } from "../../../hooks/client";
...
+  }, 0);
+};
+
+/**
+ * Convert output value to satoshis, handling both string and number formats
+ */
+const outputValueToSatoshis = (value: any): number => {
+  if (typeof value === "undefined" || value === null) return 0;
+  return Number(bitcoinsToSatoshis(value));
+};
```
bucko13, line 80: I think we can make this more concise and better typed. Something like this
```suggestion
const outputValueToSatoshis = (value?: number | string): number => 
  bitcoinsToSatoshis(value ?? 0);
```

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,211 @@ import { useGetClient } from "../../../hooks/client";
...
+ * Calculate value based on inputs and outputs when we have complete data
+ */
+const calculateValueFromCompleteData = (
+  tx: any,
+  walletAddresses: string[],
+): number => {
+  // Sum all inputs from our wallet
+  const walletInputsSum = tx.vin
+    .filter(
+      (input: any) =>
```
bucko13, line 92: do we not have types for this now? and the tx type?

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,211 @@ import { useGetClient } from "../../../hooks/client";
...
+
+  // Net value = outputs to wallet - inputs from wallet
+  return totalChange - walletInputsSum;
+};
+
+/**
+ * Estimate transaction value when we only have outputs (incomplete input data)
+ */
+const estimateValueFromOutputs = (
+  tx: any,
```
bucko13, line 114: same. don't we have a type for this?

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,211 @@ import { useGetClient } from "../../../hooks/client";
...
+ * Instead, we use a combination of available data and heuristics to provide a good
+ * estimate of transaction value without these additional network requests.
+ */
+
+/**
+ * Calculate the net value of a transaction to the wallet
+ * Returns value in satoshis (positive for incoming, negative for outgoing)
+ */
+const calculateTransactionValue = (
+  tx: any,
```
bucko13, line 194: type?

bucko13 (changes requested):
Very nice! great cleanup. very concise and readable. A couple more aesthetic comments and then one request for types if possible at this stage.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: we do have types for Input and Outputs in the clients package but they are not exported ... also the Input type does not have all the fields we need ... I also wanted to have the TransactionTab types properly referenced from the clients package ... but I guess until unless we resolve #192 properly .. :(

Legend101Zz: I have added type fo the tx fields

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: Thanks for the feedback changes [done](https://github.com/caravan-bitcoin/caravan/pull/225/commits/e64fa4d2765afe975bb0b7a0abf69c8ceca9abb5)

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: added [type](https://github.com/caravan-bitcoin/caravan/pull/225/commits/4c2dd17575f077883f4b012c935ed57a9d20885e)

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
Legend101Zz: added type[https://github.com/caravan-bitcoin/caravan/pull/225/commits/4c2dd17575f077883f4b012c935ed57a9d20885e]

bucko13: looks like this one didn't get added?

Legend101Zz: Ah yes — sorry for not pointing this out earlier. It’s a bit tricky to define a strict type for this at the moment. Ideally, this is something we should revisit during a more structured refactor, once we’ve consolidated types from a dedicated source like the clients package or a shared types module.

Right now, the `Transaction` type used in the hooks was defined before we introduced the `fees` and `value` columns. With those additions, several fields have become optional or conditionally present — like in this case, where the `calculateTransactionValue` function expects a `details` field on the transaction object.

I initially considered something like `Transaction & { details: any[] }`, but that’s more of a patch than a long-term solution. What we really need is a centralized type definition — possibly a more complete `Transaction` base type — from which we can derive partials or specific view models as needed.

For now, I’d suggest we keep it flexible, but definitely flag this for cleanup once we finalize the shared type structure.

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,213 @@ import { useGetClient } from "../../../hooks/client";
...
+): number => {
+  // If transaction is explicitly marked as received or we have outputs to our wallet addresses
+  if (tx.isReceived && totalChange > 0) {
+    return totalChange;
+  }
+
+  // If transaction is explicitly marked as sent
+  if (!tx.isReceived) {
+    // At minimum, we spent the fee
+    let spentAmount = tx.fee ? tx.fee : 0;
```
bucko13, line 126:
```suggestion
    let spentAmount = tx.fee ?? 0;
```

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
```diff
@@ -18,6 +21,213 @@ import { useGetClient } from "../../../hooks/client";
...
+const calculateTransactionValue = (
+  tx: any,
+  walletAddresses: string[],
+): number => {
+  // Skip calculation if tx is invalid
+  if (!tx) return 0;
+
+  // CASE 1: Private client with details array - most accurate calculation
+  if (tx.details && Array.isArray(tx.details)) {
+    const fees = tx.fee ? tx.fee : 0;
```
bucko13, line 204:
```suggestion
    const fees = tx.fee ?? 0;
```

apps/coordinator/src/components/Wallet/TransactionsTab/hooks.ts
bucko13: agreed. thanks for the clarification!

bucko13 (changes requested):
small nit otherwise good to go!

bucko13 (approved):

Issue #125, opened 2024-09-05
Add automatic documentation generation


bucko13:
The typescript and jsdocs code comments will need to be improved overtime but I think some value could be added from having the generation in place. Ideally it's also added to CI so new docs are generated on merges to main. 

I don't have a strong opinion on what library to use for this. It'll depend on what's most stable and used in the community but something like [ts-docs](https://github.com/ts-docs/ts-docs) seems like a good option.

asmit27rai:
@bucko13 I Want to work on this issue.

PR #227, opened 2025-04-07
feat: added automated docs generation using typedoc

krrish-sehgal:
Fixes: https://github.com/caravan-bitcoin/caravan/issues/125 

## 📝 PR Description: Add TypeDoc Support + CI Integration

This PR adds TypeDoc configuration and GitHub Actions support to generate and auto-commit documentation for the packages in this monorepo.

---

#### `typedoc.json`
Configuration for [TypeDoc](https://typedoc.org/), the tool used to generate HTML documentation from TypeScript JSDoc comments.

- **Includes entry points for all packages**: 
  - `bip32`, `caravan-bitcoin`, `caravan-clients`, `caravan-fees`, `caravan-health`, `caravan-psbt`, `caravan-wallets`, `multisig`
- Outputs to the `docs/` folder.
- Excludes internal/private/external declarations for a cleaner output.

#### `tsconfig.typedoc.json`
Dedicated `tsconfig` for controlling how TypeDoc compiles the code.

- Ensures only relevant `src/index.ts` files are processed
- Avoids emitting unnecessary files
- Supports strict typing and ES2020 modules
- Prevents noise from tests and build artifacts

---

### ⚙️ CI Changes (`.github/workflows/ci.yml`)

- Adds steps to **generate TypeDoc on every push to `main`**
- If `docs/` content has changed:
  - It **commits** the updated docs back to `main` automatically
- Uses `[skip ci]` in commit message to prevent looped CI runs
- GitHub Actions bot identity is used for auto-commits

#### ✨ Benefits:
- Always up-to-date documentation
- No need to run docs manually
- Improves dev onboarding and API clarity

---

### ✅ Example Usage

```bash
# To test docs locally
npm run docs



Images: 

<img width="1470" alt="Screenshot 2025-04-07 at 8 50 59 PM" src="https://github.com/user-attachments/assets/ce23861b-850b-48cd-ba3c-edbc9210497e" />

<img width="1467" alt="Screenshot 2025-04-07 at 8 52 24 PM" src="https://github.com/user-attachments/assets/ae02a6e8-dcb3-432a-b423-3d1b619af1f8" />

### CODE REVIEW

**Comment by krrish-sehgal** [comment]
Since its just package additions , I am not sure what to add as a changeset. 
Will wait for instructions. 

Thank you.

File: tsconfig.typedoc.json [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "include": [
+      "packages/bip32/src/index.ts",
+      "packages/caravan-bitcoin/src/index.ts",
+      "packages/caravan-clients/src/index.ts",
+      "packages/caravan-fees/src/index.ts",
+      "packages/caravan-health/src/index.ts",
+      "packages/caravan-psbt/src/index.ts",
+      "packages/caravan-wallets/src/index.ts",
+      "packages/multisig/src/index.ts"
```
bucko13, line 26: why not just `packages/**` or `packages/**/index.ts`?

.github/workflows/ci.yml
```diff
@@ -1,7 +1,10 @@
-name: Test
+name: Test & Docs
```
bucko13, line 1: why not make this a separate workflow?

bucko13:
note that this does not include the build artifacts for the docs but it is also not git ignored. So if someone runs this locally the git diff will be huge the first time. Either we can include this in version control, or you can do some git and scripting magic to have it publish to a special docs branch that keeps it in git history.

bucko13:
Also, what do you think about using the `typedoc-plugin-markdown` plugin?

.github/workflows/ci.yml
krrish-sehgal: Makes sense, I have made the changes.

krrish-sehgal:
> note that this does not include the build artifacts for the docs but it is also not git ignored. So if someone runs this locally the git diff will be huge the first time. Either we can include this in version control, or you can do some git and scripting magic to have it publish to a special docs branch that keeps it in git history.

Sure sir, I have added docs/ in .gitignore and  made a new docs.yml file that creates a new branch , docs, wipes out everything and copies the docs from the main eachtime a commit is done and force pushes it to docs branch , so that all the version of the docs can be seen in the git history. 

Let me know if this was expected, or zipping the docs and uploading it as a GitHub Actions artifact for download instead of committing.

krrish-sehgal:
> Also, what do you think about using the `typedoc-plugin-markdown` plugin?

Sir, it’s definitely a valid option, but for our use case—since we don’t store docs on the main branch—it’s more convenient to have a quick web UI to browse through, which offers a better interface. It can even be hosted via GitHub Pages since it’s static. With Markdown, users would need to use Docusaurus or browse the raw files on GitHub, which isn’t as user-friendly

krrish-sehgal:
docs.yml is working: 
Screenshot 2025-04-10 at 5 23 29 PM

Legend101Zz:
Hi @krrish-sehgal thanks alot for working on it ... docs was something we wanted to fix for a long time ... I'll throw a full review on this tomorrow and maybe then if you have some time we can try to get it merged ?

tsconfig.typedoc.json
```diff
@@ -0,0 +1,30 @@
...
+      "packages/typescript-config",
+      "**/__tests__/**",
+      "**/*.test.ts",
+      "**/*.test.js",
+      "**/node_modules",
+      "docs",
+      "temp"
+    ]
+  }
+
```
Legend101Zz, line 30: ---

**Non-blocking suggestion:**

Hey @krrish-sehgal @bucko13 — while testing the docs build, I noticed a few warnings due to unsupported JSDoc tags that TypeDoc doesn't recognize by default:

```
./packages/caravan-fees/src/transactionAnalyzer.ts:313:5 - [warning] Encountered an unknown block tag @note
./packages/caravan-fees/src/constants.ts:36:3 - [warning] Encountered an unknown block tag @constant
...
```

These tags (`@note`, `@constant`) aren't part of the standard TypeDoc configuration, which causes noisy logs during `typedoc` runs.

###  Potential solutions:

#### Option 1: Whitelist custom tags

Update `typedoc.json` to explicitly support these tags:

```json
"blockTags": [
  "@note",
  "@constant",
  "@constructor",
  "@description",
  "@example",
  "@see",
  "@returns",
  "@param",
  "@throws"
]
```

This would be the cleanest solution if we intend to consistently use these tags across the repo.

#### Option 2: Suppress warnings globally

```json
"logLevel": "Error"
```

This avoids clutter but could hide other helpful warnings, so maybe not ideal.

---

To prevent future inconsistencies and maintain JSDoc hygiene across the codebase, we should consider adding [`[eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc)`](https://github.com/gajus/eslint-plugin-jsdoc) with scoped rules. For example:

```json
"jsdoc/check-tag-names": ["error", {
  "definedTags": ["remarks", "example", "param", "returns", "see", "throws", "typeParam", "deprecated"]
}],
"jsdoc/no-undefined-types": "error",
"jsdoc/require-description": "error",
"jsdoc/check-alignment": "error"
```

Let me know what you think — happy to PR a docs-linting setup if this sounds useful!

typedoc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "entryPoints": ["packages/**/src/index.ts"],
+  "tsconfig": "tsconfig.typedoc.json",
+  "out": "docs",
+  "includeVersion": true,
+  "excludePrivate": true,
+  "excludeInternal": true,
+  "excludeExternals": true
```
Legend101Zz, line 8: ---

May I suggest we switch to using the [`[typedoc-theme-hierarchy](https://www.npmjs.com/package/typedoc-theme-hierarchy)`](https://www.npmjs.com/package/typedoc-theme-hierarchy) custom theme for our documentation?

The hierarchical structure looks significantly cleaner and more intuitive—especially for newcomers trying to navigate the codebase. It organizes modules and symbols in a way that feels much easier to traverse.

Also, can we please add the following fields to the `typedoc.json`?

```json
  "name": "Caravan Documentation",
  "favicon": "./assets/images/caravan-logo-transparent.png"
```

* `name` ensures our docs have a proper title
* `favicon` sets the small browser tab icon (I noticed it in the screenshots you shared, but didn't find it explicitly configured—maybe some custom CSS was involved?)

With these additions, the updated config using the `hierarchy` theme looks like this:

```json
{
  "entryPoints": ["packages/**/src/index.ts"],
  "tsconfig": "tsconfig.typedoc.json",
  "out": "docs",
  "includeVersion": true,
  "excludePrivate": true,
  "excludeInternal": true,
  "excludeExternals": true,
  "name": "Caravan Documentation",
  "plugin": ["typedoc-theme-hierarchy"],
  "theme": "hierarchy",
  "favicon": "./assets/images/caravan-logo-transparent.png"
}
```

Sharing a screenshot of how it renders on my end:

Screenshot 2025-05-15 at 15.02.31

Let me know what you think!

---

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+          # Save the generated docs to a temp location
+          mv docs docs_temp
+          
+          # Create and switch to an orphan docs branch
+          git checkout --orphan docs
+          
+          # Clear existing contents but preserve git directory
+          git rm -rf . || true
+          
+          # Copy the new docs in
```
Legend101Zz, line 50: This looks awesome to me :)

Legend101Zz (changes requested):
@krrish-sehgal thanks for the amazing work ... left a few comments and also some suggestions

typedoc.json
```diff
@@ -0,0 +1,9 @@
...
+}
```
Legend101Zz, line 9: ---

Looking at the screenshot you shared, it seems like you were able to render the main `README.md` as the landing page of the docs. However, I noticed there’s no explicit configuration for that in the current `typedoc.json`. To ensure consistent behavior, we should explicitly include:

```json
"readme": "./README.md"
```

in the `typedoc.json` config. This ensures that the root README is used as the docs entry point across all environments.

Also, I was wondering if we could extend this approach so that each individual package in the monorepo uses its own `README.md` as the landing section of its generated documentation. That said, setting this up cleanly across all packages seems a bit tricky with the current TypeDoc setup — might require either custom scripts or breaking the doc generation down package by package. Curious to hear your thoughts on that.

Legend101Zz: @bucko13 do you have any suggestions on how can we get `README.md` of each package included too ?

tsconfig.typedoc.json
bucko13: If it's not widespread I'd love to just update the tags rather than hide warnings/errors. if they're not being used anyway there's no point to them.

typedoc.json
krrish-sehgal: I added the README configuration, but from what I can tell, the only way to have each package display its own README is to generate separate documentation for each package. This is because TypeDoc only supports a single README per run. However, I don't think that would be an ideal approach, since users would then need to constantly switch between different documentation outputs to access other packages. It would also break the unified view we currently have, where everything is combined in one place.

.github/workflows/docs.yml
krrish-sehgal: Thanks.

typedoc.json
krrish-sehgal: this is a valuable addition , i made the change!

tsconfig.typedoc.json
```diff
@@ -0,0 +1,30 @@
...
+      "target": "es2020",
+      "allowJs": false,
+      "declaration": true,
+      "emitDeclarationOnly": true,
+      "skipLibCheck": true,
+      "strict": true,
+      "noEmit": false,
+      "outDir": "temp",
+      "baseUrl": ".",
+    "paths": {
```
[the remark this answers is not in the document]
krrish-sehgal: done.

tsconfig.typedoc.json
krrish-sehgal: This is still remaining , I was able to bring down the warnings from 44 to 18 though.

krrish-sehgal:
@Legend101Zz  one issue i have been facing is figuring out why the package-lock diff is of 10k lines with just 2 extra dependencies of typedoc and typedoc-theme-hierarchy.

Rest of the things according to me are done.

typedoc.json
Legend101Zz: Yup ... understood we can leave this for now then and just have the whole package readme added ... as was before

tsconfig.typedoc.json
Legend101Zz: @bucko13 I had two concerns ... here ..

1) If we are doing changes to packages ... then do we need corresponding changesets to them ... ?
2) I guess we can leave this for now as it's only warning due to writing JSDOC style comments and us using TSDOC ... so maybe we can have some-tooling to handle that later ?

bucko13: yeah docs shouldn't require a changeset as there are no publishable changes.

Legend101Zz:
Hey @krrish-sehgal , quick bump on this! Just a few final comments left — would be great if you could take a look when you get the time :)

Legend101Zz:
@bucko13 bump on this , also maybe I'll test it out locally first and see if it works as expected .... @krrish-sehgal do you merging the main changes to this branch ?

Legend101Zz:
@krrish-sehgal @bucko13 tested this out on latest pull , everything works superbly for me :) 

I guess we can merge this now ? 

Screenshot 2025-11-15 at 23 38 51
Screenshot 2025-11-15 at 23 39 11

Legend101Zz (approved):
Looks great to me just need to test the ci part :)

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
-          cache: 'npm'
+          cache: "npm"
```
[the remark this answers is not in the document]
bucko13: exactly

PR #228, opened 2025-04-08
Migrates @caravan/bip32 to vitest from jest

bucko13:


bucko13:
@coderabbitai what's your review

PR #229, opened 2025-04-08
fix: removed setters and added private functions to handle count updates

### LINKED ISSUE #141: Remove setters from PsbtV2.PSBT_GLOBAL_INPUT_COUNT and PsbtV2.PSBT_GLOBAL_OUTPUT_COUNT
The psbtv2 key `PSBT_GLOBAL_INPUT_COUNT` and `PSBT_GLOBAL_OUTPUT_COUNT` represent the number of inputs or outputs on the psbt. On the `PsbtV2` class, these values have setters, however, because setters are public, they might cause these values to disagree with the real number of inputs or outputs tracked in the psbt.

The two values must remain in sync with the number of inputs or outputs being tracked in the map.

I think the easiest way to do this is to remove the setters and only allow the add and delete input/output methods to modify the value of these fields. However, this is a backward incompatible change for those that might require the existence of the setters.

krrish-sehgal:
fixes #141.


krrish-sehgal:
@coderabbitai , can you review the PR and see if i have missed anything or it breaks certain code?

packages/caravan-bitcoin/src/psbtv2.ts
Shadouts, line 1: Don't update this module. `PsbtV2` from caravan-bitcoin is deprecated.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -1428,4 +1418,23 @@ export class PsbtV2 extends PsbtV2Maps {
...
   }
-}
+
+  /**
+   * Updates the PSBT_GLOBAL_INPUT_COUNT and PSBT_GLOBAL_OUTPUT_COUNT fields
+   * in the global map.
+   */
+  private updateGlobalInputCount() {
+    const bw = new BufferWriter();
+    bw.writeU8(this.inputMaps.length); // or `.size` if inputMaps is a Map
```
Shadouts, line 1428: Concerning the comment, `inputMaps` is actually an array of maps. Are you suggesting it might be converted to a map?

Shadouts (changes requested):
Overall, looks good, but the new methods should be tested.

Although this is technically a major version increment, which is why I originally refrained from making the change myself, I don't think it's likely that anyone relied on these setters since doing so would have resulted in a mostly-broken PSBT. Even the PSBT protocol allows for this breakage, but I imagine actual consumers of the PSBT probably iterate through actual inputs/outputs and don't rely on the count fields due to this possibility.

packages/caravan-bitcoin/src/psbtv2.ts
krrish-sehgal: noted.

krrish-sehgal:
> Overall, looks good, but the new methods should be tested.
> 
> Although this is technically a major version increment, which is why I originally refrained from making the change myself, I don't think it's likely that anyone relied on these setters since doing so would have resulted in a mostly-broken PSBT. Even the PSBT protocol allows for this breakage, but I imagine actual consumers of the PSBT probably iterate through actual inputs/outputs and don't rely on the count fields due to this possibility.

Thanks for the feedback , I have made the necessary changes , let me know if the changeset needs to be updated to major. 
Yes sir, you are correct i thought of it as well, but i think it would be a good practice to not expose the setters like that.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
krrish-sehgal: Apologies, the comment was just helper , while i was understanding it. Not needed now.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
```diff
@@ -1794,3 +1794,51 @@ describe("PsbtV2.toV0", () => {
...
+    });
+
+    // Use reflection to call the private method
+    (psbt as any).updateGlobalInputCount();
+
+    // Verify the global input count
+    const inputCount = psbt.PSBT_GLOBAL_INPUT_COUNT;
+    expect(inputCount).toBeDefined();
+    expect(inputCount).toBe(2);
+  });
```
Shadouts, line 1823: I'm not sure this test is properly formulated. The expected results of this test would have occurred even before your change (your removing of the setters and adding of the updater method). So, it's not necessarily testing your newly added behavior.

Here is a test that specifically checks to see if `updateGlobalInputCount` functions as expected:
```
describe("PsbtV2.updateGlobalInputCount (private)", () => {
  let psbt: PsbtV2;

  beforeEach(() => {
    psbt = new PsbtV2();
  });

  it("Properly updates the input count", () => {
    (psbt as any).inputMaps.push(new Map());
    (psbt as any).inputMaps.push(new Map());
    (psbt as any).updateGlobalInputCount();
    expect(psbt.PSBT_GLOBAL_INPUT_COUNT).toBe(2);
  });
});
```

And your test could be modified to check that `addInput` calls `updateGlobalInputCount` as expected:
```suggestion
describe("PsbtV2 addInput", () => {
  let psbt: PsbtV2;
  const updateGlobalInputCountSpy = vi.spyOn(
    PsbtV2.prototype as any,
    "updateGlobalInputCount",
  );

  beforeEach(() => {
    vi.clearAllMocks();
    psbt = new PsbtV2();
  });

  it("should call updateGlobalInputCount", () => {
    // Add inputs to the PSBT
    psbt.addInput({
      previousTxId: Buffer.alloc(32, 0), // Dummy txid
      outputIndex: 0,
    });
    psbt.addInput({
      previousTxId: Buffer.alloc(32, 1), // Dummy txid
      outputIndex: 1,
    });

    // Verify the global input count
    const inputCount = psbt.PSBT_GLOBAL_INPUT_COUNT;
    expect(inputCount).toBeDefined();
    expect(inputCount).toBe(2);
    // Once, for creation and once for each call to addInput
    expect(updateGlobalInputCountSpy).toHaveBeenCalledTimes(3);
  });
```

Of course, both would need to also be configured to test for output count. What do you think?

Shadouts (changes requested):
Please revert to a patch version. That was fine. I think I wasn't being clear.

Not as important, but I think the tests could be expanded upon since I don't think the test was sufficiently covering new behavior, but let me know what you think. I've provided a couple of examples. I don't want it to become a burden since this is surely a good change to have merged.

krrish-sehgal:
Thank you so much @Shadouts  and @bucko13 for your inputs , I'll make the changes and resolve all your points.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
krrish-sehgal: Yes sir , sorry bout that , I have fixed the tests for both, according to your inputs.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
```diff
@@ -1794,3 +1794,51 @@ describe("PsbtV2.toV0", () => {
...
+
+    // Use reflection to call the private method
+    (psbt as any).updateGlobalOutputCount();
+
+    // Verify the global output count
+    const outputCount = psbt.PSBT_GLOBAL_OUTPUT_COUNT;
+    expect(outputCount).toBeDefined();
+    expect(outputCount).toBe(2);
+  });
+});
```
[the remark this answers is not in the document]
krrish-sehgal: fixed in the latest commit.

.changeset/chilly-houses-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"@caravan/psbt": major
```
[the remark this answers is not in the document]
krrish-sehgal: fixed (reverted back to patch , in the latest commit).

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
```diff
@@ -1794,3 +1794,51 @@ describe("PsbtV2.toV0", () => {
...
+      previousTxId: Buffer.alloc(32, 0), // Dummy txid
+      outputIndex: 0,
+    });
+    psbt.addInput({
+      previousTxId: Buffer.alloc(32, 1), // Dummy txid
+      outputIndex: 1,
+    });
+
+    // Use reflection to call the private method
+    (psbt as any).updateGlobalInputCount();
```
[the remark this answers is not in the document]
krrish-sehgal: fixed. Thank you.

Shadouts (approved):
Great work!

PR #231, opened 2025-04-09
Migrates `@caravan/caravan-health` tests from jest to vitest

DhairyaMajmudar:
## Description

This PR migrates the `@caravan/caravan-health` test cases from jest to vitest. Consists the lint and format fixes following the new eslint config file `library.new.js`


packages/caravan-health/package.json
bucko13, line 1: let's add vitest to dev dependencies here.

packages/typescript-config/base.json
```diff
@@ -18,7 +18,6 @@
     "preserveWatchOutput": true,
     "skipLibCheck": true,
     "strict": true,
-    "outDir": "./dist",
```
DhairyaMajmudar, line 21: I've removed this field from the base ts config file because when I tried to build the package the very first time, it built successfully. But when I tried to rebuild the package it failed, throwing an error message it can't rewrite the existing dist files. Below is the error output.

image

bucko13 (changes requested):
beautiful! Just a couple quick comments but hopefully quick

packages/typescript-config/base.json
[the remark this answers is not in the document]
DhairyaMajmudar: Thank you for the quick reviews, I've answered it in this comment https://github.com/caravan-bitcoin/caravan/pull/231#discussion_r2035650033

packages/caravan-health/package.json
DhairyaMajmudar: vitest is installed globally in the monorepo, won't it be good to use that globally installed vitest only? This help make our lock file clean .

packages/typescript-config/base.json
bucko13: ah, ok. Thanks! good catch. I was actually seeing that too. annoying that the lint file then gives me errors in the IDE. Will have to look into that and figure out a different workaround.

packages/caravan-health/package.json
bucko13: great. thanks. could you also remove from bip32 package and just confirm that all still works. sorry and thank you!

DhairyaMajmudar: Sure thingy

bucko13 (commented):
Could you remove the now superfluous dependency from the bip32 package then we should be good to go

DhairyaMajmudar:
> Could you remove the now superfluous dependency from the bip32 package then we should be good to go

Oopsie, seems like that file didn't got into commit, it is now removed

bucko13 (approved):
Great!

Issue #113, opened 2024-07-15
Add "sign message" feature


tomunchained:
Some users may want to sign a message corresponding to a multisig address. For example, for someone to prove they still hold a key to that quorum without building a transaction.

A possible location for this feature within Caravan would be on the "Addresses" page. Upon clicking on an address, some details and options appear, including Redeem Script, Confirm on Device, UTXOs, etc. Perhaps another option for "Sign Message" could appear here.

jbrauck-unchained:
Picking this up in my spare time to see what I can do and also get comfortable with the libraries.

PR #238, opened 2025-04-10
fix: signing of a message working for ledger and tenzor

krrish-sehgal:
This fixes https://github.com/caravan-bitcoin/caravan/issues/113

1. Added the sign message option in UI
2. Once the message is signed , the signature is returned by caravan-wallet api and is verified with the address using bip322-js
3. The library bip322-js has been shifted to package level.

Demo: 
Screenshot 2025-04-10 at 2 40 32 PM

https://github.com/user-attachments/assets/3af50a47-1209-4a1a-962a-7e6b0e39c7a5

### DIFF
--- a/.changeset/twelve-eagles-admire.md
+++ b/.changeset/twelve-eagles-admire.md
@@ -1,0 +1,5 @@
+---
+"caravan-coordinator": minor
+---
+
+Added a backwards compatible message signing functionality for ledger and tenzor wallet using caravan-wallets
--- a/apps/coordinator/package.json
+++ b/apps/coordinator/package.json
@@ -115,6 +115,7 @@
     "base58check": "^2.0.0",
     "bignumber.js": "^9.0.0",
     "bip32": "^2.0.4",
+    "bip322-js": "^2.0.0",
     "bitcoin-address-validation": "^1.0.2",
     "bitcoinjs-lib": "^5.1.7",
     "bowser": "^2.6.1",
--- a/apps/coordinator/src/components/CreateAddress/AddressGenerator.jsx
+++ b/apps/coordinator/src/components/CreateAddress/AddressGenerator.jsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import {
@@ -19,7 +19,7 @@ import {
 } from "@mui/material";
 import { downloadFile } from "utils";
 import { externalLink } from "utils/ExternalLink";
-
+import SignMessageModal from "./SignMessageModal"; 
 // Actions
 import {
   sortPublicKeyImporters as sortPublicKeyImportersAction,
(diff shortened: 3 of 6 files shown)


krrish-sehgal:
Note : 
i wasnt able to test it with actual wallets, let me know if there a way to do that 
However I made sure the function running are working with each other just fine 
(used : https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/hw-app-btc#signmessage)

Legend101Zz:
@krrish-sehgal sorry for my lack of knowledge here but just wanted to confirm if this was this done for the QR-project or it's seperate ?

krrish-sehgal:
@Legend101Zz , Its separate.

tomunchained:
Hello @krrish-sehgal, I'd like to help test this and provide feedback. Fair warning: I'm not a software engineer and have limited experience in GitHub in general.

**First:** I've loaded up the most recent vercel deployment, and tried to test with a redeem script, but I'm getting an error `This keystore is not supported when exporting public keys.` I get this if I try Trezor or Ledger, either option. Do you know what might be causing this?

Screenshot 2025-06-12 at 5 52 23 PM


**Second:** Am I correct that this message signing feature is currently limited to the script explorer section of Caravan? Ideally I'd like to see it subtly built into the addresses within a wallet. For example, shown as an option "sign message" alongside "confirm on device" as shown in screenshot below. (Functionally, "sign message" could be built into the wallet itself, or maybe the button just takes the user to the script explorer and automatically loads in the redeem script. Either way, it will be more friendly to users who don't understand they'd need to manually copy the redeem script and take it to the script explorer.)

Screenshot 2025-06-12 at 5 57 56 PM


**Third:** I'd like to make sure this project includes all of the necessary features to be practically useful. The ultimate purpose is to demonstrate control over the funds in an address to a third party. Therefore, in a _k-of-n_ multisig quorum, the message will often be signed with _k_ number of keys, not just one key. Does BIP322 allow for multiple signatures to be grouped together like this? (I think so, but my understanding is fuzzy.)

Once _k_ number of keys have signed, there also needs to be a verification tool for the third party. The third party should be able to input the address, the message, and the signature (I don't think they need the redeem script, as it is built into the signature, if BIP322 works like how I understand BIP137 to work). Where this verification tool should exist, I'm not sure. The script explorer doesn't feel like the right place, if the third party is not required to enter a redeem script. Therefore I'm tempted to consider an entirely new menu item, called "sign/verify message" where all of this behavior could live, rather than within the script explorer. 

Screenshot 2025-06-12 at 6 28 46 PM


This is how Trezor builds it for BIP137:

Screenshot 2025-06-12 at 6 33 09 PM
Screenshot 2025-06-12 at 6 33 17 PM


Happy to hear your thoughts and discuss this more going forward!

Legend101Zz:
> needs some attention by me. So doesn't need to be closed just yet.

great I'll add the long lived tag

PR #239, opened 2025-04-10
An admittedly hack-y fix for using @caravan/psbt on the server side.

joelangeway:
Adds an ESBuild plugin that prepends every JS build artifact with a slap dash declaration of self so that problematic but inconsequential browser only code doesn't break server side use.

### DIFF
--- a/packages/caravan-psbt/provide-self.ts
+++ b/packages/caravan-psbt/provide-self.ts
@@ -16,7 +16,7 @@ if (typeof self === 'undefined') {
 const headerBytes = Uint8Array.from(Array.from(headerText, c => c.charCodeAt(0)))
 const n_header = headerBytes.length
 
-export default function ProvideSelf(): Plugin {
+export default function provideSelf(): Plugin {
   return {
     name: 'provide-self',
     setup(build) {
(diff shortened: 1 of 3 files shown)


packages/caravan-psbt/provide-self.ts
```diff
@@ -0,0 +1,41 @@
...
+  } else {
+    self = {}
+  }
+}
+`
+
+const headerBytes = Uint8Array.from(Array.from(headerText, c => c.charCodeAt(0)))
+const n_header = headerBytes.length
+
+export default function ProvideSelf(): Plugin {
```
bucko13, line 19: nit, but why the pascal case (and kebab case for the file)? Looks like the other plugins are camel cased?

joelangeway: I guess I'm coding with a React accent. Happy to change.

bucko13: yeah, even in react, Pascal is typically just for components (e.g. hooks are still camel for example). At least that's how I'm familiar with it.

bucko13:
Screenshot 2025-04-10 at 3 35 50 PM

Confirmed a psbt created in [the app](https://caravan-coordinator-ueuyvrrrw-unchained.vercel.app/#/wallet) still works. also that the polyfill in the built package also looks good. Thanks @joelangeway!

bucko13 (approved):

PR #241, opened 2025-04-11
feat: added xpub export through bcur2 qrcode

krrish-sehgal:
This is the proof of concept and these methods can be further used in caravan-wallet as complete new interaction classes to work with qr codes. 
To test out , use /wallet page (thats where I have currently integrated the Xpub import feature).

The interaction class built: 
1. Receives parts of a QR code (receivePart)
2. Tracks decoding progress
3. Validates and extracts structured data like xpubs
4. Has error handling for invalid QR data or unsupported types
5 .Manages internal state (error, progress, decoder)

Please check below for a full technical description of what this PR includes [below](https://github.com/caravan-bitcoin/caravan/pull/241#issuecomment-2799899147). 

https://github.com/user-attachments/assets/9b87802d-0546-4fcf-a9f6-62817aa654f0


krrish-sehgal:
This is really an abstract implementation, am working on making it more generic and add the logic to @caravan-wallet

krrish-sehgal:
Now I have shifted the logic, to the library in caravan-wallet , making it completely reusable ex: for signing when the signed PSBT needs to be imported back. 

Currently the library just contains the BC-UR2 decoder logic (since thats only what was needed for xpub import) , I will be adding the encoder logic to it as well. 

Thanks.

krrish-sehgal:
@coderabbitai , can you write a good summary of what this PR includes

Legend101Zz:
@krrish-sehgal I guess you can change this PR to a draft PR ... as it is POC a right ?

krrish-sehgal:
Yeah @Legend101Zz , sure , but @bucko sir, it'll be great if you can take a look at this and tell me if I am moving forward in the right direction , since the xpub import is a functionality that is separate from PSBT signing and message signing , and since its working ,i can give it some final touch ups in the UI(maybe generalise the  hermit UR component for scanning).

packages/caravan-wallets/tsup.config.ts
```diff
@@ -11,5 +11,5 @@ export default defineConfig({
   external: ["bitbox-api"],
   // noExternal makes sure that certain packages are bundled
   // in the final package rather than independently installed
-  noExternal: ["@caravan/psbt"],
+  noExternal: ["@caravan/psbt", "bignumber.js"],
```
bucko13, line 14: why was this needed? This will make the final package bigger.

package-lock.json
bucko13, line 1: I'm concerned seeing a package-lock diff but no change to any package.json

packages/caravan-wallets/tsup.config.ts
krrish-sehgal: Sir , I am using @keystonehq/bc-ur-registry which uses @ngraveio/bc-ur under the hood, 
This line was important, because @ngraveio/bc-ur uses require("bignumber.js") in an ES module context, so marking "bignumber.js" as noExternal ensures Vite bundles it to avoid incompatible dynamic requires at runtime.

krrish-sehgal: Screenshot 2025-04-16 at 10 21 24 AM

krrish-sehgal: I found an alternative for the error I was facing , this is fixed now.

package-lock.json
krrish-sehgal: Refer to [here](https://github.com/caravan-bitcoin/caravan/pull/241/files#r2046195273).

bucko13:
@krrish-sehgal we're getting some build errors in vercel when trying to create a deployment. Looks like prettier needs to be run (you can install a plugin for your IDE so this runs automatically on save). Once it's building it'll be easier to review and validate.

krrish-sehgal:
Sure sir, On it.

krrish-sehgal:
@bucko13 sir , Its deployed now , and I have also made sure the ci test passed on my local github actions.

bucko13:
This is very exciting. Thanks for starting things off. I definitely want to start reviewing and getting some testing in as well. I'll try and get to this within the next ~1 week.

packages/caravan-wallets/src/bcur2.ts
```diff
@@ -0,0 +1,135 @@
...
+import { ExtendedPublicKey, Network, BitcoinNetwork } from "@caravan/bitcoin";
+
+interface DecodedData {
+  type: string;
+  xpub: string;
+  xfp?: string;
+  path?: string;
+}
+
+export class BCURDecoder2 {
```
bucko13, line 11: one thing that we'll want to think about is how can we structure this like the other interactions? I.e. could this be built inheriting from the IndirectKeystore (note that Hermit is a signer that uses QR codes), and then you would have methods that match the other interactions.

krrish-sehgal: Okay, I will do this as next step , since I have this working.
I'll make a new QR scanner interaction class that inherits from IndirectKeystoreInteraction , and uses it. 
One thing that i want to clarify:  the present decoders and encoders in bcur.ts file won't work as they aren't supporting the crypto-* types , and that is required and are going for the legacy UR:Type/[Digest]/[Fragment].

bucko13: Yep, that's fine! I'm not surprised that something new is going to be required and using a dependency should be fine (might be good to lock the package version though just to be safe).

PR #242, opened 2025-04-11
Migrates `@caravan/caravan-psbt` tests from jest to vitest

DhairyaMajmudar:
## Description

This PR migrates the `@caravan/caravan-psbt` test cases from jest to vitest.


packages/caravan-psbt/README.md
```diff
@@ -253,22 +253,6 @@ Then copy the resulting built code (ends up in the lib directory) into the vendo
 we just use the asmjs build to avoid wasm complications in
 the build system.
 
-### Experimental VM Modules
```
bucko13, line 256: huge win that this isn't needed anymore! Thanks for the attention to detail and updating the readme 🙏

packages/caravan-psbt/package.json
```diff
@@ -28,11 +28,11 @@
   "scripts": {
     "build": "tsup src/index.ts --format cjs,esm --dts",
     "dev": "npm run build -- --watch",
-    "test": "NODE_OPTIONS=--experimental-vm-modules npx jest src",
-    "lint": "eslint src",
+    "test": "vitest run",
+    "test:watch": "vitest watch",
+    "lint": "eslint --fix src/",
```
bucko13, line 33: split out into `lint` and `lint:fix` here too.

packages/caravan-psbt/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "baseUrl": ".",
     "types": [
       "node",
-      "jest"
+      "vitest/globals"
```
bucko13, line 7: this might not be necessary with the base config.

bucko13 (changes requested):
just a couple of small things

packages/caravan-psbt/README.md
DhairyaMajmudar: Just Awesome ❤️

packages/caravan-psbt/package.json
DhairyaMajmudar: Done 👍🏻

packages/caravan-psbt/tsconfig.json
DhairyaMajmudar: Actually we need this, as adding `vitest/globals` in types we don't need to import vitest functions in each file.

`import {vi, describe, it, expect} from "vitest";`

Above line can be avoided in test files.

bucko13: Right. I get that we need it somewhere, but i would've thought that since it's in the base.json tsconfig file extended in [line two](https://github.com/caravan-bitcoin/caravan/pull/242/files/08ab057c5ad90aed9eef3cc01a2f982c20ee4aa8#diff-bf4243f0bbd72ee9f3432cd7d15aac43fa9663c64b56d113bb72e9f13414e9f1R2) that it wouldn't be needed here as well. If the linter complains without it in the local file then I guess we still need it.

DhairyaMajmudar: Yeah, I tried removing it from here, but unfortunately got so many linter errors. That's a bit strange because in base config it is also defined, then there should not be any such errors here.

PR #243, opened 2025-04-14
Improved Privacy test in caravan health package

vivekmaurya001:
Progresses the TODO in caravan health package
 
- added case of simple spend in topology calculation
- added case of half reused and fully used address in Address Reuse Factor calculation


packages/caravan-health/src/privacy.test.ts
```diff
@@ -3,115 +3,142 @@ import { determineSpendType, getSpendTypeScore } from "./spendType";
...
-    size: 0,
-    weight: 0,
-    fee: 0,
-    isSend: true,
-    amount: 0,
-    block_time: 0,
+    size: 200,
+    weight: 800,
+    fee: 0.0001,
+    isSend: false,
```
bucko13, line 20: if this is a spend shouldn't this be true?

vivekmaurya001: I think , I have commented it wrong , it should not be perfect spend rather it should be perfect transaction

vivekmaurya001:
@bucko13 , Any furthur suggestion on this PR ?

bucko13:
Sorry I haven't had a chance to look again (currently 20 open PRs!) but will try to asap.

packages/caravan-health/src/privacy.test.ts
```diff
@@ -3,115 +3,142 @@ import { determineSpendType, getSpendTypeScore } from "./spendType";
...
-      },
+      { prevTxId: "txid1", vout: 0, sequence: 0 },  // spends txid1's output
+      { prevTxId: "coinbase", vout: 1, sequence: 0 },
     ],
     vout: [
       {
-        scriptPubkeyHex: "scriptPubkeyHex2",
-        scriptPubkeyAddress: "scriptPubkeyAddress2",
+        scriptPubkeyHex: "hex2",
+        scriptPubkeyAddress: "addr2",  // We'll use in UTXOs
```
Legend101Zz, line 35: ---

I noticed some inconsistency in the use of addresses—at some places you’ve used realistic mainnet-style addresses, while in others, placeholder strings. While I understand the intent might be to simulate reuse patterns for transaction linkage analysis, there's a functional concern:

The `getAddressType` utility expects a valid address format to correctly infer the type (e.g., P2WPKH, P2SH, etc.). If placeholder strings don’t conform to actual Bitcoin address formats, the function might fail or behave unpredictably.

To maintain clarity and ensure consistency, I’d recommend defining reusable constants like:

```ts
const ADDR1 = "bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh";
```

…and using them throughout. This also makes your test cases easier to read and maintain.

packages/caravan-health/src/privacy.test.ts
```diff
@@ -3,115 +3,142 @@ import { determineSpendType, getSpendTypeScore } from "./spendType";
...
-        scriptPubkeyAddress:
-          "bc1qng72v5ceptk07htel0wcv6k27fkg6tmmd8887jr2l2yz5a5lnawqqeceya",
+        scriptPubkeyHex: "hex2",
+        scriptPubkeyAddress: "bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh",
         value: 0.2,
       },
       {
-        scriptPubkeyHex: "scriptPubkeyHex2",
-        scriptPubkeyAddress: "scriptPubkeyAddress2",
+        scriptPubkeyHex: "hex2",
```
Legend101Zz, line 49: same thing for the `scriptPubkeyAddress`

Legend101Zz (changes requested):
@vivekmaurya001 Thanks for making the changes — and apologies for the delay in getting an initial review over to you.

I really appreciate the thoughtful inclusion of coinbase transactions. It's a valuable real-world scenario to test how privacy metrics behave when dealing with newly mined coins versus previously transacted UTXOs. Additionally, the way you structured a transaction chain to simulate history and address reuse significantly improves the ability to assess linkage — which is crucial for evaluating transaction privacy. Great work on capturing that complexity in the tests!

Legend101Zz:
also @vivekmaurya001  — if possible, could you add some explanatory comments to the tests you wrote?
It would be really helpful for anyone revisiting the tests later to understand the reasoning behind each case. Thanks!

vivekmaurya001:
@Legend101Zz , Sure I will do that soon

Legend101Zz:
> Hi , @Legend101Zz , I was a little busy from last 2 weeks but now I have done the requested changes , added reusable addresses and explanatory comments

Thanks a lot! I’ll go ahead and drop in the reviews now.

Legend101Zz (approved):
Yup, the changes look solid to me now.
@bucko13, could you take a final look and add any reviews if needed?

packages/caravan-health/src/privacy.test.ts
bucko13: I'm still confused. What's a perfect transaction? `PerfectSpend` is an "official" tx type as far as the library is concerned. See `SpendType` enum in the types.ts file. "Perfect tx" is not in that list. If you're trying to test an incoming tx that has one input and one output, then this doesn't really factor into the spend types, and we should label it something else. 

I'm also concerned with the changing nature of this test fixture. Originally it was testing for a perfect spend, and it is a send. Now we're testing something else. Are we losing a test case now?

packages/caravan-health/src/privacy.test.ts
```diff
@@ -282,40 +315,65 @@ describe("Privacy metric scoring", () => {
 
   describe("Transaction Topology Score", () => {
     it("Calculates the transaction topology score based on the spend type", () => {
+      // not a self payment
       const score: number = privacyMetric.getTopologyScore(transactions[0]);
-      expect(score).toBe(0.75);
+      expect(score).toBe(0.5);
```
bucko13, line 320: see, now we have a different outcome. maybe having both is fine, but we shouldn't be losing a test case imo.

packages/caravan-health/src/privacy.test.ts
```diff
@@ -282,40 +315,65 @@ describe("Privacy metric scoring", () => {
 
   describe("Transaction Topology Score", () => {
     it("Calculates the transaction topology score based on the spend type", () => {
```
bucko13, line 317: might be worth changing this into a vector based test using test.each. then you can name them for each spend/tx type you're testing and consolidate the test code (also makes it easier to add more test cases in the future).

packages/caravan-health/src/privacy.test.ts
```diff
@@ -282,40 +315,65 @@ describe("Privacy metric scoring", () => {
...
+      const score3: number = privacyMetric.getTopologyScore(transactions[2]);
+      expect(score3).toBeCloseTo(0.42);
     });
   });
 
   describe("Mean Topology Score", () => {
     it("Calculates the mean topology score for all transactions done by a wallet", () => {
       const meanScore: number = privacyMetric.getMeanTopologyScore();
-      expect(meanScore).toBeCloseTo(0.583);
+      expect(meanScore).toBeCloseTo(0.445);
```
bucko13, line 333: what accounted for this change?

packages/caravan-health/src/privacy.test.ts
```diff
@@ -282,40 +315,65 @@ describe("Privacy metric scoring", () => {
...
     });
   });
 
   describe("Address Reuse Factor", () => {
```
bucko13, line 337: nice additions.

packages/caravan-health/src/privacy.test.ts
```diff
@@ -282,40 +315,65 @@ describe("Privacy metric scoring", () => {
...
+      };
+      const metricUnique = new PrivacyMetrics(transactions, uniqueUtxos);
+      expect(metricUnique.addressReuseFactor()).toBe(1);
     });
   });
 
   describe("Address Type Factor", () => {
     it.todo("Test with different combination of address types and networks");
+
+    // When wallet type not matches with any of used addresses
```
bucko13, line 367:
```suggestion
    // When wallet type does not match with any used addresses
```

packages/caravan-health/src/privacy.test.ts
```diff
@@ -282,40 +315,65 @@ describe("Privacy metric scoring", () => {
...
     it("Calculates the the address type distribution of the wallet transactions", () => {
       const addressTypeFactor: number = privacyMetric.addressTypeFactor(
         "P2SH",
         Network.MAINNET,
       );
       expect(addressTypeFactor).toBe(1);
     });
+
+    // When some used address matches with wallet type (Here it is ADDR1)
```
bucko13, line 376:
```suggestion
    // When some used address match with wallet type (Here it is ADDR1)
```

packages/caravan-health/src/privacy.test.ts
```diff
@@ -328,15 +386,15 @@ describe("Privacy metric scoring", () => {
   describe("UTXO Spread Factor", () => {
     it("Calculates the standard deviation of UTXO values which helps in assessing the dispersion of UTXO values", () => {
       const utxoSpreadFactor: number = privacyMetric.utxoSpreadFactor();
-      expect(utxoSpreadFactor).toBeCloseTo(0.1);
+      expect(utxoSpreadFactor).toBeCloseTo(0.07);
```
bucko13, line 389: same here and below. curious what accounted for the change in scoring.

bucko13 (changes requested):
Still not quite sure about some things but getting closer. Thanks!

Legend101Zz:
Hi @vivekmaurya001 just wanted to check in to see if you had the chance to look into this :)

Legend101Zz:
> Hi @Legend101Zz , Sorry for late responding , I have been really busy these weekends , will surely look into this as I got time .

great thanks :)
Just pleaase make sure you update your branch with the latest changes in the `main`

Issue #181, opened 2025-02-24
Selecting bitcoin client before importing blocks user from importing a wallet


jevidon:
If I select a "Bitcoin client" on the wallet page first and then attempt import a config, it hides the block explorer selection during import and I am unable to click the confirm button.

![Image](https://github.com/user-attachments/assets/13adb751-f5e3-4a3f-ba44-edcb89befc74)

parlelit:
@jevidon Thank you for reaching out! For further assistance, please feel free to send us an email at  instantresolution247@aol.com . We're here to help!

khadar1020:
I would like to work on this issue

khadar1020:
Hi, @jevidon I have used testnet address and then imported the config then it is working fine for me can you please check it and tell me

shashankxrm:
Even I didn't get any problem like that @jevidon . Could u screenrecord or something to clearly show the bug?

> Hi, [@jevidon](https://github.com/jevidon) I have used testnet address and then imported the config then it is working fine for me can you please check it and tell me

amitx13:
> Even I didn't get any problem like that [@jevidon](https://github.com/jevidon) . Could u screenrecord or something to clearly show the bug?

There is a bug. Follow this flow:
1. Directly import the wallet config and observe the `WalletGenerator` component. But before importing the wallet config, make sure there is no client info in the config otherwise, you won’t be able to reproduce the issue.
2. This time, select the `BitcoinClient` first, and then import the same wallet config. You’ll notice that it hides the block explorer selection.

I dug into it and found that the Redux state is correctly updated to "unknown", but the useSelector hook inside the useGetClient component doesn’t trigger a re-evaluation with the new state. This prevents the useEffect from running and updating the blockchain client.

It appears to be a component lifecycle/rendering issue. I'm working on a fix and will submit a PR shortly for review.

PR #244, opened 2025-04-16
Separate concepts of public/private and explorer

waldenraines:
Rather than combining  the concepts of public and private bitcoin clients with mempool/blockstream, make PublicBitcoin Client it's own enum.

Introduces a new format for bitcoind client, formatted thusly:

```
  "client":  {
    "type": "public",
    "provider": "mempool"
  },
```

Provides backwards compatibility to configurations that predate this change.

1. No client:

image

Selector is presented :heavy_check_mark: 

2. client.type == 'public':

i.e.
```
  "client":  {
    "type": "public"
  },
```
Defaults to blockstream :heavy_check_mark: 

image

3. client.type == 'mempool' | 'blockstream':

i.e.

```
  "client":  {
    "type": "mempool"
  },
```

Honors the selection :heavy_check_mark: 

4. Private client:
image

6. New style:

```
  "client":  {
    "type": "public",
    "provider": "mempool"
  },
```

Works as expected :heavy_check_mark: 

Also fixes https://github.com/caravan-bitcoin/caravan/issues/181.


Legend101Zz:
Hi @waldenraines — thanks for the PR! Really cool concept. I spent some time recently on the `caravan/clients` package, so I’ll dive into your changes and share some review feedback tomorrow. Looking forward to it!

packages/caravan-clients/src/client.ts
```diff
@@ -497,29 +509,33 @@ export class BlockchainClient extends ClientBase {
...
-          } else if (blocks <= 3) {
-            return fees.halfHourFee;
-          } else if (blocks <= 6) {
-            return fees.hourFee;
-          } else {
-            return fees.economyFee;
-          }
-        default:
-          throw new Error("Invalid client type");
+      if (this.type === ClientType.PRIVATE) {
```
Legend101Zz, line 512: I'd suggest sticking with the switch statement — it keeps things clean and easy to follow :)

packages/caravan-clients/src/client.ts
```diff
@@ -240,33 +247,38 @@ export class BlockchainClient extends ClientBase {
...
-    if (type === ClientType.BLOCKSTREAM) {
-      host = "https://blockstream.info";
-    } else if (type === ClientType.MEMPOOL) {
-      host = "https://unchained.mempool.space";
-    }
-    if (type !== ClientType.PRIVATE && network !== Network.MAINNET) {
-      host += `/${network}`;
-    }
-    if (type !== ClientType.PRIVATE) {
+    if (type === ClientType.PUBLIC) {
```
Legend101Zz, line 263: Would it make sense to shift this check at the very start? because [here](https://github.com/caravan-bitcoin/caravan/pull/244/files#diff-4fd3f89188bb2c009b1c8451e8ccd94db7f894cbc6456d4e1c73a83327b42287R257) too we check for provider so just to make sure that it's not undefined

packages/caravan-clients/src/client.ts
```diff
@@ -217,18 +221,21 @@ export interface BitcoindParams {
...
 
 export class BlockchainClient extends ClientBase {
   public readonly type: ClientType;
+  public readonly provider?: PublicBitcoinProvider;
   public readonly network?: Network;
   public readonly bitcoindParams: BitcoindParams;
 
   constructor({
     type,
+    provider,
```
Legend101Zz, line 238: What happens in the case ( hypothetically) we have a private client which has a provider field , should we add a check for that too ??

packages/caravan-clients/src/client.test.ts
```diff
@@ -1555,7 +1584,8 @@ describe("BlockchainClient", () => {
       );
 
       const mempoolClient = new BlockchainClient({
-        type: ClientType.MEMPOOL,
+        type: ClientType.PUBLIC,
```
Legend101Zz, line 1587: tests looks great to me :)

apps/coordinator/src/actions/clientActions.ts
```diff
@@ -37,29 +44,62 @@ const matchesClient = (
   client: ClientSettings,
   network: BitcoinNetwork,
 ) => {
+  const translatedType = getClientType(client);
+  const translatedProvider = getClientProvider(client);
+  console.log("Matching client:", {
```
Legend101Zz, line 49: I guess we can remove this debug statement?

apps/coordinator/src/actions/clientActions.ts
```diff
@@ -37,29 +44,62 @@ const matchesClient = (
...
       ? blockchainClient.bitcoindParams.url === client.url &&
         blockchainClient.bitcoindParams.auth.username === client.username &&
         blockchainClient.bitcoindParams.auth.password === client.password
-      : true)
+      : blockchainClient.provider === translatedProvider)
   );
 };
 
 const getClientType = (client: ClientSettings): ClientType => {
+  console.log("Getting client type for:", client.type);
```
Legend101Zz, line 72: this one too

apps/coordinator/src/actions/clientActions.ts
```diff
@@ -37,29 +44,62 @@ const matchesClient = (
...
+      return ClientType.PUBLIC;
     case "private":
       return ClientType.PRIVATE;
     default:
       return client.type as ClientType;
   }
 };
 
+export const getClientProvider = (client: ClientSettings) => {
+  console.log("Getting provider for:", client);
```
Legend101Zz, line 86: here too

apps/coordinator/src/components/ClientPicker/index.jsx
```diff
@@ -123,34 +137,40 @@ const ClientPicker = ({
           <FormControl component="fieldset">
             <RadioGroup>
               <FormControlLabel
-                id={ClientType.MEMPOOL}
+                id="mempool"
```
Legend101Zz, line 140: can we use `PublicBitcoinProvider.Mempool` here and in the rest of the form lables too ??

packages/caravan-clients/src/client.ts
```diff
@@ -41,8 +41,12 @@ export class BlockchainClientError extends Error {
 
 export enum ClientType {
   PRIVATE = "private",
+  PUBLIC = "public"
```
Legend101Zz, line 44: ---

Just a curious question — was there a specific reason you didn’t define `PUBLIC` as a type like `PublicBitcoinProvider`? That way, in many places we could use something like `ClientType.PUBLIC.MEMPOOL`, which might improve readability and better reflect the enum hierarchy.

Legend101Zz (changes requested):
---

@waldenraines Thanks for the thoughtful suggestions and updates — this really improves the readability and scalability of the `clients` package. I’ve added a few review comments as well.

Also, just a quick note: I think we can safely remove some of the `console` statements that are scattered across the code.

One small question I had — was there a specific reason you chose not to define `PUBLIC` as a type like `PublicBitcoinProvider`? Just wondering, since that structure might help reinforce the enum hierarchy and improve clarity in usage.

Legend101Zz:
---

Hey @waldenraines — just looping you in. Whenever you get a chance, could you take a look at the review comments? Would love to get this merged once everything looks good :)

apps/coordinator/src/actions/clientActions.ts
waldenraines: Yes, removed

apps/coordinator/src/actions/clientActions.ts
waldenraines: Removed

apps/coordinator/src/actions/clientActions.ts
waldenraines: Removed

packages/caravan-clients/src/client.ts
waldenraines: Done

packages/caravan-clients/src/client.ts
waldenraines: Addressed by throwing an error if `type === ClientType.PRIVATE && provider`

apps/coordinator/src/components/ClientPicker/index.jsx
waldenraines: Addressed by changing id and value props

packages/caravan-clients/src/client.ts
waldenraines: It was just an attempt to minimize the number of changes.  It would make more sense for it to be a type though.

waldenraines:
Okay I believe I have addressed everything but please double check me @Legend101Zz

packages/caravan-clients/src/client.ts
```diff
@@ -240,33 +249,57 @@ export class BlockchainClient extends ClientBase {
...
+      // eslint-disable-next-line no-param-reassign
+      type = ClientType.PUBLIC;
     }
-    if (type !== ClientType.PRIVATE && network !== Network.MAINNET) {
-      host += `/${network}`;
+
+    if (type === ClientType.PUBLIC && !provider) {
+        // Default to blockstream if no provider is specified for a public client
+        // eslint-disable-next-line no-param-reassign
+        provider = PublicBitcoinProvider.BLOCKSTREAM;
```
bucko13, line 280: why not mempool now?

packages/caravan-clients/src/client.ts
```diff
@@ -240,33 +249,57 @@ export class BlockchainClient extends ClientBase {
...
     }
-    if (type !== ClientType.PRIVATE) {
-      host += "/api";
+
+    let hostURL = "";
+    if (type === ClientType.PUBLIC) {
+      if (!provider) {
```
bucko13, line 285: is this possible? above you check if it's public and if provider is not set, you set it. So this check shouldn't be necessary.

packages/caravan-clients/src/client.test.ts
```diff
@@ -93,64 +94,112 @@ describe("BlockchainClient", () => {
...
       new BlockchainClient({
-        type: ClientType.BLOCKSTREAM,
+        type: ClientType.PUBLIC,
+        provider: PublicBitcoinProvider.BLOCKSTREAM,
         network: Network.SIGNET,
       });
     }).toThrow("Invalid network");
   });
 
+  it("should default to blockstream for public client if no provider is specified", () => {
```
bucko13, line 164: maybe default to mempool (where we don't get throttled)

apps/coordinator/src/reducers/clientReducer.js
```diff
@@ -13,6 +14,7 @@ import {
 
 const initialState = {
   type: "public",
+  provider: "blockstream",
```
bucko13, line 17: mempool

apps/coordinator/src/actions/clientActions.ts
```diff
@@ -37,29 +44,50 @@ const matchesClient = (
...
+    client.type === "mempool" ||
+    client.type === "blockstream"
+  ) {
+    // For public clients, we need to determine which provider to use
+    if (client.provider === "blockstream" || client.type === "blockstream") {
+      return PublicBitcoinProvider.BLOCKSTREAM;
+    } else if (client.provider === "mempool" || client.type === "mempool") {
+      return PublicBitcoinProvider.MEMPOOL;
+    }
+    return PublicBitcoinProvider.BLOCKSTREAM; // Default to Blockstream if not specified
```
bucko13, line 86: another default to mempool

bucko13 (changes requested):
Looks good overall! A couple small things. One aesthetic, the other is on the default public client which I think should be mempool

packages/caravan-clients/src/client.ts
Legend101Zz: understood ...

packages/caravan-clients/src/client.ts
waldenraines: Updated.

Legend101Zz:
> > @waldenraines thanks for the review fixes ... sorry I was not able to throw in a second round of reviews ... Just checking in to see if you had a chance to look into the reviews that @bucko13 wrote ?
> 
> @bucko13 @Legend101Zz finally getting back around to this, I think I addressed everything, let me know if not!

Thank you @waldenraines .... Everything looks good to me  :)
@bucko13 any further suggestions ?

jevidon:
Just had a chance to test this PR. I'm noticing that when the bitcoin client has been defined in the wallet config being loaded in, that section no longer appears before clicking confirm.

Here's what it looks like without the client defined in the config file: 
image

And here is what it looks like with the client defined in the config file:
image

Despite already having it defined, I would like to validate on this page what the selection is. Almost everything else in the JSON is shown on screen for final validation before proceeding (except for if it's mainnet or testnet, but displaying that could be considered out of scope for this change).

waldenraines:
> Just had a chance to test this PR. I'm noticing that when the bitcoin client has been defined in the wallet config being loaded in, that section no longer appears before clicking confirm.

This is the same behavior I am seeing in the released version of caravan, can we handle that outside of this MR as an enhancement?

PR #247, opened 2025-04-17
Migrates `@caravan/caravan-client` tests to vitest

DhairyaMajmudar:
## Description

This PR migrates the `@caravan/caravan-client` test cases from jest to vitest

### DIFF
--- a/packages/caravan-clients/src/bitcoind.ts
+++ b/packages/caravan-clients/src/bitcoind.ts
@@ -2,9 +2,10 @@
 /*
 TODO: cleanup the no explicit any. added to quickly type error catches
 */
-import axios, { AxiosBasicCredentials } from "axios";
-import BigNumber from "bignumber.js";
 import { bitcoinsToSatoshis } from "@caravan/bitcoin";
+import axios, { AxiosBasicCredentials } from "axios";
+import { BigNumber } from "bignumber.js";
+
 import {
   RPCResponse,
   ListUnspentResponse,
--- a/packages/caravan-clients/src/blockExplorer.ts
+++ b/packages/caravan-clients/src/blockExplorer.ts
@@ -2,13 +2,14 @@
 /*
 TODO: cleanup the no explicit any. added to quickly type error catches
 */
-import axios from "axios";
-import BigNumber from "bignumber.js";
 import {
   satoshisToBitcoins,
   blockExplorerAPIURL,
   Network,
 } from "@caravan/bitcoin";
+import axios from "axios";
+import { BigNumber } from "bignumber.js";
+
 import {
   BlockExplorerUTXOResponse,
   BlockExplorerAddressResponse,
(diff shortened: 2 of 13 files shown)


packages/caravan-clients/src/bitcoind.ts
```diff
@@ -2,9 +2,10 @@
 /*
 TODO: cleanup the no explicit any. added to quickly type error catches
 */
-import axios, { AxiosBasicCredentials } from "axios";
-import BigNumber from "bignumber.js";
 import { bitcoinsToSatoshis } from "@caravan/bitcoin";
+import axios, { AxiosBasicCredentials } from "axios";
```
Legend101Zz, line 6: I guess this change was accidental?

packages/caravan-clients/src/blockExplorer.ts
```diff
@@ -2,13 +2,14 @@
...
 TODO: cleanup the no explicit any. added to quickly type error catches
 */
-import axios from "axios";
-import BigNumber from "bignumber.js";
 import {
   satoshisToBitcoins,
   blockExplorerAPIURL,
   Network,
 } from "@caravan/bitcoin";
+import axios from "axios";
```
Legend101Zz, line 10: same here ... or are you trying to follow some convention ?

packages/caravan-clients/src/blockchain.ts
```diff
@@ -3,22 +3,22 @@
...
   bitcoindParams,
   isWalletAddressNotFoundError,
 } from "./bitcoind";
-
+import {
+  blockExplorerGetAddresesUTXOs,
+  blockExplorerGetFeeEstimate,
+  blockExplorerBroadcastTransaction,
+  blockExplorerGetAddressStatus,
+} from "./blockExplorer";
```
Legend101Zz, line 19: same here

packages/caravan-clients/src/client.test.ts
```diff
@@ -405,9 +408,9 @@ describe("BlockchainClient", () => {
           time: "string",
         },
       ];
-      const mockBitcoindListUnspent = jest
+      const mockBitcoindListUnspent = vi
         .spyOn(wallet, "bitcoindListUnspent")
-        .mockResolvedValue(Promise.resolve(mockUnspent));
+        .mockResolvedValue(mockUnspent);
```
Legend101Zz, line 413: nice ... prevents the nested Promises :)

packages/caravan-clients/src/bitcoind.ts
bucko13: I think this is reordering from one of the newer eslint rules.

DhairyaMajmudar: Yeah true, the new import order is from the new eslint config file.

packages/caravan-clients/src/client.test.ts
DhairyaMajmudar: Thank you : )

bucko13:
It looks like you didn't need the tsconfig for the vitest globals locally for the package in this one, which is more along the lines of what i expected.

PR #251, opened 2025-04-17
Improve CI Pipeline to Run All Tests Across Packages

abbi4code:
- This PR makes sure our CI pipeline runs all tests for every package in the monorepo.
- This should help catch issues across all packages before merging. 

using workflow_dispatch for testing in my own repo, will  remove this in the next commit after testing it

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  workflow_dispatch: 
 
 jobs:
   ci:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: Test
 
-on: [push, pull_request]
+on: 
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
 
 jobs:
   ci:
@@ -26,6 +30,8 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
-      - name: Run tests
+      - name: Run CI Tasks
         run: |
           npx turbo run ci
+        env:
+          NODE_ENV: test
--- a/turbo.json
+++ b/turbo.json
@@ -52,8 +52,9 @@
     },
     "ci": {
       "dependsOn": [
-        "^build"
-      ]
+        "^build", "build","test","lint"
+      ],
+      "outputs": []
     }
   }
 }


turbo.json
```diff
@@ -52,8 +52,9 @@
     },
     "ci": {
       "dependsOn": [
-        "^build"
-      ]
+        "^build", "build","test","lint"
```
bucko13, line 55: currently there are `ci` scripts that run these already. This would cause duplicate runs in those cases. I can see the appeal of setting it up this way though.

bucko13: yeah, exactly. I think it could make sense to do this way, but it would have to be coupled with a cleanup of existing scripts.

abbi4code: i was thinking we could make our CI stronger by adding type-checking for ts errors, coverage reporting with codecov, checks for sensitive files and file change limits, and prettier formatting checks , nothing too complex, can share a detailed plan on Discord if you’d like.

bucko13: yep! that would be the idea. the individual scripts I think do call all of those, but I think I agree that doing it via the turbo config might be more robust.

abbi4code:
@bucko13 
I've made some improvements to the code coverage and CI setup across the Caravan monorepo. Here's what’s been implemented so far:

- new pull_request.yml workflow: added a dedicated workflow (``.github/workflows/pull_request.yml``) that runs on all PRs (any branch).
- updated ``ci.yml``: added coverage checks (``npm run test:coverage``) and the Python script to ``.github/workflows/ci.yml`` for main branch pushes and PRs, ensuring consistency.
- added lcov-result-merger(for now) to root package.json for merging coverage reports.

**pending tasks still need to work on**

Root ``vitest.config.ts``: currently, we’re using package-level ``vitest.config.ts``  files, which output separate coverage reports per package. I’m working on adding a root ``vitest.config.ts`` to:

- unify coverage into a single ``./coverage`` directory for easier Codecov integration (no need for lcov-result-merger).

- once everything done, workflows are set to upload ./coverage/lcov.info to Codecov, but we need the ``CODECOV_TOKEN`` secret added to the repo settings.

.github/workflows/ci.yml
```diff
@@ -26,6 +30,29 @@ jobs:
...
+      #     files: ./coverage/lcov.info
+      #     fail_ci_if_error: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Check Coverage Disable
+        run: python .github/workflows/scripts/code_coverage_disable_check.py --directory apps packages
```
bucko13, line 58: new line

vitest.config.ts
```diff
@@ -0,0 +1,29 @@
...
+    coverage: {
+      enabled: true,
+      provider: 'istanbul',
+      reportsDirectory: './coverage',
+      reporter: ['text', 'html', 'lcov'],
+      exclude: [
+        'node_modules',
+        'dist',
+        '**/*.test.{ts,tsx}',
+        'coverage/**',
```
bucko13, line 16: are we committing coverage reports? should probably be gitignored right?

vitest.config.ts
```diff
@@ -0,0 +1,29 @@
...
+        'apps/*/build/**',
+        'packages/*/dist/**'
+      ]
+    },
+    include: [
+      'apps/**/*.{ts,tsx}',
+      'packages/**/*.{ts,tsx}'
+    ]
+  }
+});
```
bucko13, line 29: new line

vitest.config.ts
```diff
@@ -0,0 +1,29 @@
...
+        'vitest.config.ts',
+        'vitest.setup.ts',
+        'apps/*/build/**',
+        'packages/*/dist/**'
+      ]
+    },
+    include: [
+      'apps/**/*.{ts,tsx}',
+      'packages/**/*.{ts,tsx}'
+    ]
```
bucko13, line 27: why was the exclude and include needed?

.github/workflows/scripts/code_coverage_disable_check.py
```diff
@@ -0,0 +1,33 @@
+import os
+import re
+import sys
+
+def has_code_coverage_disable(file_path):
```
bucko13, line 5:
```suggestion
def has_code_coverage_disabled(file_path):
```

.github/workflows/scripts/code_coverage_disable_check.py
bucko13, line 1: Why is this file necessary? Can we just use istanbul ignore? https://github.com/istanbuljs/nyc?tab=readme-ov-file#parsing-hints-ignoring-lines

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,51 @@
...
+      #     files: ./coverage/lcov.info
+      #     fail_ci_if_error: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Check Coverage Disable
+        run: python .github/workflows/scripts/code_coverage_disable_check.py --directory apps packages
```
bucko13, line 51: new line.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,51 @@
...
+      # - name: Merge Coverage Reports
+      #   run: npx lcov-result-merger 'apps/*/coverage/lcov.info packages/*/coverage/lcov.info' './coverage/lcov.info'
+
+
+      # - name: Upload to Codecov
+      #   uses: codecov/codecov-action@v4
+      #   with:
+      #     token: ${{ secrets.CODECOV_TOKEN }}
+      #     files: ./coverage/lcov.info
+      #     fail_ci_if_error: true
```
bucko13, line 43: let's just remove for now. If we wan to upload coverage reports we can set that up later.

bucko13 (changes requested):
Left a few questions and comments

.github/workflows/scripts/code_coverage_disable_check.py
abbi4code: this wasn't necessary, instead the script serves the opposite purpose of istanbul ignore comments.
- Istanbul ignore = lets developers skip testing certain code areas
- This script = prevents using istanbul ignore, enforcing 100% coverage with no exceptions

abbi4code: Do you prefer strict 100% coverage then we can keep this script or flexible coverage where developers can skip testing edge cases when justified then we can remove this, but I suggest we can keep the script but remove it from workflows for now, we can re-enable it in the future
I'll remove it from the workflows for now, but we can always bring it back later if needed.

vitest.config.ts
abbi4code: thanks for pointing out, I initially thought the include/exclude patterns were necessary, but after testing I found that removing them doesn't break anything, as turbo, runs vitest run individually in each package rather than using the root config.

abbi4code: I've remove the include/exclude patterns since they're not actually needed for our turbo-based workflow.

Legend101Zz:
> This pull request has been inactive for 30 days and has been marked as stale. It will be closed in 7 days if no further activity occurs. To keep this PR open, add the "long-lived" label or comment on it.

added the `long-lived` tag so abhishek can get back to this once done with the e2e part :)

PR #259, opened 2025-04-23
fix(build): update build deps, cleanup dep tree, and update polyfills

bucko13:
in #239 we added a fix to get @caravan/psbt working in node environments with a custom plugin for the tsup build.  In this PR, I've moved that to its own internal package that the other packages can also use in their build. 

In working through this, it turns out that we were 1 major version behind of the latest tsup and several behind esbuild since I was finding the plugin built ontop of the latest esbuild was conflicting with tsup. So in getting everything up to date, I also decided to move these build deps to the root package.json to keep everything in sync more easily and actually update to more recent versions. This results in a pretty large package-lock change unfortunately, but not that much actually changed (thanks npm). 

In updating tsup i also found that the node polyfill plugin we were using is no longer being supported and was incompatible with the latest esbuild/tsup. So I moved that to a new package that is better supported (also now declared in the root), but had a slightly different api.


packages/build-plugins/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  var self;
+  if (typeof window !== 'undefined') {
+    self = window;
+  } else if (typeof globalThis !== 'undefined') {
+    self = globalThis;
+  } else {
+    self = {}
+  }
+}
+`;
```
Shadouts, line 14: So, it seems that this code gets prepended to any JS file. I think I understand the need for it (cross compatibility for browser vs node), but tracking evaluated code-as-a-string seems dangerous.

I seem to remember long ago having to use a node package which provides fake versions of these browser globals. Maybe I was using it for a Next project. Maybe this is just me handwaving an "Is there a better way?".

Shadouts (approved):
Also, maybe a readme for `build-plugins` would be helpful or at least a description in its `package.json`.

I ran many package.json scripts to see if everything worked on my end. It all seems good, but something which is probably unrelated to this change is I'm seeing issues running `npm run dev`. Turbo is requesting larger concurrency, and even after adjusting it, it doesn't complete its build/launch process.

packages/build-plugins/src/index.ts
bucko13: Yeah, @joelangeway even says in his original implementation of this that it's "admittedly hacky". I believe he pulled this from what was being inserted in the build artifacts in the version before it was broken in node use cases. 

All that to say, I agree, but am not sure if there's a better way. Definitely something to keep an eye on.

bucko13:
> Also, maybe a readme for build-plugins would be helpful or at least a description in its package.json.

Good call! added in [18d3478](https://github.com/caravan-bitcoin/caravan/pull/259/commits/18d3478bbff81014419fceac6940ddd781a75333)

Thanks for aggressively testing as well!

bucko13:
merging in assuming the previous approval still applies with the package.json description now added

PR #261, opened 2025-04-23
fix: moved the policy file from caravan-wallets to  caravan/multisig

### LINKED ISSUE #260: Move policy.ts from @caravan/wallets to @caravan/multisig and make public
the policy module is useful in more contexts than just the hardware wallets one it was first built for. the @caravan/multisig package provides similar utilities especially around multisig wallet config and management. Moving this there would make it more useful in other contexts that might need to manipulate and validate wallet policies and key origins.

krrish-sehgal:
1. made this as a minor change
2. listed caravan/multisig as a dependency in caravan-wallets/package.json
3. moved policy.ts, policy.test.ts and fixtures to caravan/multisig
4. also removed the braidDetailsToWalletConfig function as its ambigous with the one in the multisig/utils file and is more generic to be put in utils file.

fixes #260


krrish-sehgal:
I am currently trying to fix this test/ci errors , the error is particularly in the psbt test cases , because it was somehow dependent on the policy.ts files , and as soon as i export the files from multisig package , the test suite starts to fail : 

Screenshot 2025-04-23 at 5 15 10 PM

.changeset/sharp-bats-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@caravan/wallets": minor
```
bucko13, line 2: Hopefully should be no change in functionality at all to @caravan/wallets. Just a new build dependency which keeps everything the same. Could maybe even skip the changeset, but at most it's a patch.

packages/multisig/src/policy.ts
```diff
@@ -251,21 +252,3 @@ export const validateMultisigPolicyTemplate = (template) => {
   validateMultisigPolicyScriptType(template);
   validateMultisigPolicyKeys(template);
 };
-
-// Mostly useful for dealing with test fixtures and objects from caravan
-export const braidDetailsToWalletConfig = (braidDetails: BraidDetails) => {
```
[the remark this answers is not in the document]
krrish-sehgal: @bucko sir I removed the braidDetailsToWalletConfig function as its ambigous with the one in the multisig/utils file and is more generic to be put in utils file, let me know if we want to remove it from the utils file and shift the utils.test.ts into policy.test.ts as well.

.changeset/sharp-bats-confess.md
krrish-sehgal: sure , I'll skip the changeset

bucko13:
I'm looking at some of the failures and I think they might get fixed with #261. So hopefully once that's merged in and you bring those changes in, it should start working.

bucko13:
that was tricky to solve with that weird custom error! I think I solved it with [042fe90](https://github.com/caravan-bitcoin/caravan/pull/261/commits/042fe90e466b3541f956bdeb4cf0921d35860007) though.

.changeset/sharp-bats-confess.md
bucko13: nevermind. this is actually a `major` change to caravan-wallets because it removes previously existing functionality. So we can either leave it as being exported from caravan/wallets, or we bump the version to major here.

bucko13: case in point, coordinator build was breaking after the other pieces were fixed. i patched those in [a5bf4df](https://github.com/caravan-bitcoin/caravan/pull/261/commits/a5bf4df8cb4c2fec673f53b1b0547d692801e001)

krrish-sehgal:
@bucko sir , this looks great! , all build are successful , also i have removed the changeset... since we will be facing some first  major changes to bitcoin, wallets, fees and multisig packages , let me know if we want the version to be changed to major here.

bucko13:
>  let me know if we want the version to be changed to major here.

How about instead we just export the policy modules that are imported from @caravan/multisig that way nothing breaks from caravan/wallets and it can stay in sync with anything that might change in caravan/multisig

bucko13 (approved):

bucko13 (approved):

krrish-sehgal:
I will be actively working on this, and I'll try to find a better fix as soon as I can.

bucko13:
> I will be actively working on this, and I'll try to find a better fix as soon as I can.

That would be great if you can! I think there are some deeper issues involving how the dependencies are built and then included in the final build for the caravan coordinator. For example, when moving caravan-bitcoin and ledger-bitcoin deps to devDeps I got everything working except I think tests started failing. Other solutions would get dev server working but not production build. 

Long term the solution I think is to fix how packages are built and to make browser versions and node versions and then cleanup the vite config a little more for caravan-coordinator to handle polyfills. 

The simpler solution might just be to only move a smaller subset of code and types from policy.ts to multisig. But I might have to think more about what we want there anyway. So it might not be worth spending too much time on that, unless you had an interest in untangling the builds

krrish-sehgal:
> > I will be actively working on this, and I'll try to find a better fix as soon as I can.
> 
> That would be great if you can! I think there are some deeper issues involving how the dependencies are built and then included in the final build for the caravan coordinator. For example, when moving caravan-bitcoin and ledger-bitcoin deps to devDeps I got everything working except I think tests started failing. Other solutions would get dev server working but not production build.
> 
> Long term the solution I think is to fix how packages are built and to make browser versions and node versions and then cleanup the vite config a little more for caravan-coordinator to handle polyfills.
> 
> The simpler solution might just be to only move a smaller subset of code and types from policy.ts to multisig. But I might have to think more about what we want there anyway. So it might not be worth spending too much time on that, unless you had an interest in untangling the builds

Sure sir, makes sense, let me know and I'll try and helpout wherever I can.

PR #266, opened 2025-04-24
Add test coverage reporter to vitest pacakges 

DhairyaMajmudar:
## Description

1. Add coverage reporter to all the vitest migrated packages 
2. Migrate `@caravan/mutlisig` package tests into vitest.
3. Fixes test failures in `@caravan/bip32`


packages/bip32/package.json
```diff
@@ -28,7 +28,7 @@
     "dev": "npm run build -- --watch",
     "lint": "eslint src/",
     "ci": "npm run lint && npm run test",
-    "test": "vitest run",
+    "test": "vitest run --coverage",
```
bucko13, line 31: maybe we make this a separate test command. `test:coverage`?

DhairyaMajmudar: I added this into the existing command, Because I thought getting a test report with the CI checks would be helpful. It results like the below image.

https://github.com/caravan-bitcoin/caravan/actions/runs/14642942889/job/41089595669?pr=266

image

bucko13: the ci can get pretty busy and if you have failing tests that's usually what you want to focus on. CI should be able to run this as a separate task though and then we can use that as a different validation on PRs. I.e. tests passing vs coverage change (I'd love to have an explicit coverage change report on the PR itself for example that is separate from the tests passing).

DhairyaMajmudar: Valid point, let me make a seprate command for that : )

DhairyaMajmudar: Done, we are now good to go.

packages/caravan-bitcoin/src/fees.ts
```diff
@@ -3,8 +3,7 @@
  * transaction fees.
  */
 
-import BigNumber from "bignumber.js";
-
+import { BigNumber } from "bignumber.js";
```
bucko13, line 6: whoa, this is a surprising change.

bucko13 (approved):

PR #275, opened 2025-04-25
Migrate `@caravan/caravan-wallets` tests to vitest

DhairyaMajmudar:
## Description

This PR migrates the `@caravan/caravan-wallet` test cases from jest to vitest


packages/caravan-wallets/src/ledger.test.ts
```diff
@@ -539,6 +554,20 @@ describe("ledger", () => {
...
     it("signs psbt", async () => {
       const interaction = interactionBuilder();
+
+      Reflect.defineProperty(interaction, "SIGNATURES", {
+        get: () => [fixture.signature[0]],
+      });
+
+      vi.spyOn(interaction, "signPsbt").mockResolvedValue(expectedSigs);
+
+      mockApp.signPsbt(
```
bucko13, line 564: isn't there already a mcok for signPsbt [here](https://github.com/caravan-bitcoin/caravan/pull/275/files#diff-adcb551f711130223f915ce1ef4bfdf44c725c17854fe30cf47dea3937f5a775R350)? Curious why the tests were working before without these extra mocks.

packages/caravan-wallets/src/trezor.ts
```diff
@@ -51,11 +51,6 @@ import {
 } from "./interaction";
 import { MULTISIG_ROOT } from "./index";
 
-/**
- * Constant defining Trezor interactions.
- */
-export const TREZOR = "trezor";
```
bucko13, line 57: not clear to me why this was necessary when the other interaction files do have this constant export.

packages/caravan-wallets/vitest.setup.ts
bucko13, line 1: I'm curious why this wasn't needed before. In any case, it looks reasonable to me. Could you add some extra comments to this file to explain what it's for and why we needed to mock these two in particular.

bucko13 (commented):
Mostly looks really good to me! Just a few questions for clarity and possibly a few small cleanups

packages/caravan-wallets/src/index.ts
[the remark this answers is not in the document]
DhairyaMajmudar: The constant was unused in that particular file (`trezor.ts`)  due that corresponding test cases were failing. I moved the constant here in the `index.ts` file where it is used originally.

packages/caravan-wallets/src/trezor.ts
DhairyaMajmudar: In this file this constant was not used anywhere due to which the tests `trezor.test.ts` were failing. I moved this constant to the `index.ts` file of the package where it is used in some places.

packages/caravan-wallets/package.json
```diff
@@ -45,37 +45,31 @@
...
-    "@types/jest": "^29.5.12",
     "@types/node": "^18.13.0",
     "@types/w3c-web-usb": "^1.0.6",
     "@typescript-eslint/parser": "^5.51.0",
-    "babel-jest": "^29.7.0",
     "babel-plugin-transform-inline-environment-variables": "^0.4.3",
     "eslint": "^8.34.0",
     "eslint-plugin-import": "^2.27.5",
-    "eslint-plugin-jest": "^23.0.3",
+    "happy-dom": "^17.4.6",
```
DhairyaMajmudar, line 54: I've added this dependency for vitest `happy-dom` environment in vitest config file

Reference: https://vitest.dev/guide/environment.html#test-environment

packages/caravan-wallets/src/ledger.test.ts
DhairyaMajmudar: Done removed extra mocks

package-lock.json
```diff
@@ -15389,6 +15354,125 @@
         "@pkgjs/parseargs": "^0.11.0"
       }
     },
+    "node_modules/jake": {
```
bucko13, line 15357: noting that this is needed for ts-jest for some reason. hopefully we can remove that soon after cleaning up the coordinator.

packages/caravan-wallets/vitest.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: "happy-dom",
+    alias: {
+      "bitbox-api": "vitest",
```
bucko13, line 8: interesting. this is just to mock it in during tests basically?

bucko13: wouldn't this undermine proper mocking during the tests? or can we still confirm the right methods are called?

bucko13 (commented):
One question/concern about the alias added to the config. otherwise this looks good.

bucko13:
closing in favor of #279

PR #279, opened 2025-05-03
migrated `caravan-wallets` to vitest

narasimha-1511:
The Major Problem is 

1. before we added jsdom polyfills with node environment (implemented the same)

2. There is a circular dependency between `Trezor.ts` and `index.ts` , index.ts uses the TREZOR constant , from `trezor.ts`
    and `TREZOR.ts` uses a constant from `index.ts` , where they both are not getting initialized 

3. still the coordinator is using jest , which needs `ts-jest` package to run the ci withour errors


apps/coordinator/package.json
```diff
@@ -61,6 +61,7 @@
     "jest-junit": "^11.0.1",
     "prettier": "^3.2.5",
     "standard-version": "^9.0.0",
+    "ts-jest": "^29.3.2",
```
bucko13: wonder if we could keep the version teh same as the removed one so that the package-lock change is not as different.

bucko13 (changes requested):
one nit that I pushed up a change for 🤞

bucko13 (approved):

PR #299, opened 2025-06-02
feat: Xpub export introduction through BCUR2 in Test suite

krrish-sehgal:
Includes  a decoder integrated with test suite for xpub export in testsuite.
Screenshot 2025-06-10 at 12 51 21 PM


packages/caravan-wallets/src/bcur2/decoder.ts
```diff
@@ -0,0 +1,190 @@
...
+ */
+
+import { ExtendedPublicKey, Network, BitcoinNetwork } from "@caravan/bitcoin";
+import { CryptoHDKey, CryptoAccount, URRegistryDecoder } from "@keystonehq/bc-ur-registry";
+
+/**
+ * Interface representing the decoded data from a BCUR2 QR code
+ * @interface DecodedData
+ */
+export interface DecodedData {
```
bucko13, line 14: isn't this just the decoder for the xpub handling? surely it's different for psbts?

packages/caravan-wallets/src/bcur2/decoder.ts
```diff
@@ -0,0 +1,190 @@
...
+          chaincode: chainCode.toString('hex'),
+          pubkey: key.toString('hex'), 
+          parentFingerprint: parentFp.readUInt32BE(0),
+          network: this.network
+        });
+
+        const xpub = xpubObj.toBase58();
+        console.log("Constructed xpub:", xpub);
+        if (!xpub) throw new Error("Failed to construct xpub from HDKey");
+        return { type, xpub, xfp, path };
```
bucko13, line 101: what is it that we're actually trying to return with the crypto-account decoding? This is basically the same result as the crypto-hdkey but trying to extract the information from the descriptor. I don't think this is what we want to be using this for. This, rather, is for interacting with wallets (which is what descriptors make up). we have a library that can also parse these properly.

packages/caravan-wallets/src/bcur2/decoder.ts
```diff
@@ -0,0 +1,190 @@
...
+        // Extract components from CryptoHDKey
+        const chainCode = hdKey.getChainCode();
+        const key = hdKey.getKey();
+        const origin = hdKey.getOrigin();
+        const xfp = origin?.getSourceFingerprint()?.toString("hex")?.toUpperCase();
+        const path = origin?.getPath();
+        const depth = origin?.getDepth() || 0;
+        const components = origin?.getComponents() || [];
+        const index = components.length > 0 ? components[components.length - 1]?.getIndex() || 0 : 0;
+        const parentFp = origin?.getSourceFingerprint() || Buffer.alloc(4);
```
bucko13, line 86: use [getParentFingerprint](https://github.com/KeystoneHQ/ur-registry/blob/43e30090321108ca72826dab7a15edac612012a5/src/CryptoHDKey.ts#L65) from the hdKey instead.

bucko13 (changes requested):
I think we need to fix the decoder especially. Lots of confusing things there. it's not clear to me what the difference between the crypto-account and crypto-hdkey are. hdkey seems like more of what we want maybe? But the crypto-account should be possible to get, i'm just not seeing how we know the right descriptor (and we need to fix how we encode the xpub.

packages/caravan-wallets/src/bcur2/decoder.ts
krrish-sehgal: answered on discord.

packages/caravan-wallets/src/bcur2/decoder.ts
krrish-sehgal: okay , like initially to match with the stuff caravan test suite was matching we need these things and since the BCUR for seedsigner was actually returning type acccount here, I am using the crypto-account to handle this.

packages/caravan-wallets/src/bcur2/decoder.ts
```diff
@@ -0,0 +1,190 @@
...
+   * @returns {DecodedData|null} The decoded wallet data or null if error
+   */
+  private handleDecodedResult(type: string, cbor: Buffer): DecodedData | null {
+    try {
+      if (type === "crypto-account") {
+        const account = CryptoAccount.fromCBOR(cbor);
+        const descriptors = account.getOutputDescriptors();
+        if (!descriptors.length) throw new Error("No output descriptors found");
+        
+        const hdKey = descriptors[0].getCryptoKey();
```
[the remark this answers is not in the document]
krrish-sehgal: There's only 1 descriptor here (I tried logging the descriptors), so its not arbitrary.
Screenshot 2025-06-06 at 10 08 18 PM

packages/caravan-wallets/src/bcur2/decoder.ts
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Processes decoded CBOR data based on the UR type
+   * @private
+   * @param {string} type - The UR type (crypto-account or crypto-hdkey)
+   * @param {Buffer} cbor - The decoded CBOR data
+   * @returns {DecodedData|null} The decoded wallet data or null if error
+   */
+  private handleDecodedResult(type: string, cbor: Buffer): DecodedData | null {
+    try {
+      if (type === "crypto-account") {
```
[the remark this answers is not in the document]
krrish-sehgal: Will do this for sure. Thanks!

packages/caravan-wallets/src/bcur2/decoder.ts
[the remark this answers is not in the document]
krrish-sehgal: done

packages/caravan-wallets/src/bcur2/decoder.ts
krrish-sehgal: done, changed to extendedPublickKeyData

.changeset/green-coins-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@caravan/wallets": minor
+---
+
+This introduces a support for exporting xpubs in caravan through BCUR2 Qr codes making wallets like seedsigner , jade etc compatible for the mentioned purpose.
```
bucko13, line 5:
```suggestion
This introduces a support for exporting xpubs in caravan through BCUR2 Qr codes making wallets like seedsigner, jade etc compatible for the mentioned purpose.
```

apps/coordinator/src/tests/extendedPublicKeys.jsx
```diff
@@ -16,6 +17,18 @@ class ExportExtendedPublicKeyTest extends Test {
     if (this.params.keystore === HERMIT) {
       tempResult = this.interaction().parse(result);
     }
+
+    // BCUR2 returns a different structure
```
bucko13, line 21: Actually, you could get the pubkey of this too. 

```typescript
import { ExtendedPublicKey } from "@caravan/bitcoin";
...
const pubkey = (ExtendedPublicKey.fromBase58(tempResult.xpub)).pubkey
```

packages/caravan-wallets/src/index.ts
```diff
@@ -648,6 +658,8 @@ export function ConfigAdapter({
 export * from "./interaction";
 export * from "./bitbox";
 export * from "./bcur";
+export * from "./bcur2/interactions";
+export * from "./bcur2/decoder";
```
bucko13, line 662: I wonder, do you think we need to export the decoder even from the module or is this just an internal tool for the interactions to use?

apps/coordinator/src/components/BCUR2/BCUR2Reader.tsx
```diff
@@ -0,0 +1,136 @@
...
+  startText = "Start QR Scan",
+  width = 300,
+}) => {
+  const [isScanning, setIsScanning] = useState(false);
+  const [error, setError] = useState("");
+  const decoder = useMemo(() => new BCURDecoder2(), []);
+  const statusRef = useRef<"idle" | "active" | "complete" | "error">("idle");
+
+  const handleStart = () => {
+    onStart?.();
```
bucko13, line 27: why might this be undefined?

apps/coordinator/src/components/BCUR2/BCUR2Reader.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const handleQRResult = (result: any, scanError: any) => {
+    if (statusRef.current !== "active") return;
+
+    if (scanError) return;
+
+    const text = result?.getText?.();
+    if (!text || !text.toLowerCase().startsWith("ur:")) return;
+
+    try {
+      console.log("Received QR part:", text);
```
bucko13, line 51: Can we clean up the console logs. At some point I think this would be good to put as visual elements in the actual component that we maybe publish as its own package. so a progress bar for parts received and then "started" and "completed" feedback.

bucko13 (changes requested):
This looks great! A few comments and questions, mostly non-blocking

packages/caravan-wallets/src/index.ts
krrish-sehgal: we are exporting the decoder to be used by the BCUR2reader component.

PR #305, opened 2025-06-07
Import Partially Signed PSBT works now 

### LINKED ISSUE #162: Import PSBT is not importing partial signatures
Recreation steps:

1. Craft a transaction in caravan, download unsigned PSBT.
2. Sign transaction outside of caravan, copy partially signed PSBT
3. Import wallet config again, go to "Send", "Import PSBT"
4. Import the partially signed PSBT
5. Note that the signature did not carry over

This definitely feels like a bug, let me know if I am wrong.

Unsigned PSBT
```
cHNidP8BAHwBAAAAAsEFicXI4HjydjPnHnQeERXIJrOakYEUt/Ic0fCNiSN+AAAAAAD/////a2EtzcyeJcYcr/4SikTPdrvNCcIaCM3zNN232jfrlNkAAAAAAP////8BEBIRAAAAAAAXqRQZeKGMQeKp8RopRKHSpilPWyhzeYcAAAAATwEENYfPBBN2KQ8AAAACbVu7qYOg/Zgx5UuIN9Q1MNawFWnbzeruGp0ynLn5rtkDT/UDSQiraRWeqaqGwld0OpC5ATt+0TwadVMQd2ucfVkU6Pvs4S0AAIABAACAAAAAgAIAAABPAQQ1h88EjMepUQAAAAGa5TgU/xdnWvwtwVzx1qK8cW9w8BbrpxAwhmX4Si3nowPfnUTx/SENNZxB6nif1Iuzk0O5JE+XJKMjtVYjjPALshTrp0AYLQAAgAEAAIAAAACAAQAAAE8BBDWHzwSjrSFyAAABv9Ze7/aOPjhZ0S7GPKPDSolzOxQpnwenDbd/FWRvzd4vAwemma5/w5iGmElsDRx4LXFy57VhpZD6ggZS45OUnxfwFO+l2RYtAACAAQAAgMgAAIC/AQAAAAEA/XoCAQAAAAK8ZFiSabp7iaoT+JCKmDFEKvk8HWFqNEZ+v6T8iwPO3wEAAAD9/QAASDBFAiEA2H3veISPA4cD4m9AKrepBk1p7xtCSa78Fek/83BTnoECICT3GbPisSB9zBpogTREnfTTt8aNcYUKB4j6q68TPImrAUcwRAIgWvguGm3UITzYqwcZoNGNTsBr6OaX/cXGp8AhVs0E9bACIAQ3Su7WIyj9NLXv5iX9o3kD5HWI/KT8fxEDt6082sbeAUxpUiEChfGT5eaCapf4SbMpIe9GY0RoOTiwvbZl8mqQT3g0urshAxWLarlpmvSVuAXImxrOkOLFk/VSzyrDgxzjtEvdHNflIQOxT5pPGjk6Lx5tJzgKAH2oy9p5lfBFGMZ5CDXFdAkZQFOu/////40vbk2ThbdeCQZt4ACFzNp5Q9MEQDgJoAdgAclrWsLpAQAAAP39AABIMEUCIQC2Z8196raYGQHfChJf09yv6iKqFz7Mb8Wewl483UAOmQIgTJuLLwzQwRKsm6sXKxZn543O/ZGQSVavikwtH4ZHlOoBRzBEAiA4saVr9sH6UoGmwiAThIltYnnGHrO6GZTfytRZjgaT5gIgabqZJjjOVashh8EeLhbqTECnQ7eLOtRtXZKP/17NzTYBTGlSIQI3Rz87e/zu3iIO/lJdl7Iok3MAYoFlzZCCg5jde5EhaiEDmKadb5+lSmLrPjnBhj0L08XFoupOMSIvlUCsCZXHYIkhA71pR/I1wYn2W4eLzpfZ0AvznQNGVGu09t/MtGQuHP0LU67/////AeTUAQAAAAAAF6kUX2Zv5R3hHPxTo5sbv1WdGtsu1VSHAAAAAAEEaVIhAnU38sN/bt9PgCkeln0letuJf8keK8uczjSV3y19wVnAIQKlDi6DJJ07NNdjn2mKX/LZXX+AArwN+Cj58knyUCQEgSEDbmAjhYdwE6myUgIxvrAHOIbVljqV1w6+Sy921oWVwcRTriIGAnU38sN/bt9PgCkeln0letuJf8keK8uczjSV3y19wVnAHOunQBgtAACAAQAAgAAAAIABAAAAAAAAAA0AAAAiBgKlDi6DJJ07NNdjn2mKX/LZXX+AArwN+Cj58knyUCQEgRzo++zhLQAAgAEAAIAAAACAAgAAAAAAAAANAAAAIgYDbmAjhYdwE6myUgIxvrAHOIbVljqV1w6+Sy921oWVwcQc76XZFi0AAIABAACAyAAAgL8BAAAAAAAADQAAAAABAP2DAQEAAAAAAQLO49NC6Hph1Jsyze8C2TGE9sGYZnfYrt12koRRuRcujgAAAAAXFgAUlOPF22EchDOavNt6As9cpxG+4r79////9fZweQyZdFnV28bYCQSgMeqWMRbzT+SmiXREmumDYusAAAAAFxYAFNua+y0WxZyyvndta0uklHbMJIUV/f///wFAQg8AAAAAABepFCqLo1MKE79Dnxqs9xPcK59IZLNLhwJIMEUCIQDmd8yj+TXDsBIfMj1naCIswtH2n1ZB3SE9Ta6mJX5qQAIgBg66OBdY7gSNwRrG2WvA2qdKkQxILy+uI8O/rSpiYyIBIQO0Az9GwbQ1WG0z4h1L8iGqlLITGmkwiF57lKUaLSafbQJHMEQCIHgbH5Xtkw57r2Ge0ZpRqyv4rqtw1pkws2piKGN3zDYkAiAO54W+s8OgHd6stEVoGe+ZRA5Tsish/vYJLWRrw4rhNwEhAhVcKLhua9a/YPYHfvvETICWaM4Vtjs+AQavpuP6bgOBAAAAAAEEaVIhAwSzj3Vs77p0tOl+OR1FQFzrD7tH1TCVw3ru0HXHGZyZIQN/NVjAivWeKk519c866JaZAJY6DMXNaOjUfSf+QewtKCEDm8saiYi21P2uqlAlUPGfM2ZBdh0y5zbta3hurbB44dhTriIGAwSzj3Vs77p0tOl+OR1FQFzrD7tH1TCVw3ru0HXHGZyZHO+l2RYtAACAAQAAgMgAAIC/AQAAAAAAAAsAAAAiBgN/NVjAivWeKk519c866JaZAJY6DMXNaOjUfSf+QewtKBzrp0AYLQAAgAEAAIAAAACAAQAAAAAAAAALAAAAIgYDm8saiYi21P2uqlAlUPGfM2ZBdh0y5zbta3hurbB44dgc6Pvs4S0AAIABAACAAAAAgAIAAAAAAAAACwAAAAAA
```

Partially signed PSBT
```
cHNidP8BAHwBAAAAAsEFicXI4HjydjPnHnQeERXIJrOakYEUt/Ic0fCNiSN+AAAAAAD/////a2EtzcyeJcYcr/4SikTPdrvNCcIaCM3zNN232jfrlNkAAAAAAP////8BEBIRAAAAAAAXqRQZeKGMQeKp8RopRKHSpilPWyhzeYcAAAAATwEEiLIeBBN2KQ8AAAACbVu7qYOg/Zgx5UuIN9Q1MNawFWnbzeruGp0ynLn5rtkDT/UDSQiraRWeqaqGwld0OpC5ATt+0TwadVMQd2ucfVkU6Pvs4S0AAIABAACAAAAAgAIAAABPAQSIsh4EjMepUQAAAAGa5TgU/xdnWvwtwVzx1qK8cW9w8BbrpxAwhmX4Si3nowPfnUTx/SENNZxB6nif1Iuzk0O5JE+XJKMjtVYjjPALshTrp0AYLQAAgAEAAIAAAACAAQAAAE8BBIiyHgSjrSFyAAABv9Ze7/aOPjhZ0S7GPKPDSolzOxQpnwenDbd/FWRvzd4vAwemma5/w5iGmElsDRx4LXFy57VhpZD6ggZS45OUnxfwFO+l2RYtAACAAQAAgMgAAIC/AQAAAAEA/XoCAQAAAAK8ZFiSabp7iaoT+JCKmDFEKvk8HWFqNEZ+v6T8iwPO3wEAAAD9/QAASDBFAiEA2H3veISPA4cD4m9AKrepBk1p7xtCSa78Fek/83BTnoECICT3GbPisSB9zBpogTREnfTTt8aNcYUKB4j6q68TPImrAUcwRAIgWvguGm3UITzYqwcZoNGNTsBr6OaX/cXGp8AhVs0E9bACIAQ3Su7WIyj9NLXv5iX9o3kD5HWI/KT8fxEDt6082sbeAUxpUiEChfGT5eaCapf4SbMpIe9GY0RoOTiwvbZl8mqQT3g0urshAxWLarlpmvSVuAXImxrOkOLFk/VSzyrDgxzjtEvdHNflIQOxT5pPGjk6Lx5tJzgKAH2oy9p5lfBFGMZ5CDXFdAkZQFOu/////40vbk2ThbdeCQZt4ACFzNp5Q9MEQDgJoAdgAclrWsLpAQAAAP39AABIMEUCIQC2Z8196raYGQHfChJf09yv6iKqFz7Mb8Wewl483UAOmQIgTJuLLwzQwRKsm6sXKxZn543O/ZGQSVavikwtH4ZHlOoBRzBEAiA4saVr9sH6UoGmwiAThIltYnnGHrO6GZTfytRZjgaT5gIgabqZJjjOVashh8EeLhbqTECnQ7eLOtRtXZKP/17NzTYBTGlSIQI3Rz87e/zu3iIO/lJdl7Iok3MAYoFlzZCCg5jde5EhaiEDmKadb5+lSmLrPjnBhj0L08XFoupOMSIvlUCsCZXHYIkhA71pR/I1wYn2W4eLzpfZ0AvznQNGVGu09t/MtGQuHP0LU67/////AeTUAQAAAAAAF6kUX2Zv5R3hHPxTo5sbv1WdGtsu1VSHAAAAACICA25gI4WHcBOpslICMb6wBziG1ZY6ldcOvksvdtaFlcHESDBFAiEAh22IiOctfB0GhnIeatJ6rqf0btBRi0Xr+Bxw62k2iBkCIDXQz2VtLBgZelZWU9tW5XAnsOsvGUM0mjB0JVguOgrWAQEEaVIhAnU38sN/bt9PgCkeln0letuJf8keK8uczjSV3y19wVnAIQKlDi6DJJ07NNdjn2mKX/LZXX+AArwN+Cj58knyUCQEgSEDbmAjhYdwE6myUgIxvrAHOIbVljqV1w6+Sy921oWVwcRTriIGAnU38sN/bt9PgCkeln0letuJf8keK8uczjSV3y19wVnAHOunQBgtAACAAQAAgAAAAIABAAAAAAAAAA0AAAAiBgKlDi6DJJ07NNdjn2mKX/LZXX+AArwN+Cj58knyUCQEgRzo++zhLQAAgAEAAIAAAACAAgAAAAAAAAANAAAAIgYDbmAjhYdwE6myUgIxvrAHOIbVljqV1w6+Sy921oWVwcQc76XZFi0AAIABAACAyAAAgL8BAAAAAAAADQAAAAABAP2DAQEAAAAAAQLO49NC6Hph1Jsyze8C2TGE9sGYZnfYrt12koRRuRcujgAAAAAXFgAUlOPF22EchDOavNt6As9cpxG+4r79////9fZweQyZdFnV28bYCQSgMeqWMRbzT+SmiXREmumDYusAAAAAFxYAFNua+y0WxZyyvndta0uklHbMJIUV/f///wFAQg8AAAAAABepFCqLo1MKE79Dnxqs9xPcK59IZLNLhwJIMEUCIQDmd8yj+TXDsBIfMj1naCIswtH2n1ZB3SE9Ta6mJX5qQAIgBg66OBdY7gSNwRrG2WvA2qdKkQxILy+uI8O/rSpiYyIBIQO0Az9GwbQ1WG0z4h1L8iGqlLITGmkwiF57lKUaLSafbQJHMEQCIHgbH5Xtkw57r2Ge0ZpRqyv4rqtw1pkws2piKGN3zDYkAiAO54W+s8OgHd6stEVoGe+ZRA5Tsish/vYJLWRrw4rhNwEhAhVcKLhua9a/YPYHfvvETICWaM4Vtjs+AQavpuP6bgOBAAAAACICAwSzj3Vs77p0tOl+OR1FQFzrD7tH1TCVw3ru0HXHGZyZRzBEAiBEv5jaVRWjyL/CEQNlWhvPk3nmhrl1w6Pkg72IBn2i2QIgQlOJduxuH2leZ3W1Kp7voouWJumr90t/HXd3fX42YLwBAQRpUiEDBLOPdWzvunS06X45HUVAXOsPu0fVMJXDeu7QdccZnJkhA381WMCK9Z4qTnX1zzrolpkAljoMxc1o6NR9J/5B7C0oIQObyxqJiLbU/a6qUCVQ8Z8zZkF2HTLnNu1reG6tsHjh2FOuIgYDBLOPdWzvunS06X45HUVAXOsPu0fVMJXDeu7QdccZnJkc76XZFi0AAIABAACAyAAAgL8BAAAAAAAACwAAACIGA381WMCK9Z4qTnX1zzrolpkAljoMxc1o6NR9J/5B7C0oHOunQBgtAACAAQAAgAAAAIABAAAAAAAAAAsAAAAiBgObyxqJiLbU/a6qUCVQ8Z8zZkF2HTLnNu1reG6tsHjh2Bzo++zhLQAAgAEAAIAAAACAAgAAAAAAAAALAAAAAAA=
```

Legend101Zz:
###  **What kind of change does this PR introduce?**

Feature Added – Fixes the issue where importing a partially signed PSBT would not retain the existing partial signatures.

---

###  **Issue Number:**

Fixes #162 – *Import PSBT is not importing partial signatures*

---

### **Recreation Steps (Before the Fix):**

1. Create an unsigned PSBT in Caravan and download it.
2. Sign it outside Caravan using a compatible wallet/tool.
3. Re-import the wallet config and go to **Send > Import PSBT**.
4. Import the partially signed PSBT.
5. ❌ Observe: The partial signatures are not reflected, and Caravan treats it as unsigned.

---

###  **Summary of Changes:**

* **Core Fix**: Caravan previously failed to recognize and import partial signatures when reloading a PSBT. This PR introduces the necessary logic to parse and preserve those signatures correctly.
* **UI Enhancement**: Updated the Import PSBT flow to give visual feedback when signatures are present.
* **Refactor**: Cleaned up the import logic and modularized signature-handling utilities.
* **Type Fixes**: Added type safety for PSBT fields related to partial signatures.

---

###  **Snapshots/Videos:**
Screenshot 2025-06-07 at 20 09 59

---

### **Does this PR introduce a breaking change?**

No – this is a backward-compatible fix.

---

### **Checklist**

* [x] I have tested my changes thoroughly.
* [x] I have added or updated tests to cover my changes.
* [x] I have verified that test coverage meets or exceeds 95%.
* [x] I have run the test suite locally, and all tests pass.
* [x] I have followed the project's coding style and conventions.
* [x] I have created a changeset to document my changes (`npm run changeset`).

### DIFF
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -418,7 +418,6 @@ export function importPSBT(psbtText) {
       // ==== Extract and import signatures ====
       try {
         const signatureSets = extractSignaturesFromPSBT(psbt, inputs);
-
         if (signatureSets.length > 0) {
           // Map signatures to Caravan's signature importers
           const importerData = mapSignaturesToImporters(signatureSets);
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -388,8 +388,7 @@ export function setSignaturesFromPsbt(psbt) {
         // Transform raw signature data into Caravan's signature importer format
         // This maps each signature set to an "importer" (numbered 1, 2, 3...)
         // that corresponds to Caravan's UI signature input slots
-        const state = getState();
-        const importerData = mapSignaturesToImporters(state, signatureSets);
+        const importerData = mapSignaturesToImporters(signatureSets);
 
         // === STEP 3: Update Redux state for each signature set ===
         // For each complete signature set, we need to update multiple parts
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -4,7 +4,11 @@ import {
   satoshisToBitcoins,
 } from "@caravan/bitcoin";
 import { getSpendableSlices, getConfirmedBalance } from "../selectors/wallet";
-import { selectInputsFromPSBT } from "../selectors/transaction";
+import {
+  selectInputsFromPSBT,
+  mapSignaturesToImporters,
+  extractSignaturesFromPSBT,
+} from "../selectors/transaction";
 import {
   setSignatureImporterSignature,
   setSignatureImporterPublicKeys,
@@ -13,11 +17,7 @@ import {
 } from "./signatureImporterActions";
 
 import { DUST_IN_BTC } from "../utils/constants";
-import {
-  loadPsbt,
-  extractSignaturesFromPSBT,
-  mapSignaturesToImporters,
-} from "../utils/psbtUtils";
+import { loadPsbt } from "../utils/psbtUtils";
 
 export const CHOOSE_PERFORM_SPEND = "CHOOSE_PERFORM_SPEND";
 
@@ -348,6 +348,104 @@ function setOutputsFromPSBT(psbt) {
   };
 }
 
+/**
+ * Extracts and imports partial signatures from a PSBT into Caravan's signature importers.
+ *
+ * This is the main entry point for processing signatures when importing a PSBT that
+ * has already been partially signed by other wallets/tools. It handles the entire
+ * workflow from signature extraction to updating the Redux state.
+ *
+ * @description The function performs these key steps:
+ *   1. Extracts signature data from the PSBT using our composed selectors
+ *   2. Transforms the signatures into Caravan's internal format
+ *   3. Updates the signature importers state via multiple Redux actions
+ *   4. Handles errors gracefully without breaking the PSBT import flow
+ *
+ * @param {Psbt} psbt - The PSBT object containing partial signatures to extract
+ *
+ * @returns {Function} Redux thunk function that takes (dispatch, getState)
+ *
+ * @example
+ * // In your PSBT import flow:
+ * dispatch(setSignaturesFromPsbt(psbt));
+ *
+ * @see extractSignaturesFromPSBT - For the core signature extraction logic
+ * @see mapSignaturesToImporters - For signature format transformation
+ *
+ * @throws {Error} Logs warnings for signature extraction failures but doesn't throw
+ *                 to avoid breaking the overall PSBT import process
+ */
+export function setSignaturesFromPsbt(psbt) {
+  return (dispatch, getState) => {
+    try {
+      // === STEP 1: Extract signature sets from the PSBT ===
+      // This uses our composed selectors to:
+      // - Get wallet UTXOs that match PSBT inputs (selectInputsFromPSBT)
+      // - Parse partial signatures from PSBT data
+      // - Group signatures by signer (one signer signs ALL inputs)
+      const signatureSets = extractSignaturesFromPSBT(getState(), psbt);
+
+      // === STEP 2: Process signatures if any were found ===
+      if (signatureSets.length > 0) {
+        // Transform raw signature data into Caravan's signature importer format
+        // This maps each signature set to an "importer" (numbered 1, 2, 3...)
+        // that corresponds to Caravan's UI signature input slots
+        const importerData = mapSignaturesToImporters(
+          getState(),
+          signatureSets,
+        );
+
+        // === STEP 3: Update Redux state for each signature set ===
+        // For each complete signature set, we need to update multiple parts
+        // of the signature importer state. This is Caravan's pattern for
+        // managing signature data across the signing workflow.
+        importerData.forEach((sigData) => {
+          // Set the raw signature data (hex-encoded signatures)
+          dispatch(
+            setSignatureImporterSignature(
+              sigData.importerNumber,
+              sigData.signatures,
+            ),
+          );
+
+          // Set the public keys that correspond to these signatures
+          // These are used for signature verification
+          dispatch(
+            setSignatureImporterPublicKeys(
+              sigData.importerNumber,
+              sigData.publicKeys,
+            ),
+          );
+
+          // Mark this importer as "finalized" - meaning we have a complete
+          // signature set and it's ready for use in transaction construction
+          dispatch(setSignatureImporterFinalized(sigData.importerNumber, true));
+
+          // Set the complete signature data object
+          // This is Caravan's "master" action that bundles everything together
+          // and is used by the signing components to display signature status
+          dispatch(
+            setSignatureImporterComplete(sigData.importerNumber, {
+              signature: sigData.signatures,
+              publicKeys: sigData.publicKeys,
+              finalized: true,
+            }),
+          );
+        });
+      }
+    } catch (signatureError) {
+      // === ERROR HANDLING ===
+      // We intentionally don't re-throw here because signature extraction
+      // failing shouldn't break the entire PSBT import process. The user
+      // can still import the PSBT structure and sign it manually.
+      console.warn(
+        "⚠️ Failed to extract signatures, but continuing with PSBT import:",
+        signatureError.message,
+      );
+    }
+  };
+}
+
 export function importPSBT(psbtText) {
   return (dispatch, getState) => {
     let state = getState();
@@ -396,56 +494,7 @@ export function importPSBT(psbtText) {
       dispatch(setFee(fee));
 
       // ==== Extract and import signatures ====
-      try {
-        const signatureSets = extractSignaturesFromPSBT(psbt, inputs);
-        if (signatureSets.length > 0) {
-          // Map signatures to Caravan's signature importers
-          const importerData = mapSignaturesToImporters(signatureSets);
-
-          // Update signature importers state
-          importerData.forEach((sigData) => {
-            // Set the signature data for this importer
-            dispatch(
-              setSignatureImporterSignature(
-                sigData.importerNumber,
-                sigData.signatures,
-              ),
-            );
-
-            // Set the public keys
-            dispatch(
-              setSignatureImporterPublicKeys(
-                sigData.importerNumber,
-                sigData.publicKeys,
-              ),
-            );
-
-            // Mark as finalized since we have a complete signature set
-            dispatch(
-              setSignatureImporterFinalized(sigData.importerNumber, true),
-            );
-
-            // Use setSignatureImporterComplete for the complete signature
-            dispatch(
-              setSignatureImporterComplete(sigData.importerNumber, {
-                signature: sigData.signatures,
-                publicKeys: sigData.publicKeys,
-                finalized: true,
-              }),
-            );
-          });
-        } else {
-          console.log(
-            "ℹ️ No complete signature sets found in PSBT (this is normal for unsigned PSBTs)",
-          );
-        }
-      } catch (signatureError) {
-        // Don't fail the entire import if signature extraction fails
-        console.warn(
-          "⚠️ Failed to extract signatures, but continuing with PSBT import:",
-          signatureError.message,
-        );
-      }
+      dispatch(setSignaturesFromPsbt(psbt));
 
       // Finalize the transaction
       dispatch(finalizeOutputs(true));
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -320,10 +320,7 @@ export function setMaxSpendOnOutput(outputIndex) {
  * @returns A thunk function for Redux
  */
 function setOutputsFromPSBT(psbt) {
-  return (
-    dispatch,
-    _getState, // eslint-disable-line @typescript-eslint/no-unused-vars
-  ) => {
+  return (dispatch) => {
     let outputsTotalSats = new BigNumber(0);
 
     psbt.txOutputs.forEach((output, outputIndex) => {
@@ -383,17 +380,16 @@ export function setSignaturesFromPsbt(psbt) {
       // - Get wallet UTXOs that match PSBT inputs (selectInputsFromPSBT)
       // - Parse partial signatures from PSBT data
       // - Group signatures by signer (one signer signs ALL inputs)
-      const signatureSets = extractSignaturesFromPSBT(getState(), psbt);
+      const state = getState();
+      const signatureSets = extractSignaturesFromPSBT(state, psbt);
 
       // === STEP 2: Process signatures if any were found ===
       if (signatureSets.length > 0) {
         // Transform raw signature data into Caravan's signature importer format
         // This maps each signature set to an "importer" (numbered 1, 2, 3...)
         // that corresponds to Caravan's UI signature input slots
-        const importerData = mapSignaturesToImporters(
-          getState(),
-          signatureSets,
-        );
+        const state = getState();
+        const importerData = mapSignaturesToImporters(state, signatureSets);
 
         // === STEP 3: Update Redux state for each signature set ===
         // For each complete signature set, we need to update multiple parts
@@ -446,6 +442,28 @@ export function setSignaturesFromPsbt(psbt) {
   };
 }
 
+/**
+ * Calculates and sets the transaction fee based on current state.
+ *
+ * This action creator encapsulates the fee calculation logic by:
+ * 1. Getting the current inputsTotalSats from state
+ * 2. Calculating fee as the difference between inputs and outputs
+ * 3. Converting from satoshis to bitcoins
+ * 4. Dispatching the setFee action
+ *
+ * @param {BigNumber} outputsTotalSats - Total satoshis in all outputs
+ * @returns {Function} Redux thunk function
+ */
+export function setFeeFromState(outputsTotalSats) {
+  return (dispatch, getState) => {
+    const state = getState();
+    const inputsTotalSats = BigNumber(state.spend.transaction.inputsTotalSats);
+    const feeSats = inputsTotalSats.minus(outputsTotalSats);
+    const fee = satoshisToBitcoins(feeSats);
+    dispatch(setFee(fee));
+  };
+}
+
 export function importPSBT(psbtText) {
   return (dispatch, getState) => {
     let state = getState();
@@ -485,15 +503,9 @@ export function importPSBT(psbtText) {
       const { outputsTotalSats } = dispatch(setOutputsFromPSBT(psbt));
 
       // Calculate and set fee
-      state = getState(); //Now we get updated state after setting inputs
-      const inputsTotalSats = BigNumber(
-        state.spend.transaction.inputsTotalSats,
-      );
-      const feeSats = inputsTotalSats.minus(outputsTotalSats);
-      const fee = satoshisToBitcoins(feeSats);
-      dispatch(setFee(fee));
+      dispatch(setFeeFromState(outputsTotalSats));
 
-      // ==== Extract and import signatures ====
+      // ==== Extract and import signatures (If they are present)====
       dispatch(setSignaturesFromPsbt(psbt));
 
       // Finalize the transaction
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -1,5 +1,4 @@
 import BigNumber from "bignumber.js";
-import { reverseBuffer } from "bitcoinjs-lib/src/bufferutils.js";
 import {
   estimateMultisigTransactionFee,
   satoshisToBitcoins,
@@ -311,42 +310,20 @@ export function setMaxSpendOnOutput(outputIndex) {
 }
 
 /**
- * Processes inputs from PSBT and matches them with wallet UTXOs.
- */
-function processInputsFromPSBT(psbt, state) {
-  const createInputIdentifier = (txid, index) => `${txid}:${index}`;
-
-  const inputIdentifiers = new Set(
-    psbt.txInputs.map((input) => {
-      const txid = reverseBuffer(input.hash).toString("hex");
-      return createInputIdentifier(txid, input.index);
-    }),
-  );
-
-  const inputs = [];
-  getSpendableSlices(state).forEach((slice) => {
-    Object.entries(slice.utxos).forEach(([, utxo]) => {
-      const inputIdentifier = createInputIdentifier(utxo.txid, utxo.index);
-      if (inputIdentifiers.has(inputIdentifier)) {
-        const input = {
-          ...utxo,
-          multisig: slice.multisig,
-          bip32Path: slice.bip32Path,
-          change: slice.change,
-        };
-        inputs.push(input);
-      }
-    });
-  });
-
-  return inputs;
-}
-
-/**
- * Processes outputs from PSBT and adds them to the transaction.
+ * Sets outputs from a PSBT into Redux state.
+ * This thunk:
+ *  - Iterates over each PSBT output
+ *  - Dispatches actions to reflect each output in the UI state
+ *  - Identifies and sets the change output if applicable
+ *
+ * @param psbt - The partially signed Bitcoin transaction
+ * @returns A thunk function for Redux
  */
 function setOutputsFromPSBT(psbt) {
-  return (dispatch, getState) => {
+  return (
+    dispatch,
+    _getState, // eslint-disable-line @typescript-eslint/no-unused-vars
+  ) => {
     let outputsTotalSats = new BigNumber(0);
 
     psbt.txOutputs.forEach((output, outputIndex) => {
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -6,8 +6,8 @@ import {
 import { getSpendableSlices, getConfirmedBalance } from "../selectors/wallet";
 import {
   selectInputsFromPSBT,
-  mapSignaturesToImporters,
-  extractSignaturesFromPSBT,
+  selectSignaturesFromPSBT,
+  selectSignaturesForImporters,
 } from "../selectors/transaction";
 import {
   setSignatureImporterSignature,
@@ -380,13 +380,13 @@ export function setSignaturesFromPsbt(psbt) {
     // - Parse partial signatures from PSBT data
     // - Group signatures by signer (one signer signs ALL inputs)
     const state = getState();
-    const signatureSets = extractSignaturesFromPSBT(state, psbt);
+    const signatureSets = selectSignaturesFromPSBT(state, psbt);
     // === STEP 2: Process signatures if any were found ===
     if (signatureSets.length > 0) {
       // Transform raw signature data into Caravan's signature importer format
       // This maps each signature set to an "importer" (numbered 1, 2, 3...)
       // that corresponds to Caravan's UI signature input slots
-      const importerData = mapSignaturesToImporters(signatureSets);
+      const importerData = selectSignaturesForImporters(state, psbt);
       // === STEP 3: Update Redux state for each signature set ===
       // For each complete signature set, we need to update multiple parts
       // of the signature importer state. This is Caravan's pattern for
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -3,12 +3,21 @@ import { reverseBuffer } from "bitcoinjs-lib/src/bufferutils.js";
 import {
   estimateMultisigTransactionFee,
   satoshisToBitcoins,
-  networkData,
-  autoLoadPSBT,
 } from "@caravan/bitcoin";
 import { getSpendableSlices, getConfirmedBalance } from "../selectors/wallet";
+import {
+  setSignatureImporterSignature,
+  setSignatureImporterPublicKeys,
+  setSignatureImporterFinalized,
+  setSignatureImporterComplete,
+} from "./signatureImporterActions";
 
 import { DUST_IN_BTC } from "../utils/constants";
+import {
+  loadPsbt,
+  extractSignaturesFromPSBT,
+  mapSignaturesToImporters,
+} from "../utils/psbtUtils";
 
 export const CHOOSE_PERFORM_SPEND = "CHOOSE_PERFORM_SPEND";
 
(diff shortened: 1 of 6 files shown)


Legend101Zz:
cc: @bucko13 @jbrauck-unchained @krrish-sehgal

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -300,97 +309,179 @@ export function setMaxSpendOnOutput(outputIndex) {
...
+      const inputsTotalSats = BigNumber(
+        state.spend.transaction.inputsTotalSats,
+      );
+      const feeSats = inputsTotalSats.minus(outputsTotalSats);
+      const fee = satoshisToBitcoins(feeSats);
+      dispatch(setFee(fee));
+
+      // ==== Extract and import signatures ====
+      try {
+        const signatureSets = extractSignaturesFromPSBT(psbt, inputs);
```
bucko13, line 420: if you turn processInputs into a selector, then this can basically be another selector as well that can be composed with it because `inputs` would come from the state given a psbt. so `extractSignaturesFromPsbt` could only require the psbt as an argument, inputs would come from the other selector (to be built), `processInputs` or new name like `selectSpendableInputsFromPsbt` (or something) which can be used within this selector. This makes sure to keep the state management clean, and leverages memoization.

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -300,97 +309,179 @@ export function setMaxSpendOnOutput(outputIndex) {
...
+            dispatch(
+              setSignatureImporterComplete(sigData.importerNumber, {
+                signature: sigData.signatures,
+                publicKeys: sigData.publicKeys,
+                finalized: true,
+              }),
+            );
+          });
+        } else {
+          console.log(
```
bucko13, line 458: is this necessary? I'd rather just have a log when there are signatures and no if/else.

apps/coordinator/src/actions/transactionActions.js
bucko13: In fact, I bet this whole bit that's nested in a try/catch could just be put into its own action creator, which itself would call the selectors for what it needs. 

So then, all you would have to do is in this line would be `dispatch(setSignaturesFromPsbt(psbt))`, that itself would be a thunk-based action creator which could take the psbt and the state and then do what you're doing here (but more of it consolidated into selectors)

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -155,6 +165,95 @@ class TransactionPreview extends React.Component {
...
+      (importer) =>
+        importer.finalized &&
+        importer.signature &&
+        importer.signature.length > 0,
+    ).length;
+
+    return { signedCount, requiredSigners: requiredSigners || 0 };
+  };
+
+  renderSignatureStatus = () => {
```
bucko13, line 185: remember, this is an anti-pattern now. move this into its own component, probably something that would look like this:  

```
<SignatureStatus 
    signatureImporters={signatureImporters}
    requiredSigners={requiredSigners} 
/>
```

bucko13: actually looks like you don't even need to pass these as props and just pull them in with `useSelector` to get from the state in the component

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,309 @@
...
+/**
+ * Extracts signatures from a PSBT and organizes them by signer.
+ *
+ * @param  psbt - The PSBT object
+ * @param  inputs - Array of input objects with multisig info
+ * @param  outputs - Array of output objects
+ * @param  network - Network (mainnet, testnet, etc.)
+ * @returns  Array of signature sets, each containing signatures for all inputs
+ */
+export function extractSignaturesFromPSBT(psbt: Psbt, inputs: Input[]) {
```
bucko13, line 121: noted in the other file, but this should probably be a selector that way you can pull the inputs from the psbt using the redux store.

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -300,97 +309,179 @@ export function setMaxSpendOnOutput(outputIndex) {
...
-    state = getState();
-    const inputsTotalSats = BigNumber(state.spend.transaction.inputsTotalSats);
-    const feeSats = inputsTotalSats - outputsTotalSats;
-    const fee = satoshisToBitcoins(feeSats);
-    dispatch(setFee(fee));
+      dispatch(setInputs(inputs));
 
-    dispatch(finalizeOutputs(true));
+      // ==== PROCESS INPUTS ====
+      const { outputsTotalSats } = processOutputsFromPSBT(psbt, dispatch);
```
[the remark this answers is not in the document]
Legend101Zz: thanks changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/a404bae41769fd961edb6dbd000605864bb9d3f5#diff-d9901353915f776b9ca631bd464aada7a2a07e25535fa5f28c8ca24027b5f5f4R410)

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
Legend101Zz: thanks , changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/247c7a7bd89cc87005c73b351432c3c49554d4de)

apps/coordinator/src/actions/transactionActions.js
Legend101Zz: removed that now ...

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -0,0 +1,244 @@
...
+      console.warn(
+        `Signature validation returned true but expected pubkey string for input ${inputIndex}`,
+      );
+      return false;
+    }
+    // Return the pubkey string or false
+    return validatedPubkey;
+  } catch (error) {
+    console.error(
+      `❌ Signature validation failed for input ${inputIndex + 1}:`,
```
bucko13, line 239: Aren't you doing the same thing above? I think we can reduce some lines of code by just doing the validation in one place and then throw if there's an invalid signature.

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -0,0 +1,244 @@
...
+    const validatedPubkey = validateMultisigPsbtSignature(
+      originalPsbt.toBase64(),
+      inputIndex,
+      signature,
+      inputs[inputIndex].amountSats,
+    );
+
+    // Handle the case where validateMultisigPsbtSignature returns boolean true
+    // Convert true to false since we only want string (pubkey) or false
+    if (validatedPubkey === true) {
```
bucko13, line 229: I don't understand what's going on here? why would it be true but then we want to say it's false? Let's just throw an error if it's false. Then we don't even need to check for booleans.

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+): Psbt | null {
+  const options = { network: networkData(network) };
+
+  try {
+    // Try to determine the PSBT version
+    const version = getPsbtVersionNumber(psbtText);
+
+    if (version >= 2) {
+      // It's a PSBTv2, convert it to PSBTv0 format for compatibility
+      console.log(
```
bucko13, line 85: Do you think it's worth keeping the console logs? I'm less concerned with errors and warnings but logs I think can pollute the console.

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+        `Detected PSBTv2 (version ${version}), converting to PSBTv0 format`,
+      );
+      try {
+        const psbtv2 = new PsbtV2(psbtText, true); // Allow version 1 transactions if needed
+        const psbtv0Base64 = psbtv2.toV0("base64");
+        // Use the autoLoadPSBT from @caravan/psbt
+        return psbtPackageAutoLoad(psbtv0Base64, options);
+      } catch (e) {
+        console.error("Failed to parse PSBTv2:", e);
```
bucko13, line 94: let's remove. if we're throwing this let the caller catch and log if necessary (or show to the user)

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+        throw new Error(`Failed to parse PSBTv2: ${e.message}`);
+      }
+    } else {
```
bucko13, line 97: I think this is an unnecessary else. you can just run the auto load if it's not version 2

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+  psbtText: string | Buffer,
+  network: Network,
+): Psbt | null {
+  const options = { network: networkData(network) };
+
+  try {
+    // Try to determine the PSBT version
+    const version = getPsbtVersionNumber(psbtText);
+
+    if (version >= 2) {
```
bucko13, line 83: why not this:
```suggestion
    if (version === 2) {
```

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Determines if a PSBT string/buffer is likely a PSBTv2.
+ * This is a quick check without full parsing.
+ *
+ * @param {string|Buffer} psbtText - The PSBT data
+ * @returns {boolean} True if likely PSBTv2, false otherwise
+ */
+export function isPsbtV2(psbtText: string | Buffer): boolean {
+  try {
+    const version = getPsbtVersionNumber(psbtText);
+    return version >= 2;
```
bucko13, line 121: Same:

```suggestion
    return version === 2;
```

bucko13 (changes requested):
Looking really good! Mostly aesthetic requests and cleanup. I will test after those changes are done.

apps/coordinator/src/actions/transactionActions.js
bucko13: Yep!

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -0,0 +1,244 @@
...
+ * Groups signatures by signer (same pubkey across all inputs = same signer).
+ *
+ * *Note* not a selector a plain util function
+ *
+ * @param  psbt - The PSBT object
+ * @param  inputSignatures - Signatures organized by input
+ * @param  inputs - Input objects for validation
+ * @returns  Array of signature sets
+ */
+function groupSignaturesBySigner(
```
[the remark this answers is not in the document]
Legend101Zz: yup .. I actually moved it here from the psbtUtils file ... was not exactly sure like where it belongs

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -0,0 +1,244 @@
...
+  psbt: Psbt,
+  inputSignatures: SignatureInfo[][],
```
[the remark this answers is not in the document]
Legend101Zz: ---

> Isn't this just extracted from the PSBT itself? I feel like there's some consolidation that could happen here.

Yes, you're right — it *can* be extracted directly from the PSBT. The idea here was to break down the large `extractSignaturesFromPSBT` function into a more modular and maintainable structure.

So like, if we've already computed `inputSignatures` earlier in the flow, there's no point in looping through the inputs again just to extract the same data. So it’s more about avoiding redundant computation and reusing what we already have available.

---

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -0,0 +1,244 @@
...
+            inputIndex,
+            psbt,
+            inputs,
+          );
+
+          if (validatedPubkey) {
+            signatureSet.push(sigForThisPubkey.signature as string);
+            publicKeySet.push(validatedPubkey);
+          } else {
+            console.warn(
```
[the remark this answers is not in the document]
Legend101Zz: > This means means the psbt has a bad signature right? should we just throw and reject that PSBT?


Yup, I had the same thought initially — but then I considered cases where someone might have a PSBT with one or more invalid signatures, yet the overall structure is still valid. They might still want to load it into Caravan just to inspect or work with the transaction.

If we throw and reject outright, we’d prevent them from doing that. So the idea was to gracefully handle signature validation errors, maybe by logging a warning or marking those signatures as invalid, but still allowing the PSBT to load.

Don't know if that is what we want though right ?

apps/coordinator/src/selectors/transaction.ts
Legend101Zz: yup ... it was part of my own debugging flow (like to pin-point where the exact issue was) that ended up getting into the final code 😅

apps/coordinator/src/selectors/transaction.ts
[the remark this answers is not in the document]
bucko13: I think you should throw. We shouldn't be manipulating psbts in a quiet way in caravan but rather error and present the issue front end center so the user can address accordingly. bip370.org (or bip174.org) are the places to inspect and change psbts as necessary. having to handle changing the psbt and communicating that to the user makes for really complicated UX in my opinion if you're going to do it right. 

Take this as an example. Right now, the error is just logged in the console and the consuming component has no way to catch this issue and communicate it to the user (we can do this with a throw however). So currently they don't even know that they were given a bad PSBT in the first place (maybe one of their cosigners is compromised? seems important thing to communicate).

apps/coordinator/src/selectors/transaction.ts
Legend101Zz: Ahh, right — I remember now. I added that as a quick workaround to satisfy the type checker. Since `validateMultisigPsbtSignature` only ever returns `false`, but its return type is defined as `boolean`, the type checker was shouting at me 😅

So yeah... I did this dumb hacky fix just to keep things passing at the time. I’ll clean it up properly now!

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -300,97 +309,208 @@ export function setMaxSpendOnOutput(outputIndex) {
...
+ *  - Dispatches actions to reflect each output in the UI state
+ *  - Identifies and sets the change output if applicable
+ *
+ * @param psbt - The partially signed Bitcoin transaction
+ * @returns A thunk function for Redux
+ */
+function setOutputsFromPSBT(psbt) {
+  return (
+    dispatch,
+    _getState, // eslint-disable-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/b276c0e64c8c48fe07c7a9bd42b0dd27a9796374#diff-d9901353915f776b9ca631bd464aada7a2a07e25535fa5f28c8ca24027b5f5f4R323)

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -300,97 +309,179 @@ export function setMaxSpendOnOutput(outputIndex) {
...
-    dispatch(finalizeOutputs(true));
+      // ==== PROCESS INPUTS ====
+      const { outputsTotalSats } = processOutputsFromPSBT(psbt, dispatch);
 
-    // In the future, if we want to support loading in signatures
-    // (or sets of signatures) included in a PSBT, we likely need to do
-    // that work here. Initial implementation just ignores any signatures
-    // included with the uploaded PSBT.
+      // Calculate and set fee
+      state = getState(); //Now we get updated state after setting inputs
```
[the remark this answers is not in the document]
Legend101Zz: added [now](https://github.com/caravan-bitcoin/caravan/pull/305/commits/b276c0e64c8c48fe07c7a9bd42b0dd27a9796374#diff-d9901353915f776b9ca631bd464aada7a2a07e25535fa5f28c8ca24027b5f5f4R457)

apps/coordinator/src/selectors/transaction.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/e5a216934181acc7e8a5bfcc47143d09681cb15e#diff-ccca8d8f6174d6337c1ec048c507375db73fddbfa21905393881dd08a43dff82R213)

apps/coordinator/src/utils/psbtUtils.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/52481152fe9ea9ee7bc1c6e31ba42960f3f2831d)

apps/coordinator/src/utils/psbtUtils.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/52481152fe9ea9ee7bc1c6e31ba42960f3f2831d)

apps/coordinator/src/utils/psbtUtils.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/52481152fe9ea9ee7bc1c6e31ba42960f3f2831d)

apps/coordinator/src/utils/psbtUtils.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/52481152fe9ea9ee7bc1c6e31ba42960f3f2831d)

apps/coordinator/src/utils/psbtUtils.ts
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/52481152fe9ea9ee7bc1c6e31ba42960f3f2831d)

Legend101Zz:
adding this [comment](https://github.com/caravan-bitcoin/caravan/pull/257#discussion_r2112645748) that @bucko13 wrote, here so we can discuss it further and resolve it :)

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -300,97 +309,219 @@ export function setMaxSpendOnOutput(outputIndex) {
...
+
+      // Calculate and set fee
+      dispatch(setFeeFromState(outputsTotalSats));
+
+      // ==== Extract and import signatures (If they are present)====
+      dispatch(setSignaturesFromPsbt(psbt));
+
+      // Finalize the transaction
+      dispatch(finalizeOutputs(true));
+    } catch (error) {
```
bucko13, line 512: this is superfluous imo. just let the error bubble up.

bucko13 (changes requested):
This is looking really really good. I think the redux code is a huge improvement over what we had before. The selectors could use a little more cleanup. The rest are aesthetic nits. Thanks for this!

jbrauck-unchained:
Screenshot 2025-06-22 at 4 10 05 PM
Testing with a coldcard it seemed to fail importing the partially signed btc txn

bucko13:
> Testing with a coldcard it seemed to fail importing the partially signed btc txn

Strange. That error comes from @caravan/psbt, but the base64 you shared there works in bip370.org (which uses the same library). 

I also noticed that psbt doesn't have the global xpubs and I wonder if that's related. I think this code should support that and the error doesn't appear linked to that anyway but just thought I'd mention it.

Legend101Zz:
> > Testing with a coldcard it seemed to fail importing the partially signed btc txn
> 
> Strange. That error comes from @caravan/psbt, but the base64 you shared there works in bip370.org (which uses the same library).
> 
> I also noticed that psbt doesn't have the global xpubs and I wonder if that's related. I think this code should support that and the error doesn't appear linked to that anyway but just thought I'd mention it.

I guess it's related to our disucssion yesterday  on discord  ... regarding the .psbt format

jbrauck-unchained:
Screenshot 2025-06-23 at 9 30 57 PM
Bang! It worked!

jbrauck-unchained:
Approved!

apps/coordinator/src/actions/transactionActions.js
Legend101Zz: removed now .. thanks :)

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -13,11 +13,154 @@ import {
...
+            isFullySigned
+              ? "success"
+              : hasPartialSignatures
+                ? "warning"
+                : "default"
+          }
+          variant={isFullySigned ? "filled" : "outlined"}
+        />
+        {isFullySigned && (
+          <Chip label="Fully Signed" color="success" size="small" />
```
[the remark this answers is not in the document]
Legend101Zz: Yup... 

changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/54168ec99b1f0c73f6b0780f900bb801c4256c3d) thanks :)

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -13,11 +13,154 @@ import {
...
+        )}
+        {hasPartialSignatures && (
+          <Chip label="Partially Signed" color="warning" size="small" />
```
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/305/commits/54168ec99b1f0c73f6b0780f900bb801c4256c3d) thanks :)

apps/coordinator/src/components/Wallet/WalletSpend.jsx
```diff
@@ -157,18 +169,55 @@ class WalletSpend extends React.Component {
...
-
+      const file = target.files[0];
       const fileReader = new FileReader();
       fileReader.onload = (event) => {
         try {
-          const psbtText = event.target.result;
-          importPSBT(psbtText);
+          const arrayBuffer = event.target.result;
+
+          if (this.isBinaryPSBT(arrayBuffer)) {
```
bucko13, line 178: non-blocking nit, but this doesn't need to be a class method.

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+export interface SignatureInfo {
+  signature: Buffer | string;
+  pubkey: string;
+  inputIndex: number;
+  signerIndex?: number; // Which signer in the multisig (0, 1, 2, etc.)
+  derivedPubkey: string; // The actual derived pubkey for this input
+}
+
+export /**
```
bucko13, line 57: mistake?

bucko13 (approved):
Good stuff! Huge win.

PR #307, opened 2025-06-08
Node.js Version Standardization ( also added Husky pre-commit hooks for Node.js version and lockfile validation )

### LINKED ISSUE #306: Node.js Version Incompatibility Causing package-lock.json Conflicts
---

# ⚠️ Node.js Version Incompatibility Causing `package-lock.json` Conflicts

## 🧩 The Problem

We’re running into repeated issues with `package-lock.json` churn and merge conflicts across the monorepo — and the root cause is environment inconsistency. Specifically, contributors are using different Node.js versions (e.g. v20 vs v24), which come bundled with **different npm versions**. These npm versions generate **incompatible lockfile formats**, leading to constant diffs and unnecessary frustration.

This breaks the [12-Factor App](https://12factor.net/dev-prod-parity)  principle of *environment parity*, and introduces friction in an otherwise smooth dev workflow.

---

##  Root Cause

Different versions of Node.js ship with different versions of npm, which in turn produce different lockfile versions:

| Node.js Version   | npm Version | Lockfile Format      |
| ----------------- | ----------- | -------------------- |
| Node 20 (LTS)     | npm v8      | `lockfileVersion: 2` |
| Node 24 (Current) | npm v10     | `lockfileVersion: 4` |

So when developers switch Node versions:

1. `npm install` silently upgrades/downgrades the lockfile format
2. Git shows massive diffs in `package-lock.json`
3. Merge conflicts crop up constantly
4. CI/CD or runtime bugs could appear due to subtle dependency changes

---

##  Why It Matters

*  Merge conflicts waste contributor time
*  Inconsistent dependency resolution across machines
*  CI/CD pipeline risks and instability
*  General dev friction and onboarding overhead

---

##  Proposed Solution

To eliminate this pain point, I propose enforcing a consistent Node.js and npm environment:

1. **Standardize on Node.js v20 (LTS) which majority of current packages use ** across all environments
2. **Add a `.nvmrc`** file to help contributors automatically switch Node versions with `nvm`
3. **Specify Node & npm version** explicitly using the `packageManager` field in `package.json`:

   ```json
   {
     "packageManager": "npm@8.19.4"
   }
   ```
4. **Update project docs** to clearly communicate Node and npm version requirements
5. **Configure CI/CD** to always use Node v20 + npm v8
6. **(Optional)** Add a pre-commit or pre-install hook to warn if the wrong Node version is being used

---

##  References

* [npm lockfileVersion explained](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json#lockfileversion)
* [12-Factor App: Dev/prod parity](https://12factor.net/dev-prod-parity)
* [Node.js release schedule](https://nodejs.org/en/about/releases/)

---

Legend101Zz:
Fixes #306 

### Problem
Contributors using different Node.js versions (v20 vs v24) were experiencing conflicts in `package-lock.json` due to incompatible lockfile formats. This caused:
- Merge conflicts in lockfiles  
- Inconsistent dependency resolution
- Wasted developer time resolving conflicts
- Potential CI/CD pipeline issues

Screenshot 2025-06-08 at 13 38 48
Screenshot 2025-06-08 at 13 00 08

cc: @bucko13


.husky/_/pre-commit
```diff
@@ -0,0 +1,214 @@
...
+    # Count lines in unstaged changes
+    for UNSTAGED_LOCKFILE in $UNSTAGED_LOCKFILES; do
+        UNSTAGED_CHANGES=$(git diff --numstat "$UNSTAGED_LOCKFILE" | cut -f1)
+        if [ -n "$UNSTAGED_CHANGES" ] && [ "$UNSTAGED_CHANGES" -gt 50 ]; then
+            echo "   📊 Large unstaged changes in $UNSTAGED_LOCKFILE: $UNSTAGED_CHANGES lines"
+        fi
+    done
+
+    echo ""
+    printf "   Continue without staging these lockfiles? (y/N): "
```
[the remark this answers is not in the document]
Legend101Zz: ahh .. yes thanks for pointing it out changes [done](https://github.com/caravan-bitcoin/caravan/pull/307/commits/79b49bf5ffbc1f66cef91a8270caddb85b6b10c0)

.husky/_/pre-commit
```diff
@@ -0,0 +1,263 @@
...
+            echo ""
+            echo "⚠️  WARNING: Large changes detected in $LOCKFILE"
+            echo "   Lines changed: $LOCK_CHANGES"
+            echo "   But corresponding $PACKAGE_JSON is not staged"
+            echo ""
+            echo "   This might indicate:"
+            echo "   - Node.js version mismatch (current: $CURRENT_NODE)"
+            echo "   - Lockfile corruption"
+            echo "   - Missing package.json changes"
+            echo "   - Forgotten to stage package.json"
```
bucko13, line 102:
```suggestion
            echo "   - Forgot to stage package.json"
```

package.json
```diff
@@ -34,9 +39,6 @@
   "dependencies": {
     "@ledgerhq/hw-transport-webusb": "6.27.12"
   },
-  "engines": {
-    "node": ">=20"
```
bucko13, line 38: should we not keep engines as well? I think this has a lot of flexibility so we can add a minimum and a maximum. Actually wondering if that would be enough?

.npmrc
```diff
@@ -1 +1,8 @@
 registry=https://registry.npmjs.org/
+
+
+# Enforce Node.js version checking
+engine-strict=true
```
bucko13, line 5: this one too. if we're enforcing then shouldn't we leave the engine in the package.json's? and would this not be enough?

Legend101Zz: in the package.json as explained above "preinstall" script checks for that :)

bucko13: I'm still confused. Looking at that comment:

> That’s why I removed the "engines" field and engine-strict=true from the config 

but here you're adding it not removing it. I would've assumed that this enforces the engines check but then we've removed the field.

Legend101Zz (commented):
So ,earlier we had `${npmVer.split('.').slice(0,2).join('.')}` and this lead the shell ( in ci) to see `${...}` as a shell variable substitution, but it doesn't understand JavaScript methods like `.split()`, so the ci was failing

bucko13:
Just running an initial `npm i` I get this error:

```
> prepare
> husky install

husky - install command is DEPRECATED
```

.husky/_/pre-commit
```diff
@@ -0,0 +1,263 @@
...
+echo "🚀 Pre-commit checks (from $(pwd))..."
+
+# Find the git root directory (works from anywhere in monorepo)
+GIT_ROOT=$(git rev-parse --show-toplevel)
+echo "📁 Git root: $GIT_ROOT"
+
+# 1. Check Node.js version
+echo "1️⃣ Checking Node.js version..."
+CURRENT_NODE=$(node --version)
+EXPECTED_MAJOR="20"
```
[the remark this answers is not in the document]
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/307/commits/0c02ca2e33dfa4fc8ac6a06d65b94c5a30c949d8)

bucko13:
Looks good! Any idea why it appears to run 3 times (the unsupported engine, and the first one has no package name). I think moving the npm script into a dedicated file would cleanup the output as well so that the warnings and steps to fix are more clear. 

Screenshot 2025-06-13 at 3 47 43 PM

bucko13:
This is weird. After switching to the right node version and run `npm i` again, it seems to delete everything from husky.sh and pre-commit:

```diff
diff --git a/.husky/_/husky.sh b/.husky/_/husky.sh
index cec959a..f9d0637 100644
--- a/.husky/_/husky.sh
+++ b/.husky/_/husky.sh
@@ -1,36 +1,9 @@
+echo "husky - DEPRECATED
+
+Please remove the following two lines from $0:
+
 #!/usr/bin/env sh
-if [ -z "$husky_skip_init" ]; then
-  debug () {
-    if [ "$HUSKY_DEBUG" = "1" ]; then
-      echo "husky (debug) - $1"
-    fi
-  }
+. \"\$(dirname -- \"\$0\")/_/husky.sh\"

-  readonly hook_name="$(basename -- "$0")"
-  debug "starting $hook_name..."
-
-  if [ "$HUSKY" = "0" ]; then
-    debug "HUSKY env variable is set to 0, skipping hook"
-    exit 0
-  fi
-
-  if [ -f ~/.huskyrc ]; then
-    debug "sourcing ~/.huskyrc"
-    . ~/.huskyrc
-  fi
-
-  readonly husky_skip_init=1
-  export husky_skip_init
-  sh -e "$0" "$@"
-  exitCode="$?"
-
-  if [ $exitCode != 0 ]; then
-    echo "husky - $hook_name hook exited with code $exitCode (error)"
-  fi
-
-  if [ $exitCode = 127 ]; then
-    echo "husky - command not found in PATH=$PATH"
-  fi
-
-  exit $exitCode
-fi
+They WILL FAIL in v10.0.0
+"
\ No newline at end of file
```

.husky/pre-commit
```diff
@@ -0,0 +1,311 @@
...
+        }
+    ")
+
+    if [ $? -eq 0 ] && [ -n "$NODE_REQUIREMENT" ]; then
+        echo "   📝 Required Node.js: $NODE_REQUIREMENT"
+        echo "   💻 Current Node.js: $CURRENT_NODE"
+
+        # Use check-node-version to validate (same as preinstall script)
+        # First check if check-node-version is available
+        if ! command -v check-node-version >/dev/null 2>&1; then
```
bucko13, line 36: this will always fail the first time right since this is now a dependency which first has to be installed. Should we just do the non-`check-node-version` approach?

bucko13 (commented):
Worked as expected! Just one question that can maybe be disregarded

.husky/_/husky.sh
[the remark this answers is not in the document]
Legend101Zz: ---

Ahh… so let me digress a bit and take it from the top.

When we run `npx husky init` to set up Husky, it actually does several critical things behind the scenes:


### 1) First, it sets the Git hooks path

It automatically runs this command:

```bash
git config core.hooksPath .husky
```

Which basically tells Git:

> “Instead of looking in `.git/hooks/` for hook scripts, look in the `.husky/` directory.”

---

### 2) Then, it creates the structure we see

Stuff like `.husky/_/husky.sh`, `.husky/pre-commit`, etc. gets generated.
This sets up the whole hook system — including the core script that handles execution.

---

### 3) And now, whenever we run `git commit`:

* Git sees `core.hooksPath = .husky`
* It looks for `.husky/pre-commit`
* Executes the `pre-commit` script
* That script then calls `.husky/_/husky.sh`, which is basically the brain of Husky — it manages how hooks are run

---

So yeah, that’s why `.husky/_/husky.sh` is needed. I’m pretty sure if we delete it and then run `npm i husky` again, it’ll regenerate or restore it as part of the setup.

.husky/pre-commit
Legend101Zz: ... if you'll look closely we do that check ... :) 
 
```bash
        # First check if check-node-version is available
        if ! command -v check-node-version >/dev/null 2>&1; then
            echo "   ⚠️  check-node-version not found, using basic version check"
 ```
 
here we check that if check-node-version is not there then we use simple shell-based version checking

and if we have check-node-version ... then we use it ..

```bash
else
            # Use check-node-version for precise validation
            if ! check-node-version --node="$NODE_REQUIREMENT" >/dev/null 2>&1; then

Comment (bucko13) [reply .husky/pre-commit:36]: Yeah, i'm just wondering why we even need check-node-version dependency at all since we have to cover the situation where we don't anyway. Why not avoid the dependency (and the if/else) altogether?

Comment (Legend101Zz) [reply .husky/pre-commit]: ahh yess .. why add unnecessary complexity ...
fixed [here](https://github.com/caravan-bitcoin/caravan/pull/307/commits/2bcc28d52859b5fc3f81d58d7b239b38e8c5e524) :)

File: .husky/pre-commit [hunk 7]
```diff
@@ -0,0 +1,294 @@
...
+    echo "$STAGED_LOCKFILES" | sed 's/^/     /'
+
+    # Check each staged lockfile
+    for LOCKFILE in $STAGED_LOCKFILES; do
+        # Count lines changed in this lockfile
+        LOCK_CHANGES=$(git diff --cached --numstat "$LOCKFILE" | cut -f1)
+
+        # Find corresponding package.json
+        LOCKFILE_DIR=$(dirname "$LOCKFILE")
+        PACKAGE_JSON="$LOCKFILE_DIR/package.json"
```
bucko13, line 118: there should actually only be changes in one package-lock no matter which package.json changes.

bucko13 (changes requested):
Just two final things. Cleanup of the now unnecessary echo about the package and then my comment about the package-lock location. Perhaps it's not an issue but I'd like to confirm we're handling the monorepo/workspaces correctly (many package.json, one package-lock)

.husky/pre-commit
Legend101Zz: changes [done](https://github.com/caravan-bitcoin/caravan/pull/307/commits/43ca8541cbc088b1a6dd4c8cdc5548b436a6a731)

Legend101Zz:
ahh and @bucko13 one thing is left ... where should we add documentation for this ?

Legend101Zz:
> > ahh and @bucko13 one thing is left ... where should we add documentation for this ?
> 
> I think you can put it in the main readme. Thanks!

added [here](https://github.com/caravan-bitcoin/caravan/pull/307/commits/cc23ffb1bed5e3d6ed7301eec1c222003a90c2dd)

bucko13:
@coderabbitai any comments?

bucko13 (approved):

bucko13:
Going to merge, but I think this would be a nice addition in a quick followup:

> 3. Consider Adding .nvmrc
Since you're standardizing Node versions, adding .nvmrc would make version switching seamless for developers.

PR #319, opened 2025-06-27
 Dustchip | Dustchip file added on UI

ice-009:
This PR introduces a new DustChip component that classifies a UTXO amount as either:

Dust (too small to spend economically),

Warning (borderline wasteful), or

Economical

It uses WasteMetrics from @caravan/health to calculate dynamic dust thresholds based on:

The current feeRate

walletConfig (script type + quorum settings)

✅ Features
Supports custom tooltip via tooltipText prop

Falls back to default descriptions when no tooltip is provided

### DIFF
--- a/.changeset/old-boats-train.md
+++ b/.changeset/old-boats-train.md
@@ -1,0 +1,5 @@
+---
+"caravan-coordinator": patch
+---
+
+Introduced the DustChip for coordinator.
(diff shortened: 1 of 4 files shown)


ice-009:
@bucko13 the changeset has been added.. Thanks!!!

bucko13 (approved):

PR #322, opened 2025-06-29
Adds tanstack query and converts pending tx to new client

### LINKED ISSUE #321: Integrate TanStack Query
Long term, we'd like to move away from our dependence on redux. Using react's `context` where possible is a good first start. In addition, for managing network based queries and having a local cache we could start using [TanStack Query](https://tanstack.com/query/latest/docs/framework/react/examples/basic) instead of async redux action creators and the store. 

TanStack Query and redux can be used in parallel which should make it easy to migrate. To start, any _new_ requests that need to be cached and aren't already dependent on the redux store, we can simply use tanstack. For existing cacheing we can optionally have tanstack call dispatches to update the store when necessary (or do a full conversion anytime we want to move to the new system

bucko13:

* The only user facing change is removing the redundant refetch for tx tab
* introduces a TanStack query to caravan coordinator
* Adds some new queries for fetching transactions
* adds some new selectors to use with transaction analysis

**Issue Number:**

Fixes #321 

**Snapshots/Videos:**

https://github.com/user-attachments/assets/f962fcb6-f5c2-4ed4-9280-5971f816cd5a

N/A

**Summary**
When reviewing #257 I noticed a need to have pending transactions available to analyze at a more global level to avoid redundant network requests (and subsequent processing). 

My first instinct was to use redux, but we want to eventually deprecate it and in general it has a lot more overhead. So for _new_ stores it seemed like it would be nice to avoid redux at least. One option would be to use context and wrap around the wallet interface, but this would keep a lot of complexity still. 

TanStack seemed like the obvious choice. It could handle refetching, cacheing nicely, it also has some nice state management (isLoading, etc.) out of the box. So this PR introduces tanstack query (had to be v4 because of our older react), and moves the pending transaction functionality over to it. 

## Checklist

**Other information**

Yes


apps/coordinator/src/selectors/wallet.ts
```diff
@@ -0,0 +1,480 @@
...
+          name: string;
+        }): registration is HmacWithName => Boolean(registration.policyHmac),
+      );
+  },
+);
+
+/**
+ * @description Returns an array of all wallet addresses from both deposit and change nodes
+ */
+export const getWalletAddresses = createSelector(
```
bucko13, line 426: everything below this is new to the file. This turns some of the utility code from the old hooks file into selectors to get memoization out of the box without useMemo or useCallback

apps/coordinator/src/utils/transactionCalculations.ts
bucko13, line 1: these are all from the original hooks file that didn't qualify as a selector or part of the querying logic. when these are used in selectors and hooks though we should get the benefit of memoization since these won't be recalculated unless the underlying data has changed.

apps/coordinator/tsconfig.json
bucko13, line 1: adding some aliases to cleanup imports. for some reason i couldn't get special symbols to work with our eslint setup.

apps/coordinator/src/clients/transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+  return await client.getTransaction(txid);
+};
+
+// Hook for fetching all pending transactions
+const useFetchPendingTransactions = () => {
+  const pendingTransactionIds = useSelector(getPendingTransactionIds);
+  const blockchainClient = useGetClient();
+
+  return useQueries({
```
Legend101Zz, line 28: So as we did make the queryFn typed above then do you think it makes sense now to use `useSuspenseQuery` instead ( more [here](https://tanstack.com/query/latest/docs/framework/react/reference/useSuspenseQuery) ) 

What it does is always ensure that we get a return like `data: TData` while normal 
`useQuery` returns `data: TData | undefined` so  data is always guaranteed to be non-undefined.

apps/coordinator/src/clients/transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+    queries: pendingTransactionIds.map((txid) => ({
+      queryKey: transactionKeys.pendingTransaction(txid),
+      queryFn: () => fetchTransactionDetails(txid, blockchainClient),
+      enabled: !!blockchainClient && !!txid,
```
Legend101Zz, line 32: note : if you choose to use `useSuspenseQuery` then enabled is not supported in useSuspenseQuery.

apps/coordinator/src/selectors/wallet.ts
```diff
@@ -0,0 +1,480 @@
+import { createSelector } from "reselect";
+
+// Type definitions for the Redux state
```
Legend101Zz, line 3: ---

Ahh .. We really need to set up a centralized store or a `@caravan/types` package whose sole responsibility is to define and expose shared types across the project. Ideally, this package should also provide an API for devs to **register new types** and optionally **validate against existing ones**.

Right now, we’re ending up duplicating similar type definitions in multiple places — for example, I remember writing the same kind of types here:
https://github.com/caravan-bitcoin/caravan/blob/528e276e9d7841f95c1f569f699bc76ece2fdb2f/apps/coordinator/src/utils/psbtUtils.ts#L13-L37 

For now, I guess we’ll have to live with the duplication — but we should definitely track this as tech debt. 👀

---

apps/coordinator/src/clients/transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+// Hook for processed pending transactions with calculated values
+export const usePendingTransactions = () => {
+  const walletAddresses = useSelector(getWalletAddresses);
+  const transactionQueries = useFetchPendingTransactions();
+
+  // Calculate loading and error states
+  const isLoading = transactionQueries.some((query) => query.isLoading);
+  const error = transactionQueries.find((query) => query.error)?.error;
+
+  // Process transactions with calculated values and filter out confirmed ones
```
Legend101Zz, line 46: why do we need to filter out confirmed ones ?

Because you are calling the ` const pendingTransactionIds = useSelector(getPendingTransactionIds);` and as far as I can see the `getPendingTransactionIds` selector only [adds pending tx](https://github.com/caravan-bitcoin/caravan/pull/322/files#diff-29f893fcf8ebc6e0186189977a846ef908fd0478ca969692505ab3dc2604cf90R447)

apps/coordinator/src/selectors/wallet.ts
```diff
@@ -0,0 +1,480 @@
...
+export const getPendingTransactionIds = createSelector(
+  getWalletSlices,
+  (slices: Slice[]): string[] => {
+    const txids = new Set<string>();
+
+    slices.forEach((slice: Slice) => {
+      if (slice.utxos && slice.utxos.length > 0) {
+        slice.utxos.forEach((utxo: UTXO) => {
+          // Only include transaction IDs from unconfirmed UTXOs
+          if (utxo.txid && !utxo.confirmed) {
```
Legend101Zz, line 447: this is what I was mentioning above .. like you only add txid from unconfimed UTXO's

apps/coordinator/src/clients/transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+      };
+    });
+
+  return {
+    transactions,
+    isLoading,
+    error,
+    refetch: () => {
+      // Refetch all transaction queries
+      transactionQueries.forEach((query) => query.refetch());
```
Legend101Zz, line 67: this is really neat functionality we get out of box from tanQuery :)

apps/coordinator/tsconfig.json
Legend101Zz: So I guess this might be happening because we don’t have `eslint-import-resolver-alias` installed, and it looks like our shared ESLint config doesn’t properly handle `@` path aliases out of the box.

To fix this, we’ll likely need to:

* Install `eslint-import-resolver-alias`
* Override the `import/resolver` settings in the coordinator’s local `.eslintrc.cjs` to explicitly support the `@` alias

Without that override, ESLint can't resolve paths like `@/utils/foo` correctly, even though the code compiles fine via tsconfig.

bucko13: I tried that and unfortunately it didn't work :(. I didn't want to keep poking too much at it since I did have a workaround for now, but would like to get this working. If you can tweak it some more though I'd be more than happy to approve and merge!

apps/coordinator/src/selectors/wallet.ts
bucko13: 100% agreed. The sooner we can resolve this the better though as duplication becomes increasingly hard to manage.

apps/coordinator/src/clients/transactions.ts
bucko13: I didn't dig into this too much but without this then I was also getting confirmed showing up on the table.

apps/coordinator/tsconfig.json
Legend101Zz: great understood :)

bucko13:
> TLDR; TanStack Query is great—but unless we plan the rest of the ecosystem around it, we risk turning a good decision into long-term chaos. Would love to open up a discussion at some point on setting some state management principles or even drafting a lightweight migration plan.

great feedback! I totally agree about the challenge here. I outlined some basics around how we can migrate in the issue #322. Based on your feedback I also added to the Contributor README about how to manage this migration as well and when to use tanstack instead of redux or context. 

But for the time being a lot of the responsibility will be on us as reviewers to hold ourselves and contributors accountable.

PR #324, opened 2025-06-30
Jade wallet integration

Austin-Fulbright:

This PR introduces Jade as a wallet option for Caravan.

I have not updated the documentation yet.

**Summary**

This integration is similar to the other wallet integrations into Caravan. I have added the jade.ts file as a direct keystore interaction wallet, and updated the caravan-wallet index.ts to support it. I have also added the test suite for this wallet. One caveat about the **test suite** is that you will have to either choose **testnet** or **mainnet** when initializing the wallet with the test seed phrase. If there are any other changes that should be added, let me know. I might need to update the PR for signing messages for testnet. As the wallet only works for signing messages on **mainnet** right now. 

No breaking changes, hopefully this PR can integrate smoothly into the existing environment.

## Checklist
- [*] I have tested my changes thoroughly.
- [*] I have added or updated tests to cover my changes (if applicable).
- [*] I have verified that test coverage meets or exceeds 95% (if applicable).
- [*] I have run the test suite locally, and all tests pass.
- [*] I have written tests for all new changes/features

**Other information**

Yes


Legend101Zz:
@Austin-Fulbright thanks a ton ... super excited for this now I can finally use my jade to test things out :)

jbrauck-unchained:
@Austin-Fulbright thanks for the PR! Excited to try this out!
Right now I only have a regular Jade. Do regular Jades have usb connection support? Or is this PR only for Jade Plus?

jbrauck-unchained:
Screenshot 2025-06-30 at 1 37 00 PM
Screenshot 2025-06-30 at 1 35 23 PM
A couple of odd things. I see three serial ports as options when connecting the Jade. I chose the USB Single Serial one and got the following error when trying to begin the test. Same error on xpub export operation.

Austin-Fulbright:
> Screenshot 2025-06-30 at 1 37 00 PM Screenshot 2025-06-30 at 1 35 23 PM A couple of odd things. I see three serial ports as options when connecting the Jade. I chose the USB Single Serial one and got the following error when trying to begin the test. Same error on xpub export operation.

does this work on mainnet? or is the wallet initialized as a testnet wallet?

bitcoinhelp:
Hey @jbrauck-unchained - Can I ask what the steps you took on the device to get here? We usually see this occur when users select QR Mode/Scan SeedQR and then try to connect with USB.

QR Mode disables the USB port. If you want to load a seed to Jade temporarily and also connect to USB, you'll have to go to Options -> Temporary Signer -> then enter seed or scan QR -> then pick USB

package-lock.json
```diff
@@ -14654,15 +14732,28 @@
       }
     },
     "node_modules/get-intrinsic": {
+<<<<<<< HEAD
```
Legend101Zz, line 14735: ---

Hey @Austin-Fulbright — I’ll aim to give a full initial review soon and then we can have others also review it , but one thing to flag early: it looks like there’s an unusually large diff in the `package-lock.json`. This might be due to a mismatch in your local Node/npm versions when you initially created the PR.

We've seen similar issues before see #306 , which were caused by different lockfile versions being generated depending on the Node environment. We've since added pre-install and pre-commit checks to help enforce consistent versions—but it looks like your branch might have been created before those safeguards were added.

Austin-Fulbright: @Legend101Zz Ahh Gotchya. Would doing the pre-install and pre-commit checks or switching to the correct environment fix this issue? I had a feeling there was going to be some version mismatches, I was getting weird errors regarding version incompatibilities.

Austin-Fulbright: is it possible to just remove all generated files eg. package-lock.json and switch to the correct to versions of node v20 and then npm v8 then re install all the packages to generate the correct package-lock.json.

Legend101Zz: > @Legend101Zz Ahh Gotchya. Would doing the pre-install and pre-commit checks or switching to the correct environment fix this issue? I had a feeling there was going to be some version mismatches, I was getting weird errors regarding version incompatibilities.

---

Ahh no — the pre-install and pre-commit hooks only **prevent mistakes**, they don’t actually **fix** anything 😅

What you should do instead is:

* **Reset the `package-lock.json`** to match the one from the `main` branch.
* Make sure you're using the **correct Node.js version** (as specified in `.nvmrc` or `package.json`).
* Now that the pre-install hooks are synced in your branch, they’ll **catch any future mismatches** and stop you from installing with the wrong Node version.
* Since your `package.json` changes are intact, simply reinstalling `npm i` should  clean up the big diff in `package-lock.json` ... :)

Legend101Zz: > is it possible to just remove all generated files eg. package-lock.json and switch to the correct to versions of node v20 and then npm v8 then re install all the packages to generate the correct package-lock.json.

yup pretty much this is what you'll need to do

Austin-Fulbright: Let me know if my latest change fixed the issue.

Austin-Fulbright: Hmm switching to these versions broke the view:
npm --version
10.8.2  
node --version
v20.19.3

Austin-Fulbright: Screenshot 2025-07-05 at 11 57 12 AM


this seems to be an error that keeps coming up and I know it has something to do with my version mismatch.

Legend101Zz: > It says to always use v8 of npm but in main you are using v10.5.0. does that mean that I should be using npm@10.5.0 instead of 8?

hmm ... I thought we defined it to use v10.5 or greater, can you please point out where it says to use  v8 ??

Legend101Zz: > Let me know if my latest change fixed the issue.

I guess not :( 
We still have 40k lines of diff 😭
Screenshot 2025-07-06 at 13 01 29

package.json
```diff
@@ -52,7 +52,7 @@
       "@ledgerhq/devices": "8.0.0"
     }
   },
-  "packageManager": "npm@10.5.0",
+  "packageManager": "npm@10.8.2",
```
Legend101Zz, line 55: Hmm curious — why was this version changed?
If you check the `main` branch here:

https://github.com/caravan-bitcoin/caravan/blob/969dc3792abff56be898452a97c1c467c3bdf34c/package.json#L54-L57

we're still using version `10.5`. Updating this might have caused the build issues you ran into. Could you double-check if reverting it fixes things?

package-lock.json
Legend101Zz: please tell me know if this worked ?

https://github.com/caravan-bitcoin/caravan/pull/324#discussion_r2188085503

.changeset/swift-seals-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@caravan/wallets": minor
```
Legend101Zz, line 2: you'll also need to add a changeset for coordinator :)

packages/caravan-wallets/package.json
```diff
@@ -88,6 +88,7 @@
     "bitcoinjs-lib": "^5.1.10",
     "bowser": "^2.6.1",
     "core-js": "^2.6.10",
+    "jadets": "^1.1.18",
```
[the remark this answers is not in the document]
Austin-Fulbright: Yeah that worked! thank you so much!

Legend101Zz (approved):
The package-lock looks good now — thanks a lot, @Austin-Fulbright !

@bucko13 whenever you get a chance, can you please review this.

bucko13:
Thank you for this @Austin-Fulbright and thank you everyone for the reviews so far. Definitely want to get to this ASAP. I'm trying to work through our existing backlog of features and will get to this soon.

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,504 @@
...
+    !baseDerivation.every((v, i) => v === fullPath[i])
+  ) {
+    throw new Error(
+      `Path "${fullPathStr}" does not extend base derivation [${baseDerivation.join(",")}]`,
+    );
+  }
+  return fullPath.slice(baseDerivation.length);
+}
+
+function walletConfigToDescriptor(
```
bucko13, line 100: Can we call this something like `walletConfigToJadeDescriptor`? it's not quite a descriptor but actually something specific to jade it looks like. For string encoded descriptors we do have a library that can actually encode from a config to a string descriptor (see [@caravan/descriptors](https://www.npmjs.com/package/@caravan/descriptors))

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,504 @@
...
+
+  async withDevice<T>(
+    network: string,
+    f: (jade: IJade) => Promise<T>,
+  ): Promise<T> {
+    try {
+      await this.jade.connect();
+
+      const httpRequestFn = async (params: any): Promise<{ body: any }> => {
+        const url = params.urls[0];
```
bucko13, line 160: curious where this url gets sent from. is this authentication required for jade interactions? it seems somewhat concerning that there is some client-source url that is passed to an authentication endpoint call here. Are there any man in the middle risks we should be worried about or protecting against?

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,504 @@
...
+    return await this.withDevice(this.network, async (jade: IJade) => {
+      const descriptor = walletConfigToDescriptor(this.walletConfig);
+      let multisigName = await jade.getMultiSigName(this.network, descriptor);
+
+      if (!multisigName) {
+        multisigName = `jade${randomBytes(4).toString("hex")}`;
+        await jade.registerMultisig(this.network, multisigName, descriptor);
+      }
+      const wallet = await jade.getRegisteredMultisig(multisigName);
+      console.log("registered wallet", wallet);
```
bucko13, line 360: cleanup

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,504 @@
...
+        if (this.returnSignatureArray) {
+          const b64string = bytesToBase64(signedPSBT);
+          const parsedPsbt = parsePsbt(b64string);
+          const fingerprint = await jade.getMasterFingerPrint(
+            this.walletConfig.network,
+          );
+
+          let sigArray: string[] = [];
+
+          for (let i = 0; i < parsedPsbt.PSBT_GLOBAL_INPUT_COUNT; i++) {
```
bucko13, line 433: totally aesthetic comment, but this might be a nice utility to extract into a utility function at the top? keeps the interaction function more straightforward.

bucko13 (changes requested):
So sorry for the delay! Really excited to get this in. I started reviewing. Only touched the caravan-wallets changes to start. The interactions look great. left some feedback to start with. Will get to the caravan-coordinator code next, but wanted to provide something to start responding to :)

packages/caravan-wallets/src/jade.ts
bucko13, line 1: This is non-blocking for me since I neglected to comment on this for the bitbox integration, but I do think it would be nice to add some unit tests. You can look at the bcur2 code for the best examples of mocking external dependencies and testing that they're called. Ledger has some decent ones as well though it's mixed in with some legacy code and tests that are not as legible, so bcur2 is probably the best bet (uses a nice new utility `vitest-mock-extended`)

bucko13 (changes requested):
Couple final comments to complete the review. Overall looks great! I'll try and get setup to test fully but well done and thank you for putting this all together!

bucko13:
Got some linting errors as well: https://github.com/caravan-bitcoin/caravan/actions/runs/16734060741/job/47369921807?pr=324

Austin-Fulbright:
I would love to add unit tests. I think I need to understand how the Ledger.test.ts file works. does it mock the entire device or do you still need to have a test device connected with a certain configuration. I see some of these test do mock device interaction, but it looks like get metadata is talking to a device. Could you clear this up for me?

bucko13:
> @bucko13 I would love to add unit tests. I think I need to understand how the Ledger.test.ts file works. does it mock the entire device or do you still need to have a test device connected with a certain configuration. I see some of these test do mock device interaction, but it looks like get metadata is talking to a device. Could you clear this up for me?

Great! yeah, I think for unit tests all device interactions can be mocked. It would be great to eventually work in device tests with emulators in our e2es, but for now for unit tests just mocking the module is what we'd expect. The ledger tests do do that as well, but it might be a little disorganized. The bcur2 ones are the newest ones that were written from scratch using vitest so are probably the cleanest.

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,504 @@
...
+    const index = parseInt(numStr, 10);
+    if (isNaN(index)) {
+      throw new Error(`Invalid path segment: ${segment}`);
+    }
+    result.push(index + (hardened ? 0x80000000 : 0));
+  }
+  return result;
+}
+
+function extractPathSuffix(
```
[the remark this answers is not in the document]
Austin-Fulbright: Ok I see. So this is specific to Jade. The solution could be to turn them both the parent and child path into strings use the `getRelativeBIP32Path` function and then use the bip32PathToSequence function to turn the path back into an int array. It will add some complexity to the code but it might be worth it for consistency.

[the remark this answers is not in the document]
bucko13: right! exactly. There's also `bip32PathToSequence` which takes the string and converts to int array and also `bip32PathToSequence` that does the reverse.

bucko13:
I'm getting stuck trying to test with a jade device. Total transparency, I haven't used a jade before and just had an old one laying around that I got from a conference a few years ago. That said, if I'm having issues I imagine we will need to provide users with a little bit more feedback in the UI to make this more user friendly. 

Some questions:
1. does the blockstream desktop app need to be opened to unlock with pin? 
2. if so, does that conflict caravan then trying to access the device at the same serial port? 
3. If I don't open the desktop app then my device says I need to unlock. I click that and see "connect via USB/BLE to a companion app and select your Jade to unlcok with your PIN". then I open the app, unlock and cant do any of the interactions in the test suite or wallet (see screenshot) 

It's entirely possible I'm using the device incorrectly or set it up wrong. But it's not entirely clear to me where the mistake was, so we should at least prepare to add some clearer instructions if necessary

Screenshot 2025-08-06 at 5 35 54 PM

packages/caravan-wallets/src/jade.ts
bucko13: bump. I get this might be required, but would like to understand this a little more. Maybe we could at least add some inline comments? I actually wonder if this is related to my issues trying to test out this MR? If I don't unlock with the app then this is what happens when trying to interact with the device. it just hangs here. If I do unlock with the app then I get the serial port conflict. 

Screenshot 2025-08-06 at 5 42 36 PM

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,467 @@
...
+    this.includeXFP = includeXFP;
+  }
+
+  messages() {
+    return super.messages();
+  }
+
+  async run() {
+    return await this.withDevice(this.network, async (jade: IJade) => {
+      const path = bip32PathToSequence(this.bip32Path);
```
bucko13, line 254: 🎉

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,467 @@
...
+      let multisigName = await jade.getMultiSigName(this.network, descriptor);
+
+      if (!multisigName) {
+        multisigName = `jade${randomBytes(4).toString("hex")}`;
+        await jade.registerMultisig(this.network, multisigName, descriptor);
+      }
+      const paths = descriptor.signers.map((signer) => {
+		  let childPath = bip32SequenceToPath(signer.derivation);
+		  return getRelativeBip32Sequence(childPath, this.bip32Path);
+
```
bucko13, line 370: extra line can be removed here.

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,467 @@
...
+      this.walletConfig.network,
+      async (jade: IJade) => {
+        const signedPSBT = await jade.signPSBT(
+          this.walletConfig.network,
+          this.unsignedPsbt,
+        );
+
+        if (this.returnSignatureArray) {
+          const b64string = bytesToBase64(signedPSBT);
+          const parsedPsbt = parsePsbt(b64string);
```
bucko13, line 432: does jade not always return the same version PSBT?

bucko13 (commented):
this is looking great! Just a couple minor things on the code. I'd like to verify the interactions in the wallet and test suite as well before merging. If this is just user error then perhaps I can get some validation from @jbrauck-unchained but would like to resolve what the issue is if possible.

bucko13:
Awesome! Passed all but one of the test suite (the failure looks like a fixture issue).

Screenshot 2025-08-07 at 4 14 34 PM

Got some issues when testing though:

Looks like some copy-pasta here with the menu item doubling up "Trezor" rather than Jade. 
Screenshot 2025-08-07 at 4 16 05 PM

I got this error when testing with a regtest wallet. It looks like jade should support regtest though. Any ideas? 
Screenshot 2025-08-07 at 4 17 57 PM
Screenshot 2025-08-07 at 4 18 55 PM
Screenshot 2025-08-07 at 4 19 16 PM

bucko13 (changes requested):
Almost there! Couple last nits to fix:

- [ ] Code comment about the fetch that needs to be made
- [ ] the duplicate "trezor" in the menu
- [ ] unit tests (happy to split this up into a quick follow on, but preference is to include some tests here, fully mocking out comms with the device) 
- [ ] Regtest support

packages/caravan-wallets/src/jade.ts
```diff
@@ -0,0 +1,468 @@
...
+  let connected = false;
+
+  try {
+    await this.jade.connect();
+    connected = true;
+
+    // Perform user handshake
+    const httpRequestFn = async (params: any): Promise<{ body: any }> => {
+      const url = params.urls[0];
+      const res = await fetch(url, {
```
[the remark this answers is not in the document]
bitcoinhelp: @bucko13 in regards to the regtest comment, perhaps the following is useful?

Once Jade is initialized with PIN, it is locked to the network type it was set up with. So if it was initialized with mainnet, it will only work on mainnet when unlocking with PIN.

To test with regtest, you may need to factory reset the Jade, or use the Options -> Temporary Signer feature, in which case Jade will ignore the PIN persisted wallet settings and will work with testnet/regtest during that login

bucko13:
@bitcoinhelp going to reply to the network comment here.

> To test with regtest, you may need to factory reset the Jade, or use the Options -> Temporary Signer feature, in which case Jade will ignore the PIN persisted wallet settings and will work with testnet/regtest during that login

Yep, I did that. That's how I was able to pass all the tests in the test suite (which have both mainnet and testnet). The error pictured in the screenshots above is different from the one that comes up when the device is loaded with a different network. Note the error "Failed to extract valid network from parameters".

Austin-Fulbright:
@bucko13 So on the jade `regtest` is actually named `localtest` what is the cleanest way that I can solve this edgecase for the network you think. I could run a check for the network in every jade interaction but that seems a little redundant, is there a better way you can think of?

packages/caravan-wallets/src/jade.ts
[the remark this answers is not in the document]
Austin-Fulbright: @bucko13 The url it is fetching is from the jade device itself you can read https://github.com/Blockstream/Jade/blob/master/docs/index.rst#auth_user-request
for more information on how it works I will add a comment in the code as well.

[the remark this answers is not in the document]
bucko13: Great! Thanks. A comment would be great.

bucko13:
@Austin-Fulbright I got this comment but can't find it in the PR so just reproducing here:

> So on the jade regtest is actually named localtest what is the cleanest way that I can solve this edgecase for the network you think. I could run a check for the network in every jade interaction but that seems a little redundant, is there a better way you can think of?

Not sure if you figured it out already and that's why the comment disappeared but yeah i think converting it probably makes sense. We have to do this for bitcoinjs-lib as well since their "mainnet" is "bitcoin" so there are some functions around the codebase that convert network labels. 

The nice thing about the OOP pattern used in the wallet interactions is that since you need to do this for every interaction, you should just be able to put this on the base class and use that for everything else (i.e. do something like convert the network in the constructor and set `this.network` to the result). Hopefully that helps provide a DRYer approach.

Austin-Fulbright:
@bucko13 Yeah I wasn't sure if it was correct. But I think it is localtest. I will make sure and then add it to the constructor. The base class being this one correct:

```
export class JadeInteraction extends DirectKeystoreInteraction {
  protected transport: JadeTransport;

  protected ijade: IJadeInterface;

  protected jade: IJade;

  constructor() {
    super();
    this.transport = new SerialTransport({});
    this.ijade = new JadeInterface(this.transport);
    this.jade = new Jade(this.ijade);
  }


```

Austin-Fulbright:
Ok I think I found a better solution than I just implemented and it solves some of my other problems I have been having with passing a network through the withDevice function.

Austin-Fulbright:
Ok I added the network as an optional parameter to the base class, then when the child class gets the network they can pass it to the parent constructor through the super() function. this fixes the awkward network passing to the withDevice function, it also converts the regtest network to localtest if it is detected.

Austin-Fulbright:
I will also try and squash these commits to make it easier to review.

bucko13:
Hmmm, still getting an error when trying to export an xpub in regtest:

```
Error getting fingerprint: TypeError: Invalid network version
```

Screenshot 2025-08-08 at 4 44 03 PM

bucko13:
For confirm address I was able to go through the steps on the device, but still get an error (this didn't throw anything in the console though) 

Screenshot 2025-08-08 at 4 45 18 PM

Austin-Fulbright:
Yeah the problem is my code is still not converting "regtest" to "localtest" I am trying to fix it now. I will just keep testing it until it works.

bucko13:
Almost there! signing and xpub export work. For some reason I can't get the address verification to work for the actual wallet interactions (works in the test suite though). 

Tested (regtest):

- [x] xpub export
- [ ] address verification (P2SH)
- [ ] address verification (P2WSH)
- [x] signing (P2SH)
- [x] signing (P2WSH)

Address verification failed (P2SH)
Screenshot 2025-08-11 at 10 48 32 AM

Similar for p2wsh (thought it did walk me through wallet registration it looked like)

Screenshot 2025-08-11 at 10 51 56 AM

Note it works for the test suite with mainnet
Screenshot 2025-08-11 at 10 49 57 AM

Austin-Fulbright:
@bucko13 Just for clarification. It does work on testnet. It is just regtest that is giving these errors. correct?

bucko13:
> @bucko13 Just for clarification. It does work on testnet. It is just regtest that is giving these errors. correct?

correct

bucko13:
> Yeah I am resolving the conflicts right now. I was having a little trouble replicating the error you have. I think my node versions are causing problems in the application. I keep getting some error about unexpected context.

Weird. @Legend101Zz actually built a pre-commit hook that tries to catch mismatch node versions now so I wouldnt' expect that to be an issue. @Legend101Zz maybe you could help debug the context issues?

Austin-Fulbright:
@Legend101Zz So it happens in two places. the first one is here when trying to expand details on an address that was created by a multisig wallet. 
errors-when-clicking-description

The second is here when trying to create an address with public keys after clicking import on the 3rd address I get the same errors.

Screenshot 2025-08-25 at 3 09 27 PM

This is on testnet. The testsuite still passes for everything on testnet and mainet, but when trying to go through the actual flow I get this react errors. Perhaps I implemented some of the actions incorrectly.

Austin-Fulbright:
@Legend101Zz I have tried tinkering with the useEffect() and interaction stuff I still can't get it to show me what the actual error is without giving me the infinite Unexpected fiber popped error. I am not super knowledgeable about react renderers. But would this error have something to do with my actual interaction with jade? I am guessing this error would have something to do with the way I implemented the UI components since it passes all of the unit and suite tests. I will try and read the discord messages to see if I can stop the infinite render problem in the meantime. 

@bucko13 I added the dependency injection for the jade interface, transport interface, ect. I tried to add fetch as a dependency in the constructor. but for whatever reason when the fetch is declared in the constructor it loses its context. I can still mock fetch in the test suite using global fetch. Also I added the type safe mocking. Let me know what might need to be changed in the test. I am also going to make the http function a little more type safe but I will need to go into my jadets package and change that.

Legend101Zz:
>  I will try and read the discord messages to see if I can stop the infinite render problem in the meantime.

Let me clone your PR and check if I can be of some help :) 
These react pop errors are insane , I also suffered a lot due to them too :( 
the current commit still has this issue right ??
> the first one is here when trying to expand details on an address that was created by a multisig wallet.

Will I also be able to expect the same behaviour on regtest ??

Austin-Fulbright:
@Legend101Zz Yes you will. Thank you so much.

Legend101Zz:
> @Legend101Zz Yes you will. Thank you so much.

@Austin-Fulbright It’s been a hectic weekend, so just bear with me for a couple more days :) 

Right now I’m digging into network calls and adding debug points in DevTools to trace the root cause of the error.

My goal is to figure out whether the bug is:

* **React-specific** → in which case we’re stuck, since fixing it would require bumping versions (which isn’t possible without a major Coordinator refactor). If that’s the case, I’ll try to patch things just for your PR so it can move forward and - and maybe even say a few nice things about React while I’m at it 🙂

* **Something in our code** → which would actually be better, since we could fix it directly. In that case, I’ll propose a fix for this PR and also open a separate PR to resolve the issue more broadly  :) 

---

Austin-Fulbright:
> > @Legend101Zz Yes you will. Thank you so much.
> 
> @Austin-Fulbright It’s been a hectic weekend, so just bear with me for a couple more days :)
> 
> Right now I’m digging into network calls and adding debug points in DevTools to trace the root cause of the error.
> 
> My goal is to figure out whether the bug is:
> 
> * **React-specific** → in which case we’re stuck, since fixing it would require bumping versions (which isn’t possible without a major Coordinator refactor). If that’s the case, I’ll try to patch things just for your PR so it can move forward and - and maybe even say a few nice things about React while I’m at it 🙂
> * **Something in our code** → which would actually be better, since we could fix it directly. In that case, I’ll propose a fix for this PR and also open a separate PR to resolve the issue more broadly  :)

Sounds great thanks for working with me on this. Yeah I have said some nice things about React before too hahaha

Legend101Zz:
@Austin-Fulbright ok so this is not something your code did ... but something that an earlier BCUR integration PR did ... as I just cloned a fresh repo and I see the same bug .. what is nice is if you run the code in firefox or safari you'll see the actual error that is there :

Screenshot_2025-09-02_at_00 05 54


so maybe if not needed you can ignore this bug for now ( like if it does not hinder your development) and I'll submit a patch PR in the meantime .. thanks for catching this ... this got leaked into prod and surprisingly our `ci` did not catch it :(

Legend101Zz:
> @Austin-Fulbright ok so this is not something your code did ... but something that an earlier BCUR integration PR did ... as I just cloned a fresh-latest repo and I see the same bug .. what is nice is if you run the code in firefox or safari you'll see the actual error that is there :
> 
> Screenshot_2025-09-02_at_00 05 54
> so maybe if not needed you can ignore this bug for now ( like if it does not hinder your development) and I'll submit a patch PR in the meantime .. thanks for catching this ... this got leaked into prod and surprisingly our `ci` did not catch it :(

What I found that this only happens in **Chrome** because V8 is stricter about context stack depth during React’s reconciliation. When a `detailPanel` in **Material-Table** nests complex Material UI components (like `Tabs`), it ends up causing a stack overflow. In **Firefox** and **Safari**, their JS engines handle deep context trees more gracefully, so the crash doesn’t show up there.

bucko13:
So, I'm still getting the same issue on regtest unfortunately. I don't have a ready way to check on mainnet or testnet, however testnet did work in the test suite. So this is either network specific or implementation specific. Unfortunately the error I get doesn't give much information in the UI and there's nothing in the logs.

Screenshot 2025-09-15 at 2 01 21 PM
Screenshot 2025-09-15 at 2 02 59 PM

bucko13:
@Austin-Fulbright reproduced and resolved the problem locally. This diff should solve:

```diff
diff --git a/apps/coordinator/src/components/Slices/ConfirmAddress.jsx b/apps/coordinator/src/components/Slices/ConfirmAddress.jsx
index 00f6abf..fab15bb 100644
--- a/apps/coordinator/src/components/Slices/ConfirmAddress.jsx
+++ b/apps/coordinator/src/components/Slices/ConfirmAddress.jsx
@@ -201,7 +201,10 @@ const ConfirmAddress = ({ slice, network }) => {
 
     try {
       let confirmed = await interaction.run();
-      if (typeof confirmed === "string" && state.deviceType === LEDGER) {
+      if (
+        (typeof confirmed === "string" && state.deviceType === LEDGER) ||
+        state.deviceType === JADE
+      ) {
         confirmed = {
           address: confirmed,
           serializedPath: interaction.bip32Path,
```

Austin-Fulbright:
Sorry for not responding earlier I have been sick for a few days. I will add these changes tomorrow morning and merge. Thank you so much for finding that error.

Austin-Fulbright:
Ok I added the two changes. I checked the type and it always comes out as PsbtV2 so I just switched the any in the get signature array to take in the PsbtV2 as its type. It works on all the test suites and the unit tests so it should be good to go.

bucko13:
Success!
Screenshot 2025-09-16 at 8 18 35 PM

bucko13 (approved):
👏 🎉

Legend101Zz:
Thanks for the amazing work @Austin-Fulbright 🥳

Austin-Fulbright:
@Legend101Zz Of course! @bucko13 Thanks for helping me finish this

PR #329, opened 2025-07-06
add back all the changes to client needed for RBF PR

Legend101Zz:
Small patch PR to add back all the changes to client needed for RBF PR


bucko13:
Just needs a changeset and we should be good

bucko13 (approved):

Issue #333, opened 2025-07-08
Add Full Transaction History to Caravan


Legend101Zz:
### **Improving Transaction History in Caravan**

**Context**
When we first added the transaction history tab in [#168](https://github.com/unchained-capital/caravan/pull/168), the goal was to show both pending and completed transactions. 

But there’s a fundamental limitation: Caravan is a stateless wallet, and only keeps track of UTXOs that currently belong to the wallet. Once those UTXOs are spent, they are removed from `slice.utxos`, which makes it impossible to infer or reconstruct the original transaction from in-memory state alone.

---

### **What We Noticed**

Despite not keeping spent UTXOs, we **do** track which addresses have been used. For example, in the UI you can see spent slices still present with their associated address metadata. 

<img width="1116" height="620" alt="Image" src="https://github.com/user-attachments/assets/6e148acd-a327-4b2a-ba8c-a73861ea67a3" />

This is possible because we maintain an `addressUsed` flag even after the balance becomes zero.

We already have a selector for this:

```ts
/**
 * Returns all spent slices: addresses that have been used but have no remaining balance.
 */
export const getSpentSlices = createSelector(
  getSlicesWithLastUsed,
  (slices: SliceWithLastUsed[]): SliceWithLastUsed[] =>
    slices.filter(
      (slice: SliceWithLastUsed) =>
        slice.addressUsed && slice.balanceSats.isEqualTo(0),
    ),
);
```

This means we still have access to **addresses** that were part of past transactions — even if their UTXOs are no longer active.

<img width="601" height="323" alt="Image" src="https://github.com/user-attachments/assets/5f64269a-8e73-4ae5-a37c-a4293e837f2b" />

---

### **Proposed Approach**

We can reconstruct full transaction history by **querying the blockchain using these addresses**, rather than relying on current UTXOs.

**For public clients**, we can use APIs like:

* [GET /address/:address/txs](https://mempool.space/docs/api/rest#get-address-transactions)
* [GET /address/:address/txs/mempool](https://mempool.space/docs/api/rest#get-address-transactions-in-mempool)
* These endpoints exist on both Mempool.space and Blockstream’s Esplora.
* Pagination is supported (`?limit`, `?offset`, or `last_seen_txid` depending on the API).

**For private clients using Bitcoin Core**, it’s more involved:

* We can use [listtransactions](https://developer.bitcoin.org/reference/rpc/listtransactions.html) to return wallet-related transactions.
[...]

PR #340, opened 2025-07-11
Add Full Transaction History Support in Caravan-Client Package

ice-009:
Referencing the [ISSUE_333](https://github.com/caravan-bitcoin/caravan/issues/333)

**Issue Number:**

Fixes #

**Snapshots/Videos:**

**Summary**

## Checklist

**Other information**


packages/caravan-clients/src/client.test.ts
```diff
@@ -414,6 +414,252 @@ describe("BlockchainClient", () => {
...
+      await expect(client.listTransactions()).rejects.toThrow(
+        "listTransactions is only supported for private clients"
+      );
+    });
+
+    it("should throw error for unsupported network (REGTEST)", async () => {
+      expect(() => {
+        new BlockchainClient({
+          type: ClientType.PUBLIC,
+          provider: PublicBitcoinProvider.BLOCKSTREAM,
```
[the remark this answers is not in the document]
ice-009: done the changes in the updated commit

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,50 @@ export class BlockchainClient extends ClientBase {
...
+
+      if (response?.result) {
+        return response.result;
+      }
+
+      throw new Error("Failed to retrieve transactions list");
+    }
+    
+    // Public clients not supported
+    throw new Error("listTransactions is only supported for private clients");
```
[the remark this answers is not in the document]
ice-009: sorry solved this too :)

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,75 @@ export class BlockchainClient extends ClientBase {
...
+        basePath = "";
+        break;
+      case Network.TESTNET:
+        basePath = "/testnet";
+        break;
+      case Network.SIGNET:
+        basePath = "/signet";
+        break;
+      case Network.REGTEST:
+        basePath = "/testnet"; // Fallback for REGTEST
```
bucko13, line 696: we should probably just skip this or throw an error that regtest isn't supported for mempool/blockstream

ice-009: [corrected here ](https://github.com/caravan-bitcoin/caravan/pull/340/commits/9b0ed7fbda19bf20f0b550fe1fa0544946337402)

Legend101Zz (changes requested):
Left some more comments

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,72 @@ export class BlockchainClient extends ClientBase {
...
+        break;
+      case Network.SIGNET:
+        basePath = "/signet";
+        break;    
+      default:
+        throw new Error("Unsupported network for public transactions");
+    }
+
+    const result = await this.Get(`${basePath}/api/address/${address}/txs?count=${count}&skip=${skip}`);
+    return result;
```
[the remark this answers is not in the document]
Legend101Zz: ahh also you'll need a changeset of patch

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,72 @@ export class BlockchainClient extends ClientBase {
...
+
+      if (response?.result) {
+        return response.result;
+      }
+
+      throw new Error("Failed to retrieve transactions list");
+    }
+    
+    // Public client implementation
+    if (this.provider !== PublicBitcoinProvider.MEMPOOL) {
```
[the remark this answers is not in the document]
ice-009: added the blockstream too in the function

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+        throw new Error("Failed to retrieve transactions from Bitcoin Core");
+      }
+
+      // Filter transactions for the specific address
+      const addressTransactions = response.result.filter(
+        (tx: any) => tx.address === address
+      ).slice(0, count); // Limit to requested count
+
+      // Get detailed info and normalize each transaction
+      const detailedTransactions = await Promise.all(
```
Legend101Zz, line 679: ---

**"Just to clarify — why exactly was the extra `gettransaction` call necessary? Doesn't the `listtransactions` method already return all the data we need? Based on the [Bitcoin RPC docs](https://developer.bitcoin.org/reference/rpc/listtransactions.html?highlight=listtransactions), it seems like `listtransactions` provides most of the relevant information. So I'm wondering in what specific cases we actually *need* to call `gettransaction` in addition to it."**

also if we do need that we do this in coordinator ... where we get all the txid first using the first method and then call `gettransaction` in coordinator

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+        method: "listtransactions",
+        params: ["*", count * 5 + skip, skip, includeWatchOnly], // Get extra to account for filtering
+      });
+
+      if (!response?.result || !Array.isArray(response.result)) {
+        throw new Error("Failed to retrieve transactions from Bitcoin Core");
+      }
+
+      // Filter transactions for the specific address
+      const addressTransactions = response.result.filter(
```
Legend101Zz, line 674: why was this needed .. we are already sure that the address we asked for the above method returns us tx related to that address only ?

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+        throw new Error("Wallet name is required for private client transaction listings");
+      }
+
+      // Get all wallet transactions and filter by address
+      const response = await callBitcoindWallet({
+        baseUrl: this.bitcoindParams.url,
+        walletName: this.bitcoindParams.walletName,
+        auth: this.bitcoindParams.auth,
+        method: "listtransactions",
+        params: ["*", count * 5 + skip, skip, includeWatchOnly], // Get extra to account for filtering
```
Legend101Zz, line 666: I guess the extra is not needed , as we don't need the filtering ?

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+
+    // Public client (Mempool/Blockstream)
+    if (!this.provider) {
+      throw new Error("Provider must be specified for public clients");
+    }
+
+    let endpoint = "";    
+    if (this.provider === PublicBitcoinProvider.MEMPOOL) {
+      // Mempool API doesn't respect count properly for busy addresses
+      // Request more than needed and slice client-side
```
Legend101Zz, line 704: why ? 
and why request extra in `MEMPOOL` only and not `BLOCKSTREAM` ?

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+    }
+
+    const rawTransactions = await this.Get(endpoint);
+    
+    if (!Array.isArray(rawTransactions)) {
+      throw new Error("Invalid response format from API");
+    }
+
+    // Handle pagination issues - slice to requested count
+    const limitedTransactions = rawTransactions.slice(0, count);
```
Legend101Zz, line 721: shouldn't this be handled by the api call itself ... like if we ask it for certain amount of tx's uisng `limit=${count}&offset=${skip}` then why do we need to deal with this later again ?

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+    }
+
+    // Public client (Mempool/Blockstream)
+    if (!this.provider) {
+      throw new Error("Provider must be specified for public clients");
+    }
+
+    let endpoint = "";    
+    if (this.provider === PublicBitcoinProvider.MEMPOOL) {
+      // Mempool API doesn't respect count properly for busy addresses
```
Legend101Zz, line 703: why ? any reference or docs

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,101 @@ export class BlockchainClient extends ClientBase {
...
+      const safeCount = Math.max(count * 2, 25); // Request at least 25 or double what's needed
+      endpoint = `/address/${address}/txs?count=${safeCount}&skip=${skip}`;
+    } else if (this.provider === PublicBitcoinProvider.BLOCKSTREAM) {
+      // Blockstream API respects pagination properly
+      endpoint = `/address/${address}/txs?limit=${count}&offset=${skip}`;
+    } else {
+      throw new Error(`Unsupported provider: ${this.provider}`);
+    }
+
+    const rawTransactions = await this.Get(endpoint);
```
Legend101Zz, line 714: cannot we do something like

```ts
const query =
  this.provider === PublicBitcoinProvider.BLOCKSTREAM
    ? `limit=${count}&offset=${skip}`
    : `count=${count}&skip=${skip}`;

const endpoint = `/address/${address}/txs?${query}`;
const rawTransactions = await this.Get(endpoint);
```

Legend101Zz (commented):
@ice-009 we are getting closure but still some more work left to do here :)

packages/caravan-clients/src/client.ts
ice-009: yeah not needed i was just validating that i get enough details

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+  skip: number = 0
+): Promise<TransactionDetails[]> {
+  if (count < 1 || count > 100) {
+    throw new Error("Count must be between 1 and 100");
+  }
+  if (skip < 0) {
+    throw new Error("Skip must be non-negative");
+  }
+
+  try {
```
Legend101Zz, line 652: can we add some docs / comments here explaining why this is the case like based on the conversations we had in discord ... and what we found it would help future devs and the package users also maybe a doc string above saying what the fn does and why we cannot use it for public clients .. would work nice with IDE :)

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+    // Fetch transactions for each address
+    await Promise.all(
+      addresses.map(async (addr) => {
+        const query = this.provider === PublicBitcoinProvider.BLOCKSTREAM
+          ? `limit=${count}&offset=${skip}`
+          : `count=${count}&skip=${skip}`;
+        const endpoint = `/address/${addr}/txs?${query}`;
+        const rawTransactions = await this.Get(endpoint);
+        
+        if (Array.isArray(rawTransactions)) {
```
Legend101Zz, line 675: like we don't want to drop a tx just because it was not recieved as an array right ? maybe we get a deep nested object then it's our job to convert that to array or have you checked that the api always returns arrays ? still then `if(rawTransactions)` check is better

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+        }
+      })
+    );
+
+    // Sort by timestamp (newest first) and apply pagination
+    const sortedTransactions = allTransactions
+      .sort((a, b) => (b.time || b.timestamp || 0) - (a.time || a.timestamp || 0))
+      .slice(0, count);
+
+    return sortedTransactions.map((rawTx: any) =>
```
Legend101Zz, line 686: so I can se this definition of normalise fn 

```ts
export function normalizeTransactionData(
  txData: RawTransactionData,
  ```
  
  so does this work well ... like I am not sure why didn't ts throw any errors in your IDE ?

File: packages/caravan-clients/src/client.ts [hunk 16]
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+    return sortedTransactions.map((rawTx: any) => 
+      normalizeTransactionData(rawTx, ClientType.PUBLIC)
+    );
+
+  } catch (error: any) {
+    throw new Error(`Failed to get address transaction history: ${error.message}`);
+  }
+}
+
+public async getWalletTransactionHistory(
```
Legend101Zz, line 695: same here some docstrings :)

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+
+    const response = await callBitcoindWallet({
+      baseUrl: this.bitcoindParams.url,
+      walletName: this.bitcoindParams.walletName,
+      auth: this.bitcoindParams.auth,
+      method: "listtransactions",
+      params: ["*", count, skip, includeWatchOnly],
+    });
+
+    if (!response?.result || !Array.isArray(response.result)) {
```
Legend101Zz, line 725: again why the array check ?

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+      throw new Error("Failed to retrieve transactions from Bitcoin Core");
+    }
+
+    // Filter only "send" category transactions (spent transactions)
+    // Unspent transactions are already available through other methods
+    const spentTransactions = response.result
+      .filter((tx: any) => tx.category === "send");
+
+    return spentTransactions.map((tx: any) =>
```
Legend101Zz, line 734: same here why any ?

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+    throw new Error(`Failed to get address transaction history: ${error.message}`);
+  }
+}
+
+public async getWalletTransactionHistory(
+  count: number = 100,
+  skip: number = 0,
+  includeWatchOnly: boolean = true
+): Promise<TransactionDetails[]> {
+  if (count < 1 || count > 1000) {
```
Legend101Zz, line 700: just a question here ... do we restrict to 1000 or allow asking for even more like 100k tx's ?

Legend101Zz (approved):
@ice-009 really nice work ... no complaints in terms of functionality, just a few minor nits from my end and once you are done with those we can ask Buck for a final review and merge it :)

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+    // Public client implementation
+    if (!this.provider) {
+      throw new Error("Provider must be specified for public clients");
+    }
+
+    const addresses = Array.isArray(address) ? address : [address];
+    const allTransactions: any[] = [];
+
+    // Fetch transactions for each address
+    await Promise.all(
```
Legend101Zz, line 667: maybe `Promise.allSettled` is better choice ... more here https://stackoverflow.com/questions/59784175/differences-between-promise-all-and-promise-allsettled-in-js

packages/caravan-clients/src/client.ts
ice-009: i have added the necessary steps for this

packages/caravan-clients/src/client.ts
ice-009: I have changed from any to rawTxnData

packages/caravan-clients/src/client.ts
ice-009: added docstrings

packages/caravan-clients/src/client.ts
ice-009: added promise.allsettled thanks

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,109 @@ export class BlockchainClient extends ClientBase {
...
+      addresses.map(async (addr) => {
+        const query = this.provider === PublicBitcoinProvider.BLOCKSTREAM
```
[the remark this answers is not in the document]
Legend101Zz: @bucko13 your takes on this also please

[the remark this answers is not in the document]
bucko13: Hmmm, good question. Kind of suggests that this method shouldn't accept an array of addresses imo. If you want to collect across addresses then you can build that out by building out multiple calls of this method.

packages/caravan-clients/src/client.test.ts
```diff
@@ -1823,4 +1825,327 @@ describe("BlockchainClient", () => {
...
+    vi.restoreAllMocks();
+  });
+
+  describe('parameter validation', () => {
+    it('should validate count parameter', async () => {
+      const client = new BlockchainClient({
+        type: ClientType.PUBLIC,
+        provider: PublicBitcoinProvider.MEMPOOL,
+        network: Network.MAINNET,
+      });
```
bucko13, line 1849: you can move this into a beforeEach (or just create once in the `describe` block if these are all the same args)

packages/caravan-clients/src/client.test.ts
```diff
@@ -1823,4 +1825,327 @@ describe("BlockchainClient", () => {
...
+    });
+
+    it('should use correct endpoint for BLOCKSTREAM', async () => {
+      const client = new BlockchainClient({
+        type: ClientType.PUBLIC,
+        provider: PublicBitcoinProvider.BLOCKSTREAM,
+        network: Network.MAINNET,
+      });
+
+      const mockResponse = [
```
bucko13, line 1935: no need to re-create the same fixture. Would the tests fail if you didn't mock the results? Unless you're testing any mutations that the function does, it shouldn't be necessary to return a full object.

packages/caravan-clients/src/client.test.ts
```diff
@@ -1823,4 +1825,327 @@ describe("BlockchainClient", () => {
...
+    mockCallBitcoindWallet = vi.spyOn(wallet, 'callBitcoindWallet');
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  describe('parameter validation', () => {
+    it('should validate count parameter', async () => {
+      const client = new BlockchainClient({
```
bucko13, line 2017: pull this out. If you need to have flexibility with args, you can create a function `const getNewClient = (argOverrides) => new BlockChainClient({ ...defaultArgs, ...argOverrides })`

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,228 @@ export class BlockchainClient extends ClientBase {
...
+            ? `limit=${count}&offset=${skip}`
+            : `count=${count}&skip=${skip}`;
+          const endpoint = `/address/${addr}/txs?${query}`;
+          const response = await this.Get(endpoint);
+          
+          // Robust response handling
+          let transactions: any[] = [];
+          
+          // Case 1: Response is already an array
+          if (Array.isArray(response)) {
```
bucko13, line 709: Are we testing the handling of different response shapes and the transformations that might be done in them?

bucko13 (changes requested):
I tested this out on a local regtest and unfortunately it doesn't work due to the response for this particular rpc call not being the same as other transaction requests to the bitcoind node. 

I've also requested a couple of other changes to cleanup testing and hopefully target the specific wallet.

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,228 @@ export class BlockchainClient extends ClientBase {
...
+    if (!this.bitcoindParams.walletName) {
+      throw new Error("Wallet name is required for private client transaction listings");
+    }
+
+    const response = await callBitcoindWallet({
+      baseUrl: this.bitcoindParams.url,
+      walletName: this.bitcoindParams.walletName,
+      auth: this.bitcoindParams.auth,
+      method: "listtransactions",
+      params: ["*", count, skip, includeWatchOnly],
```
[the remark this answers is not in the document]
Legend101Zz: > does the api support passing in a wallet name? If there are multiple wallets loaded we could get unrelated history.

I guess if we need to call a method using `callBitcoindWallet` then by default that method is specific to a wallet , because we do provide as RPC arguments the wallet name and also `listunspent` and other methods work the same way :)

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +637,228 @@ export class BlockchainClient extends ClientBase {
...
+      throw new Error("Failed to retrieve transactions from Bitcoin Core");
+    }
+
+    // Filter only "send" category transactions
+    const spentTransactions = response.result
+      .filter((tx: any) => tx.category === "send")
+      // Ensure we don't exceed requested count
+
+    return spentTransactions.map((tx: RawTransactionData) => 
+      normalizeTransactionData(tx, ClientType.PRIVATE)
```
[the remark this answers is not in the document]
Legend101Zz: @ice-009 any reasons why the vin , vout are empty and size of tx is 0 in your tx ? 
just curious :)

packages/caravan-clients/src/client.ts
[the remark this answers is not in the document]
ice-009: i guess the implementation is fine @Legend101Zz since we are selecting the wallet name by rpcwallet and since it allows our wallet selection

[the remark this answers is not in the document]
bucko13: ah ok, so as long as we're passing rpcwallet name then the `*` should just be all history for that wallet? that works for me!

packages/caravan-clients/src/client.test.ts
```diff
@@ -1823,4 +1825,411 @@ describe("BlockchainClient", () => {
...
+          }
+        ];
+
+        const mockGet = vi.fn().mockResolvedValue(mockResponse);
+        client.Get = mockGet;
+
+        const result = await client.getAddressTransactionHistory(testAddress, 15, 30);
+
+        expect(mockGet).toHaveBeenCalledWith(`/address/${testAddress}/txs?limit=15&offset=30`);
+        expect(result).toHaveLength(1);
```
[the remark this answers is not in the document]
ice-009: solved this in this commit [https://github.com/caravan-bitcoin/caravan/pull/340/commits/01eb5e2b7fbaa8fb062d252e40e9571f81ab56a1](https://github.com/caravan-bitcoin/caravan/pull/340/commits/01eb5e2b7fbaa8fb062d252e40e9571f81ab56a1)

packages/caravan-clients/src/client.ts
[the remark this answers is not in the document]
ice-009: yes we are doing that only!!

packages/caravan-clients/src/client.ts
```diff
@@ -637,6 +638,234 @@ export class BlockchainClient extends ClientBase {
...
+        blockHeight: tx.blockheight,
+        blockHash: tx.blockhash,
+        blockTime: tx.blocktime,
+      },
+      confirmations: tx.confirmations,
+      category: tx.category,
+      address: tx.address,
+      abandoned: tx.abandoned,
+      time: tx.time,
+    } as TransactionDetails;
```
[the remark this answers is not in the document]
ice-009: we have confirmed this here [https://github.com/caravan-bitcoin/caravan/pull/340/commits/c841fd74652813b6027dfe183a6385280d667243](https://github.com/caravan-bitcoin/caravan/pull/340/commits/c841fd74652813b6027dfe183a6385280d667243)

packages/caravan-clients/src/client.ts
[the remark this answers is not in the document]
ice-009: my regtest is abnormal bro :) It dont even mine coins now :)

bucko13 (approved):
Looks good and tested with a console log again and returned the transacitons as expected! I have one question about the types but I think that's non-blocking (though if we later changed it would be a backwards incompatible change so we should consider first before merging)

PR #348, opened 2025-07-19
Feat/rbf/UI/part3

Legend101Zz:
Final Phase of RBF integration to caravan 🥳

### DIFF
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -507,15 +507,20 @@ export function importPSBT(psbtText, inputs, isRbfPBST) {
 
     // ==== Extract and import signatures (If they are present)====
 
-    // For RBF PSBTs, we skip signature extraction because:
-    // 1. The UTXOs used for signature verification aren't in wallet state
-    // 2. The signature extraction functions expect UTXOs to be in spendable slices
-    // 3. RBF PSBTs are typically unsigned anyway (they're new replacement transactions)
-    // 4. Attempting signature extraction would throw errors ...
+    // We currently skip signature extraction for RBF PSBTs because:
+    // 1. The UTXOs required for signature verification might not be available in the current wallet state.
+    // 2. Signature extraction is tightly coupled to `selectAvailableInputsFromPSBT`, which relies on spendable slices.
+    // 3. In some flows (e.g., external RBF workflows), partially signed PSBTs may still be valid, but our logic doesn’t currently support those cases.
+    // 4. Attempting extraction without proper UTXO context may cause errors or incorrect behavior.
     //
-    // Note: If we don't skip this step on RBFed PSBT then setSignaturesFromPsbt confirms for signature by comparing
-    // PSBT inputs against wallet UTXO's and it throws error then that UTXO does not belong to wallet
-    // Also saves compute if skipped :)
+    // NOTE: This is a limitation of the current implementation, not a fundamental restriction.
+    // It's entirely possible for RBF PSBTs to be partially signed—e.g., in shared wallets where one party creates and signs an RBF,
+    // then sends it to a co-signer for completion.
+    //
+    // TODO: Decouple signature extraction from wallet-bound spendable slices so we can support
+    // partial sig extraction for externally signed or collaboratively constructed PSBTs.
+    //
+    // For now, we skip extraction to avoid breaking flows that lack full UTXO context.
     if (!isRbfPBST) {
       dispatch(setSignaturesFromPsbt(psbt));
     }
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -474,7 +474,7 @@ export function setFeeAndFeeRateFromState(outputsTotalSats, psbt) {
   };
 }
 
-export function importPSBT(psbtText, inputs, isRbfPBST) {
+export function importPSBT(psbtText, inputs, hasPendingInputs) {
   return (dispatch, getState) => {
     let state = getState();
     const { network } = state.settings;
@@ -521,7 +521,7 @@ export function importPSBT(psbtText, inputs, isRbfPBST) {
     // partial sig extraction for externally signed or collaboratively constructed PSBTs.
     //
     // For now, we skip extraction to avoid breaking flows that lack full UTXO context.
-    if (!isRbfPBST) {
+    if (!hasPendingInputs) {
       dispatch(setSignaturesFromPsbt(psbt));
     }
 
--- a/apps/coordinator/src/actions/transactionActions.js
+++ b/apps/coordinator/src/actions/transactionActions.js
@@ -3,7 +3,6 @@ import {
   estimateMultisigTransactionFee,
   estimateMultisigTransactionFeeRate,
   satoshisToBitcoins,
-  bitcoinsToSatoshis,
 } from "@caravan/bitcoin";
 import { getSpendableSlices, getConfirmedBalance } from "../selectors/wallet";
 import {
@@ -449,12 +448,28 @@ export function setSignaturesFromPsbt(psbt) {
  * @param {BigNumber} outputsTotalSats - Total satoshis in all outputs
  * @returns {Function} Redux thunk function
  */
-export function setFeeFromState(outputsTotalSats) {
+export function setFeeAndFeeRateFromState(outputsTotalSats, psbt) {
   return (dispatch, getState) => {
     const state = getState();
+    const {
+      settings: { addressType, requiredSigners, totalSigners },
+    } = state;
     const inputsTotalSats = BigNumber(state.spend.transaction.inputsTotalSats);
     const feeSats = inputsTotalSats.minus(outputsTotalSats);
     const fee = satoshisToBitcoins(feeSats);
+
+    const feeRate = estimateMultisigTransactionFeeRate({
+      addressType,
+      numInputs: psbt.txInputs.length,
+      numOutputs: psbt.txOutputs.length,
+      m: requiredSigners,
+      n: totalSigners,
+      feesInSatoshis: feeSats,
+    });
+
+    if (feeRate) {
+      dispatch(setFeeRate(Math.floor(parseFloat(feeRate)).toString()));
+    }
     dispatch(setFee(fee));
   };
 }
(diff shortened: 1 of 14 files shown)


apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
```diff
@@ -0,0 +1,295 @@
...
+  onImport,
+  network,
+  disabled = false,
+}) => {
+  const [isProcessing, setIsProcessing] = useState(false);
+  const [error, setError] = useState<string>("");
+  const [psbtText, setPsbtText] = useState<string>("");
+  const [parsedPsbt, setParsedPsbt] = useState<any>(null);
+
+  const allSlices = useSelector(getWalletSlices);
```
Legend101Zz, line 45: do you want me to make this chunk into a hook like line 45 to 100 ?

Legend101Zz:
demo video here ...


https://github.com/user-attachments/assets/9bea530d-c61a-40d5-9e8a-abbf82722e6b

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -500,7 +489,19 @@ export function importPSBT(psbtText) {
     dispatch(setFeeFromState(outputsTotalSats));
 
     // ==== Extract and import signatures (If they are present)====
-    dispatch(setSignaturesFromPsbt(psbt));
+
+    // For RBF PSBTs, we skip signature extraction because:
+    // 1. The UTXOs used for signature verification aren't in wallet state
+    // 2. The signature extraction functions expect UTXOs to be in spendable slices
+    // 3. RBF PSBTs are typically unsigned anyway (they're new replacement transactions)
```
bucko13, line 496: These two things don't necessarily need to be true. We could fix sig extraction to be more decoupled from spendable slices and it's definitely not true that RBF psbts are unsigned. In fact, even caravan only use cases, it's possible that you and I are in a wallet together, I create an RBF transaction, sign it, send you the signed PSBT to sign as well and broadcast. 

So even if these are restrictions currently, we shouldn't say so definitively. They don't need to be and shouldn't be long term.

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
bucko13: Huh yeah that seems reasonable! Seems like it could all be part of getting g reconstructed utxos right?

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
```diff
@@ -0,0 +1,323 @@
...
+    reconstructionLoading,
+    parsedPsbt,
+    allInputs.length,
+    isRbfPSBT,
+    psbtText,
+    onImport,
+  ]);
+
+  // Helper function to detect if content is binary PSBT
+  const isBinaryPSBT = useCallback((arrayBuffer: ArrayBuffer): boolean => {
```
bucko13, line 153: This can be pulled out as a utility function. No need for it to be in scope of the component or memoized

apps/coordinator/src/actions/transactionActions.js
Legend101Zz: > it's possible that you and I are in a wallet together, I create an RBF transaction, sign it, send you the signed PSBT to sign as well and broadcast.

Yup I agree I was testing few things and I am not able to import a partially signed RBF PSBT because of this ... so we need a better check and need to remove this

Legend101Zz: But  @bucko13 — the process starts becoming a bit too convoluted just to handle this specific caveat.

Right now, extracting partial signatures from a PSBT is super clean using this selector:

```ts
/**
 * Selector to extract signatures from PSBT
 */
export const selectSignaturesFromPSBT = createSelector(
  [selectAvailableInputsFromPSBT, (state: any, psbt: Psbt) => psbt],
  (inputs, psbt) => extractSignaturesFromPSBT(psbt, inputs),
);
```

The issue is that this only works when we have the `selectAvailableInputsFromPSBT` precomputed — which is fine for the standard flow. But if we want to support something like `selectAllPSBTInputs` (for cases like RBF), it gets trickier. The RBF flow handles reconstruction in a separate hook, and wiring that into the selector logic isn’t straightforward.

One possible workaround would be to have that hook dispatch the reconstructed UTXOs it already computes and persist them into state. That way, we could plug those into the partial sig extraction logic more cleanly.

But it’s a bit hacky and adds more moving parts than I’d like. Feels like we’re introducing workaround code just to patch over a structural mismatch.

Would love to hear your thoughts — on this ?

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
Legend101Zz: changes done https://github.com/caravan-bitcoin/caravan/pull/348/commits/c6d46503c61627c2fe01151758400d78e2da224d

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
Legend101Zz: changes done https://github.com/caravan-bitcoin/caravan/pull/348/commits/c6d46503c61627c2fe01151758400d78e2da224d

apps/coordinator/src/actions/transactionActions.js
bucko13: Yeah definitely doesn't need to be handled in this PR! But something that I'd like to try and address at some point if we can. Mainly I just wanted to make sure we clean up the language around the comments which made it sound like something that wasn't possible or we wouldn't do and I don't think either of those things are true.

apps/coordinator/src/hooks/utxos.ts
```diff
@@ -201,3 +239,162 @@ export const useGetAvailableUtxos = (transaction?: TransactionDetails) => {
...
+      : new Set<string>(),
+  );
+
+  const availableInputs = useSelector((state: WalletState) =>
+    parsedPsbt ? selectAvailableInputsFromPSBT(state, parsedPsbt) : [],
+  );
+
+  const { transactions: pendingTxs } = usePendingTransactions();
+
+  // Only fetch reconstructed UTXOs if there are missing inputs
```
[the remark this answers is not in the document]
Legend101Zz: ahh sorry redundant now ... also we cannot make hooks conditionally rendered so skipped :)

apps/coordinator/src/components/Wallet/WalletSpend.jsx
```diff
@@ -146,90 +146,13 @@ class WalletSpend extends React.Component {
...
-      uint8Array.length >= 5 &&
-      uint8Array[0] === 0x70 &&
-      uint8Array[1] === 0x73 &&
-      uint8Array[2] === 0x62 &&
-      uint8Array[3] === 0x74 &&
-      uint8Array[4] === 0xff
-    );
-  };
-  handleImportPSBT = ({ target }) => {
+  handleImportPSBT = (psbtText, resolvedInputs, isRbfPBST) => {
```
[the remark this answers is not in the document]
Legend101Zz: Left this as is — because if we directly pass the function, and a user imports a wrong PSBT (e.g., one with invalid signatures), we end up in an infinite render loop ( our old friend )

This happens because:

* The parent creates a new inline `onImport` function on each render.
* Inside `PSBTImportComponent`, `onImport` is called inside a `useEffect` once inputs are resolved and then ,
* `importPSBT` triggers a state update casuing the parent to re-renders.
* That causes a new `onImport` to be passed, which triggers the effect again.
* This cycle continues indefinitely here after happily :)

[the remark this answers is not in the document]
bucko13: I'm still not tracking. you're still passing in an inline function that would be recreated on re-renders.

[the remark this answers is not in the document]
Legend101Zz: > I'm still not tracking. you're still passing in an inline function that would be recreated on re-renders.

Hmm ... this helped me clarify my doubts too  https://youtu.be/sdhHsyazoaI?si=zAR5HK4okaZFSKWl&t=366 near about 6 mins :)

apps/coordinator/src/hooks/utxos.ts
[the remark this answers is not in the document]
Legend101Zz: done https://github.com/caravan-bitcoin/caravan/pull/348/commits/c7eda0e42674058f0b768c8c4f28da3cb7444bee. thanks

apps/coordinator/src/utils/uxtoReconstruction.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  // =====================================================================
+  // STRATEGY 2: The RBF Path - Check Reconstructed UTXOs
+  // =====================================================================
+  //
+  // This is where the magic happens for RBF PSBTs. If we get here, it means
+  // Strategy 1 didn't find all the UTXOs the PSBT needs. That's a strong
+  // indicator we're dealing with an RBF scenario.
+  //
+  // The reconstructed UTXOs come from analyzing pending transactions to
```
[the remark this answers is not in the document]
Legend101Zz: yup so should I add that check ? 
ahh ... we alrady do that 

```js

    if (!reconstructionLoading && reconstructedUtxos.length > 0) {
      return matchPsbtInputsToUtxos(
        psbtInputIdentifiers,
        availableInputs,
        reconstructedUtxos,
      );
      ```
     
    `reconstructedUtxos.length > 0` does same `:)

File: apps/coordinator/src/utils/uxtoReconstruction.ts [hunk 7]
```diff
@@ -0,0 +1,296 @@
...
+ * @param pendingTransactions - Unconfirmed transactions from our wallet (used as data source).
+ * @param originalTxLookup - Lookup map of txid → original transaction and hex data.
+ * @param allSlices - Wallet metadata (used to verify ownership of reconstructed UTXOs).
+ * @param neededInputIds - Set of UTXO input IDs (txid:vout) we want to reconstruct.
+ *
+ * @throws {Error} When network is unavailable or reconstruction fails
+ */
+export function reconstructUtxosFromPendingTransactions(
+  pendingTransactions: TransactionDetails[],
+  originalTxLookup: Map<
```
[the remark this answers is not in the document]
Legend101Zz: hmm basically originalTx lookup we give in the pending tx's inputs original tx's ( whose data has been collected via tanQuery)

> This utility scans through a set of unconfirmed (pending) transactions and identifies which of their inputs correspond to UTXOs we are attempting to reconstruct. It then extracts the unique txids associated with those relevant inputs.

apps/coordinator/src/hooks/utxos.ts
```diff
@@ -201,3 +239,162 @@ export const useGetAvailableUtxos = (transaction?: TransactionDetails) => {
...
+
+    const txLookup = new Map(
+      transactionQueries
+        .filter((q) => q.data)
+        .map((q) => [q.data!.txid, q.data!]),
+    );
+
+    return reconstructUtxosFromPendingTransactions(
+      pendingTransactions,
+      txLookup,
```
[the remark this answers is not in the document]
bucko13: nevermind. I think I was confusing myself. pendingTransactions are txs in the mempool. txLookup is actually already confirmed txs that created the inputs that are being spent in a pending tx. So you're combining:

- pending wallet txs in the mempool
- a map for tx data for all txs that created inputs being spent in the mempool
- allSlices for the wallet
- the inputs that we care about for a specific spend condition

I think this data is all sufficiently isolated that this makes sense to me.

bucko13 (changes requested):
one minor nit left. Just gotta finish the testing and then I think we're good!

bucko13 (approved):
I tested this on regtest, full flow with a ledger signing. But have not had an opportunity to test with a public client yet.

PR #349, opened 2025-07-22
feat(fee-bumping): Implement Child-Pays-for-Parent (CPFP) Fee Bumping

Legend101Zz:
### **PR Description:**

This pull request introduces support for **Child-Pays-for-Parent (CPFP)** as a new fee-bumping strategy within the Caravan Coordinator. This provides users with a powerful method to accelerate unconfirmed transactions, especially in cases where the original transaction did not signal Replace-by-Fee (RBF).

### **TODOs:**

* [ ] **Initial Integration Testing:** ( Maybe in another PR) Perform end-to-end testing of the new CPFP flow within the Caravan Coordinator UI to ensure all components work together as expected.
* [X] **Verify & Refine Review Step:** Double-check that the `TransactionComparison` component accurately reflects all CPFP scenarios and correctly calculates the combined package statistics.
* [X] **Update `@caravan/fees` Package:** Add comprehensive PSBT tests to the underlying `createCPFPTransaction` function to validate its logic ...
* [X] **Final Testing:** Conduct final, thorough testing of the entire feature before merging.

### DIFF
--- a/apps/coordinator/src/utils/uxtoReconstruction.ts
+++ b/apps/coordinator/src/utils/uxtoReconstruction.ts
@@ -363,6 +363,13 @@ export const buildUtxoFromSpendingTransaction = (
   }
 };
 
+// TODO: Type conversion complexity in convertCaravanUtxoToCoin
+// This conversion chain (ReconstructedUtxos → Coin → UTXO) highlights unnecessary
+// type complexity. We should consolidate ReconstructedUtxos and Coin into a single
+// type to eliminate this intermediary conversion. This would simplify the codebase
+// by having reconstructSingleUtxo directly return a Coin object that can be passed
+// to getUtxoFromCoin.
+
 /**
  * Converts a Caravan-style UTXO object to a Coin object.
  *


Legend101Zz:
https://github.com/user-attachments/assets/f692fa15-3ce7-4fed-af93-32949e7a5c70

@bucko13 CPFP is almost ready , just left with a few fixes to get the import functionality working for new CPFP — I’ll work on finalizing it tomorrow and will ping you once it’s ready for review :)

Legend101Zz:
https://github.com/user-attachments/assets/eb911bb4-f64f-443c-a538-f06cfad85795

CPFP works now in caravan 🥳

apps/coordinator/src/clients/transactions.ts
```diff
@@ -83,27 +123,44 @@ export const usePendingTransactions = () => {
...
+  const missingFeeTxids = pendingTransactions
+    .filter((txid) => !txid.fee)
+    .map((tx) => tx.txid);
+
+  const {
+    feeMap,
+    isLoading: feesLoading,
+    error: feesError,
+    refetchAll: refetchFees,
+  } = useFetchPendingFeesForTxids(missingFeeTxids);
```
bucko13, line 140: feels like this could be optimized and consolidated. What if we just called `fetchPendingTransactionFee(txid, blockchainClient!),` as part of the `useFetchPendingTransactions` call (under line 106 [here](https://github.com/caravan-bitcoin/caravan/pull/349/files#diff-dd401d860af81fb3974ed95aacbf0a436e5aa46d71872a8166f16f52d8455831R106))?

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/AccelerationModal.tsx
```diff
@@ -78,7 +79,10 @@ const AccelerationModalWithContext: React.FC<AccelerationModalProps> = ({
   txHex,
 }) => {
   return (
-    <AccelerationModalProvider transaction={transaction} txHex={txHex}>
+    <AccelerationModalProvider
+      transaction={transaction as TransactionDetails}
```
bucko13, line 83: should `AccelerationModalProps` be changed so the `as` isn't necessary?

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/AccelerationModal.tsx
```diff
@@ -195,7 +199,9 @@ const AccelerationModalContent: React.FC<
     if (analysisError) {
       return (
         <ErrorDialog
-          error="Failed to load available UTXOs"
+          error={
+            analysisError ? analysisError : "Failed to load available UTXOs"
+          }
```
bucko13, line 204: this shouldn't be necessary. If there's no analysis error, then the `if` on line 199 won't return true and you'll never get here. 
```suggestion
          error={analysisError}
```

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/ConfigurationStep/index.tsx
```diff
@@ -15,16 +16,10 @@ export const ConfigurationStep: React.FC = () => {
   const {
     state: { selectedStrategy },
   } = useAccelerationModal();
-
+  console.log("selectedStrategy", selectedStrategy);
```
bucko13, line 19: Cleanup

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/ReviewStep/TransactionComparison.tsx
```diff
@@ -46,33 +49,48 @@ export const TransactionComparison: React.FC = () => {
...
         >
           <AlertTitle>
-            {rbfType === RBF_TYPES.CANCEL
-              ? "Cancel Transaction"
-              : "Accelerated Transaction"}
+            {isCPFP
+              ? "Child-Pays-for-Parent Transaction"
+              : rbfType === RBF_TYPES.CANCEL
+                ? "Cancel Transaction"
+                : "Accelerated Transaction"}
```
bucko13, line 86: I'm not a fan of nested ternary as I find them hard to read. Can we pull this out into some if/else statements to set a variable at the top? There are a few others so maybe we can set an object with several different properties (`title`, `body`, etc.)

bucko13: Not a huge deal if it makes things overly complex. Mainly just an aesthetic preference.

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/hooks.ts
```diff
@@ -58,6 +62,10 @@ export const useAnalyzeTransaction = (
     isError: isErrorAvailableUtxos,
   } = useGetAvailableUtxos(transaction!);
 
+  //  ChangeOutputIndex is critical for fee bumping strategies , like in CPFP (Child-Pays-For-Parent) it helps,
+  //  Identifies which output can be spent in a child transaction
```
bucko13, line 66:
```suggestion
  //  ChangeOutputIndex is critical for fee bumping strategies, specifically for CPFP it helps
  //  identify which output can be spent in a child transaction by the current wallet.
```

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/hooks.ts
```diff
@@ -98,15 +105,32 @@ export const useAnalyzeTransaction = (
         requiredSigners,
         totalSigners,
         addressType: addressType as MultisigAddressType,
+        ...(changeOutputIndex !== undefined && { changeOutputIndex }),
       });
-      return analyzer.analyze();
+      let cpfpData = null;
+      try {
+        // added here so that in case `analyzer.cpfpFeeRate` throws error it gets catched
```
bucko13, line 112:
```suggestion
        // added here in case `analyzer.cpfpFeeRate` throws an error
```

bucko13: why not let the error get thrown though and get caught and set on line 127 so it can be surfaced?

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/hooks.ts
```diff
@@ -117,10 +141,20 @@ export const useAnalyzeTransaction = (
...
     totalSigners,
+    changeOutputIndex,
   ]);
 
   return {
-    analysis: analysis ?? null,
+    analysis: analysis?.analysis ?? null,
+    cpfp: analysis?.cpfpData
+      ? {
+          // we need this because for CPFP we cannot calculate feeRate simply as in RBF by using vsize and RBFFees
```
bucko13, line 151:
```suggestion
          // we need this because for CPFP we cannot calculate feeRate simply by using vsize and fees (as we do for RBF) 
```

apps/coordinator/src/components/Wallet/TransactionsTab/TransactionsTable.tsx
```diff
@@ -56,32 +56,40 @@ export const FeeDisplay: React.FC<FeeDisplayProps> = ({
...
+    // feeInSats is actually in BTC format when isReceived is true
+    const feeInBTC = satoshisToBitcoins(feeInSats!) || "0";
+    const actualFeeInSats = Number(feeInSats);
+
     return (
-      <Tooltip title="Fee information not available" placement="top">
-        <Box display="flex" alignItems="center">
-          <Typography variant="body2" color="textSecondary" sx={{ mr: 0.5 }}>
-            --
+      <Tooltip title="You did not spend this fee" placement="top">
```
bucko13, line 80: I don't understand this comment in the title?

apps/coordinator/src/selectors/wallet.ts
```diff
@@ -282,10 +282,9 @@ export const getSpendableSlices = createSelector(
   (slices: SliceWithLastUsed[]): SliceWithLastUsed[] => {
     return slices.filter(
       (slice: SliceWithLastUsed) =>
-        // pending change is considered spendable
-        (slice.lastUsed !== "Pending" || slice.change) &&
-        slice.lastUsed !== "Spent" &&
-        slice.utxos.length,
+        // All pending change/deposit is considered spendable
+        // Needed to do this to make CPFP work
```
bucko13, line 286:
```suggestion
        // All pending change/deposit is considered spendable when we have to 
        // consider fee bumping strategies as in CPFP
```

apps/coordinator/src/utils/uxtoReconstruction.ts
```diff
@@ -294,3 +297,83 @@ export function matchPsbtInputsToUtxos(
...
+    txHex,
+    walletSlices,
+  );
+
+  // If reconstruction failed, this output doesn't belong to our wallet
+  if (!caravanUtxo) {
+    return null;
+  }
+  const coinForConversion = convertCaravanUtxoToCoin(caravanUtxo);
+  console.log("coinForConversion", coinForConversion, caravanUtxo);
```
bucko13, line 351: cleanup

apps/coordinator/src/utils/uxtoReconstruction.ts
```diff
@@ -294,3 +297,83 @@ export function matchPsbtInputsToUtxos(
...
+  const caravanUtxo = reconstructSingleUtxo(
+    input,
+    parentTransaction,
+    txHex,
+    walletSlices,
+  );
+
+  // If reconstruction failed, this output doesn't belong to our wallet
+  if (!caravanUtxo) {
+    return null;
```
bucko13, line 348: Is this an edge case you don't expect to happen or will happen for most txs because this is run on all outputs of the parent tx? if it's an edge case, maybe a `console.warn` to indicate that this was improperly called?

apps/coordinator/src/utils/uxtoReconstruction.ts
```diff
@@ -294,3 +297,83 @@ export function matchPsbtInputsToUtxos(
...
+  }
+};
+
+/**
+ * Converts a Caravan-style UTXO object to a Coin object.
+ *
+ * @param caravanUtxo - UTXO object returned by reconstructSingleUtxo
+ * @returns Coin object ready for getUtxoFromCoin conversion
+ */
+function convertCaravanUtxoToCoin(caravanUtxo: any): Coin {
```
bucko13, line 369: seems like we can define the type rather than `any`

packages/caravan-clients/src/client.ts
```diff
@@ -1054,7 +1059,7 @@ export class BlockchainClient extends ClientBase {
    * @see https://mempool.space/docs/api/rest#get-transaction
    *
    * @param txid - Transaction ID to get fees
-   * @returns Tx fees in BTC, or null if transaction is not pending
+   * @returns Tx fees in satoshis, or null if transaction is not pending
```
bucko13, line 1062: this is a backwards incompatible change and warrants a major version bump.

packages/caravan-fees/src/transactionAnalyzer.ts
```diff
@@ -295,14 +295,29 @@ export class TransactionAnalyzer {
...
     const desiredPackageFee = new BigNumber(this.targetFeeRate).multipliedBy(
       this.CPFPPackageSize,
     );
-    const expectedFeeRate = BigNumber.max(
-      desiredPackageFee.minus(this.fee).dividedBy(this.estimatedCPFPChildSize),
-      new BigNumber(0),
+    const feeDifference = desiredPackageFee.minus(this.fee);
+
+    // Check if the parent transaction already meets or exceeds the target fee rate , this was added because of a near impossible edge cause
+    // Discussed in this thread : https://discord.com/channels/1352754651044515941/1352758476870516796/1417181497302974484
```
bucko13, line 311: Let's not link to discord as I'm not sure how permanent that is. try and explain the edge case in a code comment if necessary to clarify the code.

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/ConfigurationStep/CPFP/CPFPForm.tsx
```diff
@@ -0,0 +1,372 @@
...
+        />
+      )}
+
+      {/* CPFP Information Alert */}
+      <Alert severity="info" sx={{ mb: 3 }}>
+        <Typography variant="body2">
+          Child-Pays-for-Parent (CPFP) creates a new transaction that spends an
+          output from your unconfirmed transaction. The child transaction pays a
+          high enough fee to incentivize miners to confirm both transactions
+          together.
```
bucko13, line 235: 👌

bucko13 (changes requested):
Code looks really good! Haven't tested yet but left some comments for possible cleanup.

apps/coordinator/src/clients/transactions.ts
bucko13: I think this can be done in a way that still helps with cacheing. It'd be nice if we just had the fee information available for all txs that don't include it from the beginning. Sometimes modularity can be taken too far imo.

If you add this to fetch pendingTransactions, and only make the extra call when there's no fee, then we will always just have the fee available and can avoid unnecessary extra hooks. It also means they're all in the same cache which is more readable and will probably be more efficient in the long run I think.

Legend101Zz: changes done thank you https://github.com/caravan-bitcoin/caravan/pull/349/commits/c6ebc6ffb46512809e599b58e5c0d5ec1bf0b1d0

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/hooks.ts
```diff
@@ -117,10 +141,20 @@ export const useAnalyzeTransaction = (
...
-    analysis: analysis ?? null,
+    analysis: analysis?.analysis ?? null,
+    cpfp: analysis?.cpfpData
+      ? {
+          // we need this because for CPFP we cannot calculate feeRate simply by using vsize and fees (as we do for RBF)
+          feeRate: analysis?.cpfpData?.feeRate,
+          childSize: analysis?.cpfpData?.childSize,
+          combinedEstimatedSize: analysis?.cpfpData?.combinedEstimatedSize,
+        }
+      : null,
```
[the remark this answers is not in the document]
Legend101Zz: done https://github.com/caravan-bitcoin/caravan/pull/349/commits/85a789dcb078d5f23d47f52e02b715d55c392cd0

apps/coordinator/src/clients/transactions.ts
```diff
@@ -15,6 +15,9 @@ const transactionKeys = {
   all: ["transactions"] as const,
   tx: (txid: string) => [...transactionKeys.all, txid] as const,
   pending: () => [...transactionKeys.all, "pending"] as const,
+  pendingTx: (txid: string) => [...transactionKeys.pending(), txid],
+  // fees for pending transaction
+  pendingTxFee: (txid: string) => [...transactionKeys.pendingTx(txid), "fee"],
```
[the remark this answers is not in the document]
bucko13: Can remove this now right/

apps/coordinator/src/utils/uxtoReconstruction.ts
```diff
@@ -294,3 +297,86 @@ export function matchPsbtInputsToUtxos(
...
+  if (!caravanUtxo) {
+    console.warn(
+      `Failed to reconstruct UTXO for output ${spendableOutputIndex} of transaction ${parentTransaction.txid}. ` +
+        `This output may not belong to the wallet or wallet slice data may be incomplete.`,
+    );
+    return null;
+  }
+  const coinForConversion = convertCaravanUtxoToCoin(caravanUtxo);
+  try {
+    return getUtxoFromCoin(coinForConversion);
```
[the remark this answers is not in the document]
bucko13: So I'm not totally convinced that this has to be unavoidable. It looks like that might mainly be because we've had disparate types that would be better to be consolidated if possible. Especially for future maintenance. 

The logic would be quite hard to follow in the future:

`reconstructSingleUtxo` => `convertCaravanUtxoToCoin` => `getUtxoFromCoin`

You strip out some data going from `caravanUtxo` from the intermediary coin object to what `getUtxoFromCoin` expects. But what if we consolidated the idea of the "caravan utxo" and a coin? Then `reconstructSingleUtxo` could actually just construct a `Coin` object and then that could be passed to `getUtxoFromCoin`. That would also unify our definitions and avoid the issue you were highlighting with the "Chimera" project. 

`reconstructSingleUtxo` becomes `reconstructSingleCoin` and change the property names to match of the object returned to match expectations.

Comparing the two types, they're not all that different and consolidating would be a big cleanup/tech debt win.

```typescript
export interface ReconstructedUtxos {
  txid: string;
  index: number;
  amountSats: string;
  amount: string;
  confirmed: boolean;
  transactionHex: any;
  multisig: any;
  bip32Path: string;
  change: boolean;
}
```
and 

```typescript
export interface Coin {
  prevTxId: string;
  vout: number;
  address: string;
  value: string;
  prevTxHex: string;
  slice?: Slice;
}
```

This would increase the scope of the PR quite a bit though (lots of function name changes like `useReconstructedUtxos` and `matchPsbtInputsToUtxos`) so maybe it makes sense to wait for a separate PR, but I think this would be beneficial for long term maintainability to do at some point, maybe as a follow up.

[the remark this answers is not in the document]
Legend101Zz: added todo https://github.com/caravan-bitcoin/caravan/pull/349/commits/fbe54a3e0c74364887479475e104f39d265c6044

apps/coordinator/src/clients/transactions.ts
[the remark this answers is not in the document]
Legend101Zz: done thanks https://github.com/caravan-bitcoin/caravan/pull/349/commits/cd4e3b6549ea0e1719950a07864e097b683a4c60

bucko13 (approved):

PR #355, opened 2025-07-29
Fix sig import

### LINKED ISSUE #354: Signature for input [X] is invalid
Manual UTXO selection appears to no longer work in Caravan. This has been reproduced with Trezor and Coldcard both being unable to sign.

Image

bucko13:
We need to be smarter about how we handle sequence numbers for inputs now. Changing the sequence will change the signatures and we're not passing it around correctly which is causing signing to fail. 

**Issue Number:**

Fixes #354 

**Snapshots/Videos:**

**Summary**

## Checklist

**Other information**

### DIFF
--- a/apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
+++ b/apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
@@ -30,7 +30,7 @@ class IndirectSignatureImporter extends React.Component {
   }
 
   interaction = () => {
-    const { signatureImporter, network, inputs, outputs } = this.props;
+    const { signatureImporter, network, inputs, outputs, psbt } = this.props;
     const keystore = signatureImporter.method;
     const bip32Paths = inputs.map((input) => {
       if (typeof input.bip32Path === "undefined")
@@ -44,6 +44,7 @@ class IndirectSignatureImporter extends React.Component {
       inputs,
       outputs,
       bip32Paths,
+      psbt,
     });
   };
 
@@ -235,6 +236,7 @@ IndirectSignatureImporter.propTypes = {
   Signer: PropTypes.shape({}).isRequired,
   fee: PropTypes.string,
   inputsTotalSats: PropTypes.shape({}),
+  psbt: PropTypes.string,
 };
 
 IndirectSignatureImporter.defaultProps = {
@@ -246,6 +248,7 @@ IndirectSignatureImporter.defaultProps = {
   extendedPublicKeyImporter: {},
   fee: "",
   inputsTotalSats: {},
+  psbt: "",
 };
 
 export default IndirectSignatureImporter;
--- a/apps/coordinator/src/hooks/utxos.ts
+++ b/apps/coordinator/src/hooks/utxos.ts
@@ -27,6 +27,8 @@ import {
   ReconstructedUtxos,
   matchPsbtInputsToUtxos,
 } from "utils/uxtoReconstruction";
+import { Psbt } from "bitcoinjs-lib";
+import { getInputIdentifiersFromPsbt } from "utils/psbtUtils";
 
 /*
  * need to create a function that given a coin and a slice returns a utxo that can be used
@@ -336,13 +338,13 @@ export const useReconstructedUtxos = (
  * @param  parsedPsbt - The parsed PSBT object
  * @returns  Contains allInputs, isRbfPSBT, loading/error states
  */
-export const usePsbtInputs = (parsedPsbt: any) => {
+export const usePsbtInputs = (parsedPsbt: Psbt) => {
   const allSlices = useSelector(getWalletSlices);
 
-  const psbtInputIdentifiers = useSelector((state: WalletState) =>
-    parsedPsbt
-      ? selectInputIdentifiersFromPSBT(state, parsedPsbt)
-      : new Set<string>(),
+  const psbtInputIdentifiers = useMemo(
+    () =>
+      parsedPsbt ? getInputIdentifiersFromPsbt(parsedPsbt) : new Set<string>(),
+    [parsedPsbt],
   );
 
   const missingInputIds = useSelector((state: WalletState) =>
--- a/apps/coordinator/src/selectors/transaction.ts
+++ b/apps/coordinator/src/selectors/transaction.ts
@@ -11,6 +11,8 @@ import {
   SignatureInfo,
   SignatureSet,
   SignerIdentity,
+  getSequenceForInput,
+  getInputIdentifiersFromPsbt,
 } from "utils/psbtUtils";
 
 // ====================
@@ -22,12 +24,6 @@ import {
  */
 const createInputIdentifier = (txid: string, index: number): string =>
   `${txid}:${index}`;
-
-/**
- * Converts transaction ID from big-endian to little-endian format
- */
-const convertTxidToLittleEndian = (hash: Buffer): string =>
-  reverseBuffer(hash).toString("hex");
 
 /**
  * Maps extracted signature sets to Caravan's signature importer format.
@@ -138,27 +134,9 @@ export const extractSignaturesFromPSBT = (psbt: Psbt, inputs: Input[]) => {
 // SELECTORS
 // ====================
 
-/**
- * Extract input identifiers from PSBT (txid:index)
- */
 export const selectInputIdentifiersFromPSBT = createSelector(
   [(state: WalletState, psbt: Psbt) => psbt],
-  (psbt: Psbt) => {
-    return new Set(
-      psbt.txInputs.map((input) => {
-        return createInputIdentifier(
-          /*
-           * All input TXIDs are expected to be in **big-endian**
-           * format (human-readable format). Which we get from block explorers, wallets, APIs
-           * But PSBTs will need txid to be in little-endian format to ensure compatibility with Bitcoin's
-           * internal data structures and processing so here we convert the txid to little-endian format
-           */
-          convertTxidToLittleEndian(input.hash),
-          input.index,
-        );
-      }),
-    );
-  },
+  (psbt: Psbt) => getInputIdentifiersFromPsbt(psbt),
 );
 
 /**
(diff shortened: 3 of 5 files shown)


apps/coordinator/jest.config.ts
bucko13, line 1: change required to support aliases in tests.

apps/coordinator/package.json
bucko13, line 1: this makes sure that the custom blockbook backend will be used for trezor connect in dev mode (which assumes regtest). This is still overridable though so hopefully uncontroversial.

apps/coordinator/src/actions/transactionActions.js
bucko13, line 1: anytime we run setInputs now, we should set the sequences correctly. I believe the RBF rule says only one input needs to signal but I don't think there's any harm in signaling (or not) for all inputs.

apps/coordinator/src/components/Coldcard/ColdcardSigner.jsx
bucko13, line 1: this is now just using the psbt text directly. this helps maintain consistency and avoid mutating the psbt by having to constantly recreate it.

apps/coordinator/src/components/ScriptExplorer/DirectSignatureImporter.jsx
```diff
@@ -63,9 +57,6 @@ class DirectSignatureImporter extends React.Component {
     return SignMultisigTransaction({
       network,
       keystore,
-      inputs,
-      outputs,
-      bip32Paths,
```
bucko13, line 68: these are no longer needed if the psbt is being passed in (which it is on line 72)

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
```diff
@@ -30,20 +30,13 @@ class IndirectSignatureImporter extends React.Component {
...
-      return `${signatureImporter.bip32Path}${input.bip32Path.slice(1)}`; // xpub/pubkey slice away the m, keep /
-    });
 
     return SignMultisigTransaction({
       keystore,
       network,
-      inputs,
-      outputs,
-      bip32Paths,
+      psbt,
```
bucko13, line 39: All interactions can now support building the instance based on a PSBT (trezor needed some surgery and a companion bump to caravan-wallets and caravan-psbt)

apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
```diff
@@ -398,21 +391,8 @@ class SignatureImporter extends React.Component {
 
         let publicKey;
         try {
-          const sequence = enableRBF ? 0xfffffffd : 0xffffffff;
```
bucko13, line 401: this was causing issues because we don't actually know what the sequence is and the signature commits to this value. This really should jut come from the psbt.

apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
```diff
@@ -398,21 +391,8 @@ class SignatureImporter extends React.Component {
...
-                ...convertedInput,
-                sequence: sequence, // Apply the same RBF sequence as in finalizeOutputs so if RBF signalling we don't lose that
-              };
-            }),
-            outputs: outputs.map(convertLegacyOutput),
-          };
-          const psbt = getUnsignedMultisigPsbtV0(args);
           publicKey = validateMultisigPsbtSignature(
-            psbt.toBase64(),
+            unsignedPSBT,
```
bucko13, line 395: rather than recreate the psbt based on the internal state, we can use the unsigned psbt we have already built

apps/coordinator/src/components/Wallet/WalletSpend.jsx
```diff
@@ -51,8 +50,6 @@ import {
 class WalletSpend extends React.Component {
   outputsAmount = new BigNumber(0);
 
-  coinSelection = naiveCoinSelection;
```
bucko13, line 54: this appeared to be unused.

apps/coordinator/src/hooks/utxos.ts
```diff
@@ -336,13 +337,13 @@ export const useReconstructedUtxos = (
...
+export const usePsbtInputs = (parsedPsbt: Psbt) => {
   const allSlices = useSelector(getWalletSlices);
 
-  const psbtInputIdentifiers = useSelector((state: WalletState) =>
-    parsedPsbt
-      ? selectInputIdentifiersFromPSBT(state, parsedPsbt)
-      : new Set<string>(),
+  const psbtInputIdentifiers = useMemo(
+    () =>
+      parsedPsbt ? getInputIdentifiersFromPsbt(parsedPsbt) : new Set<string>(),
```
bucko13, line 345: i don't know if this was necessary, but as I was debugging I did some cleanup where this selector wasn't doing anything with state. not sure if what I got was actually cleaner but some of the refactor that was necessary actually made this less necessary as a selector in other places (in fact I couldn't use it as a selector).

apps/coordinator/src/reducers/transactionReducer.js
```diff
@@ -472,7 +472,15 @@ export default (state = initialState(), action) => {
...
       return updateState(state, { spendingStep: action.value });
     case SET_ENABLE_RBF:
-      return updateState(state, { enableRBF: action.value });
+      return updateState(state, {
+        enableRBF: action.value,
+        // if RBF is toggled we need to make sure to update the
+        // sequence of all inputs to reflect this change
+        inputs: state.inputs.map((input) => ({
+          ...input,
+          sequence: action.value ? 0xfffffffd : 0xffffffff,
```
bucko13, line 481: if RBF is toggled, then the sequences needed to be similarly toggled.

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -138,36 +133,19 @@ export const extractSignaturesFromPSBT = (psbt: Psbt, inputs: Input[]) => {
...
-           * But PSBTs will need txid to be in little-endian format to ensure compatibility with Bitcoin's
-           * internal data structures and processing so here we convert the txid to little-endian format
-           */
-          convertTxidToLittleEndian(input.hash),
-          input.index,
-        );
-      }),
-    );
-  },
+  (psbt: Psbt) => getInputIdentifiersFromPsbt(psbt),
```
bucko13, line 138: taking this out as a utility function made it more reusable.

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -138,36 +133,19 @@ export const extractSignaturesFromPSBT = (psbt: Psbt, inputs: Input[]) => {
...
 );
 
 /**
  * Selector for "available inputs" from the unspent store
  * This is the first strategy - normal PSBT case where we get all the inputs we have in our wallet
  */
 export const selectAvailableInputsFromPSBT = createSelector(
-  [getSpendableSlices, selectInputIdentifiersFromPSBT],
-  (slices: any[], inputIdentifiers: Set<string>) => {
+  [getSpendableSlices, (state: WalletState, psbt: Psbt) => psbt],
```
bucko13, line 146: this selector needed to pass in the psbt to get the sequences as well. maybe there was another way to do this more idiomatically for reselect?

packages/caravan-psbt/src/psbtv0/psbt.test.ts
```diff
@@ -103,9 +103,15 @@ describe("translatePsbt", () => {
...
   const tx = _.cloneDeep(TEST_FIXTURES.transactions[0]);
+  tx.inputs.forEach((input) => {
+    input.sequence = 0xffffffff;
+  });
   const ms = MULTISIGS[0];
   it("handles P2WSH transactions", () => {
     const fixture = TEST_FIXTURES.transactions.find((tx) => tx.segwit);
+    fixture.inputs.forEach((input) => {
+      input.sequence = 0xffffffff;
+    });
```
bucko13, line 114: i didn't want to change the caravan-bitcoin fixtures to just support tests for sequence, so I just piggyback on the existing fixture.

packages/caravan-psbt/src/psbtv0/psbt.ts
```diff
@@ -349,6 +349,7 @@ inputs to protect against large fee attack`);
       index: input.index,
       transactionHex: fundingTxHex,
       txid: reverseBuffer(input.hash).toString("hex"),
+      sequence: input.sequence,
```
bucko13, line 352: this is required for the trezor interaction which needs to have inputs converted in a way that its api will work with so the input sequence needs to be passed through as well.

packages/caravan-wallets/src/trezor.ts
```diff
@@ -118,9 +117,7 @@ const ENV_TREZOR_BLOCKBOOK_URL =
 
 const TREZOR_CONNECT_URL =
   ENV_TREZOR_CONNECT_URL ||
-  (typeof window === "undefined"
-    ? ""
-    : `https://${window?.location.hostname}:8088/`);
+  (typeof window === "undefined" ? "" : "https://connect.trezor.io/9.6.0/");
```
bucko13, line 120: this change got local development working with a local blockbook instance (along with setting TREZOR_DEV=true)

apps/coordinator/src/components/Coldcard/ColdcardSigner.jsx
Legend101Zz: so if I am understanding it correctly, earlier what we did was use the whole transaction state to get the information about the transaction ... now instead we depend on unsignedPSBT that was originally created and avoid recreating PSBT again and again in the signing process using somthing like ...

```js
 let psbt = getUnsignedMultisigPsbtV0({
          network,
          inputs: inputs.map(convertLegacyInput),
          outputs: outputs.map(convertLegacyOutput),
        });
        ```

apps/coordinator/src/hooks/utxos.ts
Legend101Zz: hmm ... so does it mean we can remove it safely ?
But we do need it for `matchPsbtInputsToUtxos` fn right ?

Legend101Zz (changes requested):
It looks really great to me , almost ready to go ,thanks for putting time on this ... just a couple of questions and one request on adding doc string ... also I guess we need another cleanup in the Hermit component ....

apps/coordinator/src/components/Coldcard/ColdcardSigner.jsx
bucko13: Yep! Exactly. Everytime we try and recreate the PSBT we risk losing something. In the future as we solidify this pattern we can use the PSBT to track the state in the [PSBT saga](https://www.unchained.com/blog/operator-roles-life-stages-in-the-saga-of-a-psbt) which gives us even more "soundness" guarantees (something I know @Shadouts has thought a lot about with his work in caravan-psbt and bip370.org development)

apps/coordinator/src/hooks/utxos.ts
bucko13: I'm just referring to the change, the use of the function is still necessary, but maybe could've been kept as a selector.

PR #358, opened 2025-08-01
fix linting errors

bucko13:

Should fix issues that broke the build in main.

Screenshot 2025-08-01 at 1 11 34 PM


bucko13:
Turns out this was actually because there was an import in @caravan/wallets from @caravan/wallets rather than a relative import, but the other changes aren't bad to keep

PR #363, opened 2025-08-05
feat: xpub and PSBT signing integrated with BCUR2 support

krrish-sehgal:
Demo: 

https://github.com/user-attachments/assets/665bb3d6-d3f5-48cb-b8e1-6adf4344b36c

### DIFF
--- a/.changeset/brown-dragons-relate.md
+++ b/.changeset/brown-dragons-relate.md
@@ -1,5 +1,0 @@
----
-"@caravan/clients": patch
----
-
-Add getFeesForPendingTransaction to retrieve the fee information for a pending (incoming) transaction .
--- a/.changeset/few-rockets-drive.md
+++ b/.changeset/few-rockets-drive.md
@@ -1,5 +1,0 @@
----
-"caravan-coordinator": patch
----
-
-Include bip32 derivation for change output in unsignedPsbt
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -4,10 +4,6 @@ on:
   workflow_dispatch:
   workflow_call:
 
-permissions:
-  contents: read
-  checks: write
-
 jobs:
   e2e:
     name: End to End tests
@@ -56,7 +52,7 @@ jobs:
           BITCOIN_RPC_HOST=localhost
           BITCOIN_RPC_PORT=18443
           EOF
-      
+
       - name: Run e2e tests (with retry only on failures)
         uses: nick-invision/retry@v3
         with:
@@ -68,7 +64,7 @@ jobs:
             on_retry_command: |
               echo "Cleaning up before retry..."
               cd apps/coordinator/e2e
-              docker-compose down --remove-orphans || true 
+              docker-compose down --remove-orphans || true
               docker system prune -f || true
 
       # Upload results
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  checks: write
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ $ npm install turbo --global
 $ git clone https://github.com/caravan-bitcoin/caravan.git
 $ cd caravan
 $ npm install
-$ turbo run dev # or `npm run dev`
+$ turbo run dev --concurrency 11 # or `npm run dev`
 ```
 
 #### What's happening?
--- a/apps/coordinator/CHANGELOG.md
+++ b/apps/coordinator/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 1.11.3
+
+### Patch Changes
+
+- [#369](https://github.com/caravan-bitcoin/caravan/pull/369) [`66c44d2`](https://github.com/caravan-bitcoin/caravan/commit/66c44d2f09a22a49b1faf5e84ac50315a120b279) Thanks [@bucko13](https://github.com/bucko13)! - Include bip32 derivation for change output in unsignedPsbt
+
+- [#370](https://github.com/caravan-bitcoin/caravan/pull/370) [`944d1e0`](https://github.com/caravan-bitcoin/caravan/commit/944d1e0d6118b6df838125c6045ba6cf1da8ab3b) Thanks [@sambradbury](https://github.com/sambradbury)! - Updated @ledgerhq/hw-app-btc from ^5.34.1 to 10.10.0, updated @ledgerhq/hw-transport-webusb from 6.27.12 to 6.29.8
+
+- Updated dependencies [[`625191d`](https://github.com/caravan-bitcoin/caravan/commit/625191ddc8e1ddb1292c033448fee4fd77007aa6), [`944d1e0`](https://github.com/caravan-bitcoin/caravan/commit/944d1e0d6118b6df838125c6045ba6cf1da8ab3b)]:
+  - @caravan/clients@0.6.1
+  - @caravan/wallets@0.8.2
+
 ## 1.11.2
 
 ### Patch Changes
--- a/apps/coordinator/package.json
+++ b/apps/coordinator/package.json
@@ -1,7 +1,7 @@
 {
   "name": "caravan-coordinator",
   "private": true,
-  "version": "1.11.2",
+  "version": "1.11.3",
   "description": "Unchained Capital's Bitcoin Multisig Coordinator Application",
   "main": "index.jsx",
   "type": "module",
@@ -20,11 +20,7 @@
     "npm": ">=10.5.0 <11.0.0"
   },
   "overrides": {
-    "jspdf": "2.5.1",
-    "@ledgerhq/hw-transport-webusb": {
-      "@ledgerhq/hw-transport": "6.28.1",
-      "@ledgerhq/devices": "8.0.0"
-    }
+    "jspdf": "2.5.1"
   },
   "devDependencies": {
     "@babel/plugin-transform-runtime": "^7.21.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,6 @@
         "apps/*",
         "packages/*"
       ],
-      "dependencies": {
-        "@ledgerhq/hw-transport-webusb": "6.27.12"
-      },
       "devDependencies": {
         "@caravan/eslint-config": "*",
         "@caravan/typescript-config": "*",
@@ -37,7 +34,7 @@
     },
     "apps/coordinator": {
       "name": "caravan-coordinator",
-      "version": "1.11.2",
+      "version": "1.11.3",
       "license": "MIT",
       "dependencies": {
         "@caravan/bip32": "*",
@@ -5705,50 +5702,6 @@
       "resolved": "https://registry.npmjs.org/@ledgerhq/errors/-/errors-5.50.0.tgz",
       "integrity": "sha512-gu6aJ/BHuRlpU7kgVpy2vcYk6atjB4iauP2ymF7Gk0ez0Y/6VSMVSJvubeEQN+IV60+OBK0JgeIZG7OiHaw8ow=="
     },
-    "node_modules/@ledgerhq/hw-app-btc": {
-      "version": "5.51.1",
-      "resolved": "https://registry.npmjs.org/@ledgerhq/hw-app-btc/-/hw-app-btc-5.51.1.tgz",
-      "integrity": "sha512-/39HZZBl2yuZMIhe029uf44c7YTtKOtHaHCH8ZxOOXF0+/PvU6p5l8BLFZ1KjhVrJ5xKTBKxB/KACKZTAp4hgg==",
-      "dependencies": {
-        "@ledgerhq/hw-transport": "^5.51.1",
-        "@ledgerhq/logs": "^5.50.0",
-        "bip32-path": "^0.4.2",
-        "invariant": "^2.2.4",
-        "ripemd160": "2",
-        "semver": "^7.3.5",
-        "sha.js": "2"
-      }
-    },
-    "node_modules/@ledgerhq/hw-app-btc/node_modules/lru-cache": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-6.0.0.tgz",
-      "integrity": "sha512-Jo6dJ04CmSjuznwJSS3pUeWmd/H0ffTlkXXgwZi+eq1UCmqQwCh+eLsYOYCwY991i2Fah4h1BEMCx4qThGbsiA==",
-      "dependencies": {
-        "yallist": "^4.0.0"
-      },
-      "engines": {
-        "node": ">=10"
-      }
-    },
-    "node_modules/@ledgerhq/hw-app-btc/node_modules/semver": {
-      "version": "7.5.4",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.5.4.tgz",
-      "integrity": "sha512-1bCSESV6Pv+i21Hvpxp3Dx+pSD8lIPt8uVjRrxAUt/nbswYc+tK6Y2btiULjd4+fnq15PX+nqQDC7Oft7WkwcA==",
-      "dependencies": {
-        "lru-cache": "^6.0.0"
-      },
-      "bin": {
-        "semver": "bin/semver.js"
-      },
-      "engines": {
-        "node": ">=10"
-      }
-    },
-    "node_modules/@ledgerhq/hw-app-btc/node_modules/yallist": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/yallist/-/yallist-4.0.0.tgz",
-      "integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A=="
-    },
     "node_modules/@ledgerhq/hw-transport": {
       "version": "5.51.1",
       "resolved": "https://registry.npmjs.org/@ledgerhq/hw-transport/-/hw-transport-5.51.1.tgz",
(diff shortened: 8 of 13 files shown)


apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }, []);
+
+  const handleClearError = useCallback((): void => {
+    if (!isMountedRef.current) return;
+    setError("");
+    setShowScanner(false);
+  }, []);
+
+  const handlePSBTSuccess = useCallback(
+    (psbtData: string): void => {
```
bucko13, line 57: I feel like we should be able to split this up and modularize a bit. Some of the errors (like missing inputs) are separate from the processing of the PSBT itself. if there's no inputs passed in, can we do anything with this component? 

Furthermore, can the psbt processing be extracted out into its own utility function? Can we describe what this callback is doing exactly? The important bit is taking PSBT and extracting the signatures right? I'm looking at what the other signature importers do and it looks like the same logic doesn't happen. 

In fact, I'm curious why this looks so different from the Coldcard Signature importer. Coldcard and BCUR2 are both indirect, meaning they are airgapped. They also both deal directly in PSBTs. The PSBT is passed around as a file for ColdCard and for BCUR2 it's a QR code. Couldn't you just use the `IndirectSignatureImporter` (See `ColdcardSignatureImporter.jsx`) and pass in the Bcur2Reader as the Signer (or some wrapper around it)? Looking at the coldcard code, the `Signer`arg passed to the indirect importer is a component that gets as a prop `onReceivePSBT` which does this:

```
      // signatureData is one or more sets of signatures that are keyed
      // based on which pubkey the signatures are signing.
      const signatureData = this.interaction().parse(data);
```
which is basically some of what you're trying to do here.

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
bucko13, line 1: great idea adding a new PSBT import method!

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
```diff
@@ -196,41 +213,131 @@ export const PSBTImportComponent: React.FC<PSBTImportComponentProps> = ({
...
+  // Handle BCUR2 clear/cancel
+  const handleBCUR2Clear = useCallback(() => {
+    clearError();
+  }, [clearError]);
+
   // Determine current status
   const isLoading = isProcessing || reconstructionLoading;
 
+  // Render import method based on selection
+  const renderImportMethod = () => {
```
bucko13, line 256: this `renderComponent` pattern is kind of an anti-pattern from the older legacy code. This effectively is a signal that this should be written as its own component and used in the return part of this parent component. E.g.

```typescript
const ImportMethodSelector = (props) => { ... }
```

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -39,8 +44,234 @@ export const mapSignaturesToImporters = (signatureSets: SignatureSet[]) => {
...
+const addUtxoData = (
+  psbtInput: any,
+  multisig: any,
+  walletInput: Input,
+): void => {
+  if (psbtInput.witnessUtxo || psbtInput.nonWitnessUtxo) {
+    return; // Already has UTXO data
+  }
+
+  const { isP2WSH, isNestedSegwit } = getAddressTypeInfo(multisig.name);
```
bucko13, line 147: this shouldn't be necessary. The wallet config of the current wallet already has the script type loaded up. No need to check the name.

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -39,8 +44,234 @@ export const mapSignaturesToImporters = (signatureSets: SignatureSet[]) => {
...
+  isNestedSegwit:
+    multisigName &&
+    multisigName.includes("p2sh") &&
+    multisigName.includes("p2wsh"),
+});
+
+/**
+ * Adds scripts to a PSBT input using existing Caravan multisig data
+ */
+const addScriptsFromExistingMultisig = (
```
bucko13, line 89: @Legend101Zz didn't you have to add some very similar code as well?

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -39,8 +44,234 @@ export const mapSignaturesToImporters = (signatureSets: SignatureSet[]) => {
...
+  psbtInput: any,
+  multisig: any,
```
bucko13, line 91: would prefer to avoid `any` types if we can. psbtInput should be `LegacyInput` and there's also a `LegacyMultisig` type exported from one of the packages.

apps/coordinator/src/selectors/transaction.ts
```diff
@@ -39,8 +44,234 @@ export const mapSignaturesToImporters = (signatureSets: SignatureSet[]) => {
...
+    return;
+  }
+
+  // Strategy 2: Generate scripts from raw multisig data
+  if (hasRawMultisigData(walletInput)) {
+    generateScriptsFromRawData(psbtInput, multisig, index);
+    return;
+  }
+
+  // If neither strategy works, log and continue
```
bucko13, line 245: should this throw an error?

apps/coordinator/src/selectors/transaction.ts
bucko13, line 1: might be good to move a lot of these functions to `psbtUtils`.

apps/coordinator/src/actions/transactionActions.js
```diff
@@ -529,7 +545,12 @@ export function importPSBT(psbtText, inputs, hasPendingInputs) {
     //
     // For now, we skip extraction to avoid breaking flows that lack full UTXO context.
     if (!hasPendingInputs) {
-      dispatch(setSignaturesFromPsbt(psbt));
+      try {
+        dispatch(setSignaturesFromPsbt(psbt));
+      } catch (sigError) {
```
bucko13, line 550: why should we not just be throwing the error and let the component error boundaries handle? If there's an error importing signatures, I think we want to let the user know about these very clearly and not continue on with the bad PSBT.

bucko13 (changes requested):
Looking really good! I like the inclusion of the new import psbt method as well.

The biggest thing I have a question about is the signature importer. I'm hoping we can simplify that a bit by reusing some of the existing components. 

I also noticed some minor aesthetic things:
The stop scanning button here seems like it's disabled. I think we need different styling. 

Screenshot 2025-08-05 at 10 19 20 PM

Need some spacing around the button here: 

Screenshot 2025-08-05 at 10 19 07 PM

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
bucko13: btw, [here's](https://github.com/caravan-bitcoin/caravan/pull/324/files#diff-302d113fd31affcbc956bbc166b224091c563f32fc93962d5e4831b8eb51b279R83) an example in the new Jade PR that has to do a similar thing to support extracting signature array but it's done at the caravan-wallets level. I wonder if this would help simplify this code.

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
```diff
@@ -0,0 +1,186 @@
...
+          let signatureFound = false;
+
+          // Check if this input has a signature in the signed PSBT
+          if (i < signedPsbt.data.inputs.length) {
+            const signedInput = signedPsbt.data.inputs[i];
+
+            if (signedInput.partialSig && signedInput.partialSig.length > 0) {
+              // Extract the first signature for this input
+              const signature =
+                signedInput.partialSig[0].signature.toString("hex");
```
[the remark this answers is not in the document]
krrish-sehgal: Well , I thought the same thing initially as well,  that exporting a partially signed PSBT and then signing it again , should essentially give us the QR codes for the fully signed PSBT , but to my surprise , crypto-psbt was just dealing with 1 signature at the time, like when i tried to export the partially signed psbt to do a second sign , the first sign gets automatically removed here , and the newly signed PSBT just has the one new sig.

apps/coordinator/src/components/BCUR2/BCUR2ExtendedPublicKeyImporter.jsx
```diff
@@ -34,19 +25,54 @@ class BCUR2ExtendedPublicKeyImporter extends React.Component {
...
+          validateAndSetExtendedPublicKey(xpub, (error) => {
+            if (error) {
+              this.setError(error);
+            } else {
+              this.setError(""); // Clear any previous errors
+            }
+          });
+        },
+        (error) => {
+          this.setError(error);
```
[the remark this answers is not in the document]
bucko13: right, but they would've been already. By passing `this.setError` as the second arg it should be doing that no? basically your code does what was already happening as far as I can tell.

apps/coordinator/src/components/BCUR2/BCUR2ExtendedPublicKeyImporter.jsx
```diff
@@ -34,19 +25,54 @@ class BCUR2ExtendedPublicKeyImporter extends React.Component {
...
+        fingerprint = data.xfp || data.rootFingerprint || data.fingerprint;
+        path = data.path || data.bip32Path || data.derivationPath;
+      } else {
+        this.setError("Invalid QR code data format");
+        return;
+      }
 
+      // Validate required fields
+      if (!xpub) {
+        this.setError("QR code does not contain extended public key");
```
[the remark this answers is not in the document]
bucko13: Ok, i see what you're saying. I'm not sure why I was thinking this was more general. I do like that error message though!

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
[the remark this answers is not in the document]
bucko13: wow, very strange/interesting. I think this deserves a code comment then to explain just so we can keep track for the future. I still think that this is something that probably should be done in the interaction to provide a list of signatures as we have in the other interactions and then this importer can just be a wrapper around the indirect importer component (I think that would work)

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
krrish-sehgal: i addressed this in this [commit](https://github.com/caravan-bitcoin/caravan/pull/363/commits/879f3b9eb2b831746497b85d4207087fc57ff6c8).

apps/coordinator/src/components/BCUR2/BCUR2ExtendedPublicKeyImporter.jsx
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
```diff
@@ -0,0 +1,186 @@
+import React, { useState, useCallback, useRef, useEffect } from "react";
+import { Box, Alert, Button } from "@mui/material";
+import BCUR2Reader from "./BCUR2Reader";
+import { loadPsbt } from "../../utils/psbtUtils";
+import { useSelector } from "react-redux";
+
+interface SignatureImporter {
+  // Add specific properties if known, otherwise use a generic object
+  [key: string]: any;
```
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/Wallet/PSBTImportComponent.tsx
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const handleClearError = useCallback((): void => {
+    if (!isMountedRef.current) return;
+    setError("");
+    setShowScanner(false);
+  }, []);
+
+  const handlePSBTSuccess = useCallback(
+    (psbtData: string): void => {
+      // Early exit if component is unmounted
+      if (!isMountedRef.current) return;
```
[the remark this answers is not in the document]
krrish-sehgal: it didn't cause any issues, but it could, so added this. 
Outdated now.

krrish-sehgal:
> Looking really good! I like the inclusion of the new import psbt method as well.
> 
> The biggest thing I have a question about is the signature importer. I'm hoping we can simplify that a bit by reusing some of the existing components.
> 
> I also noticed some minor aesthetic things: The stop scanning button here seems like it's disabled. I think we need different styling.
> 
> Screenshot 2025-08-05 at 10 19 20 PM
> Need some spacing around the button here:
> 
> Screenshot 2025-08-05 at 10 19 07 PM

Screenshot 2025-08-08 at 12 23 38 AM

Fixed.

apps/coordinator/src/selectors/transaction.ts
krrish-sehgal: fixed.

apps/coordinator/src/selectors/transaction.ts
krrish-sehgal: fixed.

apps/coordinator/src/selectors/transaction.ts
krrish-sehgal: fixed.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
```diff
@@ -22,13 +22,28 @@ import InteractionMessages from "../InteractionMessages";
...
+      initialStatus = interaction.isSupported() ? PENDING : UNSUPPORTED;
+    } catch (error) {
+      initialStatus = PENDING; // Safe fallback
+    }
+
     this.state = {
       bip32PathError: "",
       signatureError: "",
-      status: this.interaction().isSupported() ? PENDING : UNSUPPORTED,
+      status: initialStatus,
```
krrish-sehgal, line 36: just a stylitic change, let me know if its not needed.

apps/coordinator/src/selectors/transaction.ts
krrish-sehgal: This shows an error , if I remove this, since its now moved in psbtUtils.

apps/coordinator/src/components/BCUR2/BCUR2Signer.jsx
```diff
@@ -0,0 +1,105 @@
...
+      // Error in handleClearError - silent fallback
+    }
+  }, [setError]);
+
+  const handlePSBTSuccess = useCallback(
+    (psbtData) => {
+      try {
+        // Validate inputs
+        if (!inputs || inputs.length === 0) {
+          if (setError) {
```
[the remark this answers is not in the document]
bucko13: Convert this to a `tsx` file, define the props as required and then remove the extraneous check.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
bucko13: seems ok. what determines support or not?

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
```diff
@@ -83,43 +98,78 @@ class IndirectSignatureImporter extends React.Component {
...
-      this.props;
-    const { signatureError, status } = this.state;
-    const interaction = this.interaction();
-    if (status === UNSUPPORTED) {
+    try {
+      const { disableChangeMethod, extendedPublicKeyImporter, Signer } =
+        this.props;
+      const { signatureError, status } = this.state;
+
+      if (!Signer) {
```
bucko13, line 106: this is a required prop, so why would this be missing?

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
```diff
@@ -83,43 +98,78 @@ class IndirectSignatureImporter extends React.Component {
...
+      } catch (error) {
+        return (
+          <div
+            style={{ padding: "16px", textAlign: "center", color: "#d32f2f" }}
+          >
+            <p>Error initializing signature interaction.</p>
+            <p>Please try refreshing the page.</p>
+          </div>
+        );
+      }
```
bucko13, line 128: you have the error catch boundary already above. Can we remove this.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
```diff
@@ -83,43 +98,78 @@ class IndirectSignatureImporter extends React.Component {
...
+              <InteractionMessages
+                messages={interaction.messagesFor({ state: status })}
+              />
+            </FormGroup>
+          </Box>
         </Box>
-      </Box>
-    );
+      );
+    } catch (error) {
```
bucko13, line 164: same

apps/coordinator/src/utils/psbtSignatureUtils.ts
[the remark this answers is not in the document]
bucko13: works for me!

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -73,6 +81,50 @@ export interface SignatureSet {
   publicKeys: string[];
 }
 
+/**
+ * Represents a bitcoinjs-lib PSBT input with the properties we use
+ */
+export interface BitcoinjsPsbtInput {
```
bucko13, line 87: couldn't we just import the type from bitcoinjs directly?

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -206,3 +258,466 @@ export const getInputIdentifiersFromPsbt = (psbt: Psbt): Set<string> => {
...
+    multisig.requiredSigners &&
+    multisig.totalSigners &&
+    multisig.publicKeys
+  );
+};
+
+/**
+ * Determines the address type from multisig name
+ */
+const getAddressTypeInfo = (multisigName: string) => ({
```
bucko13, line 304: we shouldn't be determining this from the name. we should eb determining this from the address type defined in the wallet config which gets applied to all slices generated. We should remove this so it's not relied upon.

apps/coordinator/src/utils/psbtUtils.ts
bucko13, line 1: would be great to have tests for these vital functions, but we didn't before so we can address later (maybe in a follow up...?)

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
```diff
@@ -0,0 +1,137 @@
...
+          Error details:{" "}
+          {error instanceof Error ? error.message : String(error)}
+        </p>
+        <p>Please try refreshing the page.</p>
+      </div>
+    );
+  }
+};
+
+BCUR2SignatureImporter.propTypes = {
```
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2Signer.jsx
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2Signer.jsx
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2Signer.jsx
```diff
@@ -0,0 +1,105 @@
...
+        // Extract signatures using the utility function
+        const signatures = extractSignaturesFromPSBTData(
+          psbtData,
+          inputs,
+          network,
+        );
+
+        if (signatures && signatures.length > 0) {
+          // Call onReceive with the signatures directly since we've already extracted them
+          if (onReceive && typeof onReceive === "function") {
```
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2Signer.jsx
```diff
@@ -0,0 +1,105 @@
...
+ * BCUR2 Signer component that follows the same pattern as ColdcardSigner.
+ * This component handles QR code scanning and PSBT processing for BCUR2 format.
+ */
+const BCUR2Signer = ({ setError, onReceive }) => {
+  // Wrap selectors in try-catch to handle any Redux state issues
+  let network, inputs;
+  try {
+    network = useSelector((state) => state.settings.network);
+    inputs = useSelector((state) => state.spend.transaction.inputs || []);
+  } catch (error) {
```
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/BCUR2/BCUR2Signer.jsx
```diff
@@ -0,0 +1,105 @@
...
+      if (setError) {
+        setError("");
+      }
+    } catch (error) {
+      // Error in handleClearError - silent fallback
+    }
+  }, [setError]);
+
+  const handlePSBTSuccess = useCallback(
+    (psbtData) => {
```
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
krrish-sehgal: done.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
krrish-sehgal: done.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
krrish-sehgal: done.

apps/coordinator/src/utils/psbtUtils.ts
krrish-sehgal: done.

apps/coordinator/src/utils/psbtUtils.ts
krrish-sehgal: sure , would do a follow up.

apps/coordinator/src/utils/psbtUtils.ts
krrish-sehgal: @bucko13 , I am getting this error with ur suggested implementation. 
`` "Error processing PSBT: Invalid signature for input 0, pubkey 02a0f713b43c7150..." ``

But it works fine with current implementation. Would try to reach deeper why is it happening,also would appreciate any thoughts.

thanks!

apps/coordinator/src/components/BCUR2/BCUR2SignatureImporter.tsx
```diff
@@ -0,0 +1,137 @@
...
+        <p>
+          Please ensure the transaction is properly set up before importing
+          signatures.
+        </p>
+      </div>
+    );
+  }
+
+  try {
+    return (
```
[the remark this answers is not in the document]
krrish-sehgal: done.

apps/coordinator/src/components/ScriptExplorer/IndirectSignatureImporter.jsx
krrish-sehgal: The support status you see is determined by the isSupported()  method of each keystore interaction class.

bucko13:
Great job so far! The code is looking great. I wanted to get ahead of some of the testing so wanted to try with an (old) Jade device and ended up running into some issues. I think @jevidon is going to help validate with a seedsigner as well. 

First is this issue where it couldn't find the signature.

Screenshot 2025-08-10 at 9 22 34 AM

When I tried exporting an xpub, then I was having trouble importing anything at all. I could get 20% read but it wouldn't go any further. 

One UX feedback I have that doesn't have to be blocking here but I think is worth fixing, is I found it confusing that the QR for signing was outside of the key signing element. I get why it was maybe done this way, but I think it's a little counter-intuitive, especially because the text just says "Export for signing" rather than something about the QR. I think it would be better to mirror what we have for coldcard where the signing flow includes a button to download the PSBT and then upload the signed version. We could maybe have two buttons, one to show QR for signing and then one to scan QR for importing (Show PSBT QR / Scan Signed PSBT). When one is active, it can be disabled and the other can be clickable. 

Screenshot 2025-08-10 at 9 35 07 AM

Coldcard PSBT flow:
Screenshot 2025-08-10 at 9 38 00 AM

One last thing is that I found with a device like Jade which has a lower quality camera, I needed to zoom in to the QR to be readable. Can we maybe double the QR code in size to make it more readable? 

Also, could we add a speed control so if you have a better camera (and a bigger transaction) it doesn't take too long to read?

krrish-sehgal:
Added speed controller:
Screenshot 2025-08-11 at 4 43 44 PM

Button Placement corrected: 
Screenshot 2025-08-11 at 5 23 17 PM

apps/coordinator/src/components/BCUR2/BCUR2Signer.tsx
```diff
@@ -0,0 +1,86 @@
...
+        setError(
+          `Error processing PSBT: ${error instanceof Error ? error.message : String(error)}`,
+        );
+      }
+    },
+    [inputs, network, onReceive, setError],
+  );
+
+  return (
+    <div>
```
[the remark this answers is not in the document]
krrish-sehgal: done.

jevidon:
Testing this again with the latest changes. Thanks for updating the button arrangement.

Some other feedback on the QR display modal:
- the size is much better, reads more reliably on my SeedSigner
- can the border around the QR be removed? the top portion is missing and the bottom just adds vertical height
- on a 14" laptop the controls and progress indicator are off screen, let's collapse those inside an Accordion titled "QR display controls"

Current:
image

Proposed:
image (6)

krrish-sehgal:
@bucko13 , just tested in the last vercel deployed version, The scanning progress bar is there.... seems ok to me.

bucko13:
Ah! you're right. I didn't see it outside of the box. What do you think about moving it in the same box as the video? I totally missed it when it was outside of the frame like that. Maybe that got moved out because of the "display controls"? 

Screenshot 2025-08-11 at 9 21 58 PM

krrish-sehgal:
> Code looks good! If we can move the QR code progress bar into the same frame as the QR itself then I think we're good to go!

fixed -> [preview](https://cdn.discordapp.com/attachments/1369080722102423615/1405052179148050464/Screenshot_2025-08-13_at_10.25.33_AM.png?ex=689d6c50&is=689c1ad0&hm=85ad91ef913b4f265b89ec737a375b43d0914a9c93caf0c4a23c68f848aafad3&)

bucko13 (approved):
Looks great! I'd like to get a final approval from either @jevidon or @jbrauck-unchained but I think we're good to merge.

jevidon:
Tested once again with Jade Plus and Seedsigner, both could scan the animated QR and Caravan could scan the signed version without issue. Excited to see this get merged!

PR #369, opened 2025-08-10
Change multisig

### LINKED ISSUE #368: PSBT doesn't respect mainnet config and doesn't encode wallet info
While testing I noticed that my SeedSigner was warning the spending conditions were such that all funds were going to an external wallet despite both the defined destination address and the change address were in the current wallet. Upon further investigation it became apparent that the exported PSBT did not include the wallet info, and was displaying the addresses as regtest instead of mainnet. This appears to be the case for both Coldcard PSBT file as well as the new bcur2 method currently in development.

bucko13:

This isn't the best way to resolve this but the state management around transactions is sufficiently complex that I think this "quick and dirty" approach is probably, lowest code complexity to resolve the issue.

As we move the code to depend more on the unsignedPsbt value in the state as the source of truth, we need to make sure that it's "final" value is correct. Unfortunately when we create that value, we don't actually have all the info about _outputs_ that we need since the "change slice" hasn't been identified yet. This happens during tx preview but by that time the unsignedPsbt is set. This would result in signers not being able to identify change outputs from the PSBT. 

This simply does another "finalizeOutputs" once the change slice has been identified and the data added and updates the unsignedPsbt. Again, not great, but it gets the job done. 

**Issue Number:**
Fixes #368 

**Summary**
To confirm, load up a wallet, create a transaction, download unsigned PSBT and copy the psbt into BIP370.org. previously the output would not have had derivation information (just address and amount), now it will for the change address. 

No.

## Checklist

**Other information**

### DIFF
--- a/.changeset/few-rockets-drive.md
+++ b/.changeset/few-rockets-drive.md
@@ -1,0 +1,5 @@
+---
+"caravan-coordinator": patch
+---
+
+Include bip32 derivation for change output in unsignedPsbt
--- a/apps/coordinator/src/components/Wallet/TransactionPreview.jsx
+++ b/apps/coordinator/src/components/Wallet/TransactionPreview.jsx
@@ -28,7 +28,10 @@ import {
 import UTXOSet from "../ScriptExplorer/UTXOSet";
 import { downloadFile } from "../../utils";
 import UnsignedTransaction from "../UnsignedTransaction";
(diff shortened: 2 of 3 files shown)


jevidon (approved):
Tested and found no issues getting the correct PSBT data to render.

Issue #382, opened 2025-09-17
Upload partially signed PSBT via file upload is not persisting signatures


jbrauck-unchained:
Recreation steps -

Try to upload a partially signed PSBT
Note you see all transaction details accurately
Then click "Sign transaction"
The signing screen is an empty state one asking for both signatures again.

We need to bring in the already signed pubkeys if we want this to work in a collaborative async manner.

PSBT used
```
[...]

PR #384, opened 2025-09-18
fixing an uplaod bug

jbrauck-unchained:

**Issue Number:**

Fixes https://github.com/caravan-bitcoin/caravan/issues/382

**Snapshots/Videos:**

**Summary**

## Checklist

**Other information**

### DIFF
--- a/apps/coordinator/src/utils/psbtUtils.ts
+++ b/apps/coordinator/src/utils/psbtUtils.ts
@@ -61,6 +61,7 @@ export interface SignatureInfo {
   pubkey: string;
   inputIndex: number;
   signerIndex?: number; // Which signer in the multisig (0, 1, 2, etc.)
+  masterFingerprint?: string; // XFP of the signer for stable identity across inputs
   derivedPubkey: string; // The actual derived pubkey for this input
 }
 
@@ -68,7 +69,7 @@ export interface SignatureInfo {
  * Interface for signer identification across multiple inputs
  */
 export interface SignerIdentity {
-  signerIndex: number; // Position in the multisig setup (0, 1, 2, etc.)
+  signerIndex?: number; // Position in the multisig setup (0, 1, 2, etc.)
   masterFingerprint?: string; // Master key fingerprint if available
   signatures: Buffer[] | string[]; // Signatures for each input (in order)
   publicKeys: string[]; // Derived pubkeys for each input (in order)
@@ -78,8 +79,8 @@ export interface SignerIdentity {
  * Interface for signature sets grouped by signer
  */
 export interface SignatureSet {
-  signatures: Buffer[] | string[];
-  publicKeys: string[];
+  signatures: (Buffer | string | null)[];
+  publicKeys: (string | null)[];
 }
 
 /**
@@ -249,12 +250,19 @@ export const getInputIdentifiersFromPsbt = (psbt: Psbt): Set<string> => {
  * @returns Array formatted for Caravan's signature importers
  */
 export const mapSignaturesToImporters = (signatureSets: SignatureSet[]) => {
-  return signatureSets.map((sigSet, index) => ({
-    importerNumber: index + 1, // As in Caravan we use 1-based indexing
-    signatures: sigSet.signatures,
-    publicKeys: sigSet.publicKeys,
-    finalized: true, // Mark as complete since we have all signatures
-  }));
+  return signatureSets.map((sigSet, index) => {
+    const signatures = (sigSet.signatures || []).map((sig: any) => {
+      if (!sig) return "";
+      return typeof sig === "string" ? sig : sig.toString("hex");
+    });
+    const publicKeys = (sigSet.publicKeys || []).map((pk: any) => pk || "");
+    return {
+      importerNumber: index + 1, // 1-based indexing in UI
+      signatures,
+      publicKeys,
+      finalized: true,
+    };
+  });
 };
 
 /**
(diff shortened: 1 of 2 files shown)


bucko13:
Things to validate:

- [ ] simple psbt (no sigs) upload works
- [ ] simple psbt (one key signed) upload works
- [ ] RBF psbt upload works (no sigs)
- [ ] RBF psbt upload works (some sigs)

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -639,61 +643,68 @@ export const extractSignaturesFromPSBT = (psbt: Psbt, inputs: Input[]) => {
...
-  }
+  if (inputSignatures.length === 0) return [];
 
   // Now we need to group signatures by signer identity (not individual pubkeys)
   const signerGroups = groupSignaturesBySigner(inputSignatures, inputs);
   const signatureSets: SignatureSet[] = [];
 
   signerGroups.forEach((signerGroup) => {
     signatureSets.push({
+      // Keep array shape aligned with inputs; allow nulls for missing
```
bucko13, line 707: errant comment

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -639,61 +643,68 @@ export const extractSignaturesFromPSBT = (psbt: Psbt, inputs: Input[]) => {
     const psbtInput = psbtWithScripts.data.inputs[inputIndex];
     const walletInput = inputs[inputIndex];
 
+    //
+
```
bucko13, line 647: cleanup

bucko13:
I found one edgecase that did not work:

- create an RBF transaction
- upload that PSBT
- sign and save the signed PSBT (e.g. using coldcard flow)
- abandon the transaction
- import the signed psbt. 
- note that the signature is not identified (this is is on a tx that is RBFing so its signatures on coins that aren't easily identifiable by our legacy system. This might be interfering with the sig identification) 

There's an easy workaround though which is that you can use the signed psbt in the normal signing flow. Also, worth noting that this isn't a regression as the same thing happens when I import the signed psbt in prod. 

CC @Legend101Zz

Legend101Zz (changes requested):
some initial comments on the `getInputIdentifiersFromPsbt` fn

apps/coordinator/src/utils/psbtSignatureUtils.ts
```diff
@@ -62,10 +62,10 @@ export function extractSignaturesFromPSBTData(
   const allSignatures: string[] = [];
 
   signatureSets.forEach((signatureSet) => {
-    // Convert Buffer signatures to hex strings if needed
-    const hexSignatures = signatureSet.signatures.map((sig) =>
-      typeof sig === "string" ? sig : sig.toString("hex"),
-    );
+    // Convert Buffer signatures to hex strings; skip nulls
```
Legend101Zz, line 65: note :
```ts
  // Extract signatures using existing robust utility
  const signatureSets = extractSignaturesFromPSBT(signedPsbt, inputs);
```
already does 
```ts
        const signature = partialSig.signature.toString("hex");
        const pubkeyFromPsbt = partialSig.pubkey.toString("hex");
```
so this can be reverted

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -61,14 +61,15 @@ export interface SignatureInfo {
...
+  masterFingerprint?: string; // XFP of the signer for stable identity across inputs
   derivedPubkey: string; // The actual derived pubkey for this input
 }
 
 /**
  * Interface for signer identification across multiple inputs
  */
 export interface SignerIdentity {
-  signerIndex: number; // Position in the multisig setup (0, 1, 2, etc.)
+  signerIndex?: number; // Position in the multisig setup (0, 1, 2, etc.)
```
Legend101Zz, line 72: why was this made optional ?

Legend101Zz:
> I found one edgecase that did not work:
> 
> * create an RBF transaction
> * upload that PSBT
> * sign and save the signed PSBT (e.g. using coldcard flow)
> * abandon the transaction
> * import the signed psbt.
> * note that the signature is not identified (this is is on a tx that is RBFing so its signatures on coins that aren't easily identifiable by our legacy system. This might be interfering with the sig identification)
> 
> There's an easy workaround though which is that you can use the signed psbt in the normal signing flow. Also, worth noting that this isn't a regression as the same thing happens when I import the signed psbt in prod.
> 
> CC @Legend101Zz

regarding this comment ... I founds @jbrauck-unchained  had PR earlier
#383 that did address RBF signing part as that had a clever solution to consider like :

```js
  // Use the inputs we just placed into Redux (may include reconstructed/RBF inputs)
    const resolvedInputs = state.spend?.transaction?.inputs || [];
```

but he closed that PR ... so does that look correct to you and should a follow up PR handle signing a RBF Psbt using this logic ?

apps/coordinator/src/utils/psbtUtils.ts
```diff
@@ -639,61 +643,68 @@ export const extractSignaturesFromPSBT = (psbt: Psbt, inputs: Input[]) => {
...
-            derivedPubkey: validatedPubkey,
-          });
-        }
+        const xfp = walletInput?.multisig?.bip32Derivation?.[0]
+          ?.masterFingerprint
+          ? walletInput.multisig.bip32Derivation[0].masterFingerprint.toString(
+              "hex",
+            )
+          : undefined;
+        // if signer index not found, we still persist the signature via xfp grouping when available
```
Legend101Zz, line 686: maybe we should fail if we are not able to find the fingerprint , as else we are just hoping that we dont have a tx where two or more inputs not having different addresses

Legend101Zz (approved):
so the logic and changes look good to me even checked the `xfp` code to see if it exists in the coded field and it looks good to me :) 
Screenshot 2025-09-29 at 18 23 48

apps/coordinator/src/utils/psbtUtils.ts
bucko13: looks like this is for typescript as well. looking upstream for where this data comes from, it comes from `SignatureInfo` which has signerIndex as optional

bucko13: and I think the idea is with this we're moving away from signer index where possible and preferring rootFingerprint for identifying sigs

apps/coordinator/src/utils/psbtUtils.ts
bucko13: I think we really just need to remove the signerIndex entirely and just fail if there are no root Fingerprints. Trying to do this locally and see if it works. I think we can always expect there to be derivations in the psbt for inputs which have a root fingerprint since signers need this so we should be good.

PR #390, opened 2025-10-03
feat: add QR for test suite wallet for easy transfer in different wallets

krrish-sehgal:
Screenshot 2025-10-03 at 11 32 03 PM


krrish-sehgal:
Yes , would be fixing it in a while... 
(was just messing around with address verfiication, and thought that this would be helpful).

bucko13 (dismissed):

bucko13 (approved):

PR #392, opened 2025-10-07
feat: add wallet export

krrish-sehgal:
This PR adds the next wallet export feature. 

I feel like I am lagging behind in the UI , and can still be improved. 

Screenshot 2025-10-13 at 10 43 29 PM
Screenshot 2025-10-13 at 10 43 39 PM
Screenshot 2025-10-13 at 10 44 00 PM
Screenshot 2025-10-13 at 10 44 04 PM

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,0 +1,15 @@
+{
+    "redirects": [
+      {
+        "source": "/:path*",
+        "has": [
+          {
+            "type": "host",
+            "value": "bip370.caravanmultisig.com"
+          }
+        ],
+        "destination": "https://bip370.org/",
+        "permanent": true
+      }
+    ]
+  }


krrish-sehgal:
> Note: when trying to export a wallet from my jade, it does animate so it should be possible to split up into "smaller" QRs.
> 
> I also got an invalid wallet config qr when trying to import. I suspect this is because it's expecting a caravan wallet config rather than the wallet descriptors like what the jade is providing.
> 
> Perhaps we want to add two options:
> 
> * wallet config json
> * wallet descriptors
> 
> What do you think?

Yup a choice between two would denitely make more sense, would add the descriptors one , once [this ](https://github.com/caravan-bitcoin/descriptors/pull/16) is merged.

PR #394, opened 2025-10-08
Transaction flow

### LINKED ISSUE #393: Transaction flow diagram
Many coordinators have employed visually stunning flow diagram to explain to users what is happening in their transaction. This is a nice touch that will make the app more education and improve the UI.

jbrauck-unchained:

This introduces a new tranasction flow diagram feature

**Issue Number:**

Fixes #393 https://github.com/caravan-bitcoin/caravan/issues/393

**Snapshots/Videos:**

Screenshot 2025-10-05 at 10 54 04 PM

**Summary**

I like what sparrow does to try and give a user a sense of understanding their own transaction. I think this is a small lift to gain feature parity in a place that makes caravan more educational.

## Checklist

**Other information**


apps/coordinator/src/components/Wallet/TransactionFlowDiagram.tsx
```diff
@@ -0,0 +1,1703 @@
...
+      feeBtc,
+      recipientOutputs,
+      changeOutputs,
+      inputCount: inputs.length,
+      outputCount: outputs.length,
+    };
+  }, [inputs, outputs, fee, changeAddress, inputsTotalSats]);
+
+  // Build smooth cubic-bezier path from (x1,y1) to (x2,y2)
+  const buildCurvePath = (x1: number, y1: number, x2: number, y2: number) => {
```
Legend101Zz: ok so this component seems to getting quite big , can we brak this down all into a TransactionPreview folder/directory and then have a utils , hooks and other components ( broken from this one ) there so that this file becomes manageable?

apps/coordinator/src/components/Wallet/TransactionFlowDiagram.tsx
```diff
@@ -0,0 +1,1703 @@
...
+        return theme.palette.success.light;
+      case "p2pkh":
+        return theme.palette.warning.light;
+      default:
+        return theme.palette.grey[500];
+    }
+  };
+
+  // Format script type for display
+  const formatScriptType = (scriptType?: string) => {
```
Legend101Zz: can you add this and formatAddress in the general utils file as they can be used by other components too :)

Legend101Zz (changes requested):
@jbrauck-unchained thanks for the awesome work , I tested it and it looks super good to me , now I guess we can move towards the code changes , the main concern I have is over modularity and an IIFE used in rendering of the flow status

Screenshot 2026-01-06 at 07 22 52

Legend101Zz (approved):
Thanks alot for the changes , it's looking really good now , nothing from my end now :) 

@bucko13 can you please review it when you get the chance ...

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/FlowDrawers.tsx
```diff
@@ -0,0 +1,356 @@
...
+        <Box
+          sx={{
+            p: 2,
+            display: "flex",
+            flexDirection: "column",
+            gap: 1.5,
+          }}
+        >
+          {inputs.map((input, idx) => {
+            const inputAmount = BigNumber(
```
bucko13, line 109: We should pull this out to be its own component and then just pass in the props.

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/FlowSummary.tsx
```diff
@@ -0,0 +1,350 @@
...
+        >
+          Input Dust Status:
+        </Typography>
+        <Box display="flex" gap={2} flexWrap="wrap">
+          <Box display="flex" alignItems="center" gap={0.5}>
+            <Chip
+              label="Economical"
+              color="success"
+              size="small"
+              sx={{ height: 20, fontSize: "0.7rem" }}
```
bucko13, line 118: seems like there's some room for code consolidation here. this styling is repeated on every box as far as I can tell and it should be generalized. If we change the look somehow you'll have to change every single line. do we want this to be a specialized kind of component?

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/index.tsx
```diff
@@ -0,0 +1,976 @@
...
+          })()}
+          <Box
+            sx={{
+              display: "flex",
+              flexDirection: "column",
+              gap: 1.5,
+            }}
+          >
+            {inputs.slice(0, 4).map((input, idx) => {
+              const inputAmount = BigNumber(
```
bucko13, line 307: another one that should be its own component to simplify this file a bit. and what can we do to clean up the inline styles?

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/index.tsx
```diff
@@ -0,0 +1,976 @@
...
+                >
+                  {`+${flowData.recipientOutputs.length - 4} more output${flowData.recipientOutputs.length - 4 > 1 ? "s" : ""}`}
+                </Typography>
+              </Box>
+            </Button>
+          )}
+
+          {/* Change Outputs */}
+          {flowData.changeOutputs.map((output, idx) => {
+            const amount = BigNumber(output.amount);
```
bucko13, line 763: modular component opportunity here.

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/index.tsx
bucko13, line 1: lots of simplification we can get here.

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Get script type color based on type
+ */
+export const getScriptTypeColor = (
+  scriptType?: string,
+  theme?: any,
+): string => {
+  if (!theme) return "#9e9e9e"; // fallback grey
+
+  switch (scriptType?.toLowerCase()) {
+    case "p2wsh":
```
bucko13, line 43: we have constants exported from caravan-bitcoin that we can use to standardize on for the script types. 

also note for outputs we can have a p2tr output

apps/coordinator/src/components/Wallet/TransactionFlowDiagram/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+export const getStatusDisplay = (
+  status?: string,
+  confirmations?: number,
+  theme?: any,
+) => {
+  if (!theme) return { label: "Unknown", color: "#9e9e9e" }; // fallback
+
+  switch (status) {
+    case "draft":
```
bucko13, line 70: status would be nice to have as a type-safe enum

apps/coordinator/src/components/Wallet/TransactionsTab/TableComponents/PendingTransactionsView.tsx
```diff
@@ -90,13 +93,16 @@ export const PendingTransactionsView: React.FC<Props> = ({
...
+            rawTransactions={pendingTransactions}
             onSort={handleSort}
             sortBy={sortBy}
             sortDirection={sortDirection}
             network={network}
+            walletAddresses={walletAddresses}
             onClickTransaction={onClickTransaction}
             onAccelerateTransaction={onAccelerateTransaction}
             showAcceleration={true}
+            expandable={true}
```
bucko13, line 105: cleaner way to do this is
```suggestion
            expandable
```

but also why do we need an option here if it's always expandable? are there some views where we don't expand it?

apps/coordinator/src/components/Wallet/TransactionsTab/TableComponents/TransactionsTable.tsx
```diff
@@ -252,151 +267,310 @@ const TransactionTableHeader: React.FC<{
...
   renderActions?: (tx: TransactionT) => React.ReactNode;
+  colSpan: number;
 }> = ({
   tx,
+  rawTx,
+  expandable,
+  isExpanded,
+  onToggleExpand,
   network,
+  walletAddresses = [],
```
bucko13, line 292: the component above uses a hook to get the addresses just to pass them down here. this component is then also using hooks to populate its internal state. Why not just have the hook for the wallet addresses here and simplify the function signature?

apps/coordinator/src/components/Wallet/TransactionsTab/TableComponents/TransactionsTable.tsx
```diff
@@ -422,12 +596,18 @@ export const TransactionTable: React.FC<TransactionTableProps> = ({
                 <TransactionTableRow
                   key={tx.txid}
                   tx={tx}
+                  rawTx={rawTxMap.get(tx.txid)}
                   showAcceleration={showAcceleration}
+                  expandable={expandable}
+                  isExpanded={expandedTxid === tx.txid}
+                  onToggleExpand={() => handleToggleExpand(tx.txid)}
                   network={network}
+                  walletAddresses={walletAddresses}
```
bucko13, line 605: actually we're prop drilling, passed through multiple levels. definitely doesn't need to do that.

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -193,34 +184,38 @@ class TransactionPreview extends React.Component {
     downloadFile(psbtData, "transaction.psbt");
   }
 
+  getTransactionStatus() {
```
bucko13, line 187: seems like a nice, standalone utility function. could make it type safe too so we know exactly what we expect it to return.

apps/coordinator/src/components/Wallet/TransactionPreview.jsx
```diff
@@ -193,34 +184,38 @@ class TransactionPreview extends React.Component {
...
+    const { signatureImporters, requiredSigners, broadcasting, txid } =
+      this.props;
+
+    if (broadcasting) return "broadcast-pending";
```
bucko13, line 191: why is `broadcasting` translate to `broadcast-pending`? is this actually that it's in the mempool? if so "broadcast pending" isn't correct.

bucko13: if not, maybe we call it "authoring"? if "Broadcasting" being true indicates it's being authored then I think this variable is mislabeled.

apps/coordinator/src/hooks/useTransactionDetails.ts
```diff
@@ -0,0 +1,100 @@
...
+  return useQuery(
+    [TRANSACTION_DETAILS_KEY, txid],
+    async () => {
+      if (!txid || !blockchainClient) return null;
+
+      // For public clients, fetch directly from the API to get raw data with prevout
+      if (blockchainClient.type === "public") {
+        try {
+          // Fetch raw transaction from mempool.space which includes prevout
+          const response = await blockchainClient.Get(`/tx/${txid}`);
```
bucko13, line 32: something funky going on here. the whole point of the blockchain client and having its own package `@caravan/clients` is to add an abstraction layer so that this type of code doesn't need to be written. If the package supports these requests, then use the existing methods. if not, then let's add support there.

apps/coordinator/src/hooks/useTransactionDetails.ts
```diff
@@ -0,0 +1,100 @@
...
+ * the overhead of fetching prevout data for all transactions in the list.
+ */
+export const useTransactionDetails = (
+  txid: string | null,
+  enabled: boolean,
+) => {
+  const blockchainClient = useGetClient();
+
+  return useQuery(
+    [TRANSACTION_DETAILS_KEY, txid],
```
bucko13, line 24: see some of the other hooks for how we setup the query cache keys.

apps/coordinator/src/reducers/transactionReducer.js
```diff
@@ -455,6 +457,8 @@ export default (state = initialState(), action) => {
       return updateState(state);
     case SET_TXID:
       return updateState(state, { txid: action.value });
+    case SET_BROADCASTING:
```
bucko13, line 460: now i'm curious what this is actually indicating. also wonder if there's a way we can avoid adding to the redux store. It's hard in caravan since we have a lot of tightly coupled legacy code, but if we can avoid adding to the redux store that would definitely be the preference.

apps/coordinator/src/utils/transactionFlowUtils.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+export interface FlowDiagramProps {
+  inputs: FlowDiagramInput[];
+  outputs: FlowDiagramOutput[];
+  fee: string; // in BTC
+  changeAddress?: string;
+  inputsTotalSats: BigNumber;
+  status:
+    | "draft"
```
bucko13, line 39: see we're using these values in other places. what if we made this a shareable type and then could make all uses of it type safe.

apps/coordinator/src/utils/transactionFlowUtils.ts
```diff
@@ -0,0 +1,266 @@
...
+    | "conflicted"
+    | "rejected"
+    | "unknown";
+  confirmations?: number;
+}
+
+/**
+ * Detect script type from address prefix (heuristic)
+ */
+const detectScriptTypeFromAddress = (address?: string): string | undefined => {
```
bucko13, line 57: we have address type detection tools in caravan-bitcoin

apps/coordinator/src/utils/transactionFlowUtils.ts
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * Transform a raw transaction from the blockchain client into
+ * props suitable for the TransactionFlowDiagram component.
+ *
+ * @param tx - Raw transaction from blockchain client
+ * @param walletAddresses - Array of wallet addresses to identify change outputs
+ * @returns FlowDiagramProps ready for the component
+ */
+export const transformTransactionToFlowDiagram = (
+  tx: any,
```
bucko13, line 98: I think we have some types for this don't we @Legend101Zz ?

bucko13 (commented):
I don't love all of the inline stylings that we have throughout this. it will make the aesthetics come out inconsistent with the rest of the app and also harder to maintain. we should be able to use existing theming or components and/or update to the theme.

few other opportunities for code consolidation/cleanup. Also would like to add type safety where possible.

jevidon:
Overall really solid. Just a couple nits I noticed, but not blocking.

image

PR #401, opened 2025-10-25
feat:download signed psbt

jbrauck-unchained:

**Issue Number:**

Fixes #

**Snapshots/Videos:**

**Summary**

## Checklist

**Other information**

### DIFF
--- a/.changeset/vast-emus-shake.md
+++ b/.changeset/vast-emus-shake.md
@@ -1,0 +1,5 @@
+---
+"caravan-coordinator": minor
+---
+
+add download psbt with signatures functionality on a per key basis and for full tx. Also fixes bug on psbt importer where multiple keys signed for are not imported (only one sig being imported)


bucko13 (commented):
I'm having trouble confirming this against unchained's system but that could be just how we combine and obfuscate PSBTs. I confirmed the signed PSBT in bip370.org though and it seems to be fine on the surface at least. I also re-imported to caravan and it had the signature. 

I wonder if actually maybe a better way to accomplish this though is to store signed psbts on the signature importer object in redux. So whenever you have signatures, you make sure there's a psbt version of that and that can be downloaded. Then also when we get signatures from devices, we can actually keep the psbt version rather than just extracting the individual signatures. 

I'm not against this approach just wondering if that would be a bit more "psbt-native"

bucko13 (approved):

PR #408, opened 2025-11-04
Fix jade tests

bucko13:
I was finding some bugs and bad formatting in the jade tests.


packages/caravan-wallets/src/jade.test.ts
```diff
@@ -7,14 +7,10 @@ import {
   PsbtV2,
   bip32PathToSequence,
   bip32SequenceToPath,
-} from '@caravan/bitcoin';
```
Legend101Zz: hmm yes there is some error as when I hit cmd save in my IDE I see it reformat this too  but in my case the formatting happens correctly ...  ans I guess the jade contributor that worked on this did have some node version and conflict issues as far as I remember, when we were reviewing that PR , so maybe editor (VS Code, etc.)  using its own Prettier settings or a different Prettier config than what's in the repo, so when you save, it's formatting with different rules (like not enforcing double quotes).

Like I ran npm run format on a pull of latest main and this fixes that too so I guess maybe update root ci as `  "ci": "npm run format:check && turbo run build lint test", or even better would be to add this in Pre-commit Hook

Legend101Zz (approved):
Looks good to me just curious what you think about the ci command and pre commit hook changes to handle this to prevent something like this happening again ?

PR #414, opened 2025-11-23
Fix: Documentation Deployment Failing Due to Git Hook Errors

Legend101Zz:
While working on the documentation publishing workflow, I tracked down an issue where the process was consistently failing during the commit step on the `docs` orphan branch.

### **What was actually happening**

Our workflow does this:

1. Checks out `main`, which includes all Husky git hooks
2. Builds the project and generates the Typedoc output
3. Creates an orphan `docs` branch and runs `git rm -rf .` to wipe everything
4. Copies the generated documentation back in
5. Tries to commit with `git commit -m "chore(docs): update typedocs [skip ci]"`

The failure happens at step **5**.

### **The root cause**

Even though we deleted all files from the working directory (including the `.husky` folder), the **git hooks themselves still remain active** because Git stores them in `.git/hooks/`.

So when the commit runs, Git automatically triggers Husky's `prepare-commit-msg` hook. That hook tries to execute:

```
.husky/_/h
```

But since we wiped the working directory earlier, that file no longer exists.

As a result, the hook throws an error → the commit fails → the workflow stops.

### **Why this happens**

* Husky installs its hooks into `.git/hooks/`
* The workflow removes the Husky helper scripts from the working tree
* The hooks inside `.git/hooks/` still point to those removed files
* Git tries to run them anyway → boom gh action fails

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,6 +38,12 @@ jobs:
           git config --global user.name 'github-actions[bot]'
           git config --global user.email 'github-actions[bot]@users.noreply.github.com'
 
+          # Verify docs were generated
+          if [ ! -d "docs" ] || [ -z "$(ls -A docs)" ]; then
+            echo "Error: docs directory is empty or doesn't exist"
+            exit 1
+          fi
+
           # Save the generated docs to a temp location
           mv docs docs_temp
 


bucko13 (approved):
🙏

PR #419, opened 2025-12-15
fix: make RBF partial sig import work

Legend101Zz:
### Description

This PR resolves an issue where signature extraction was skipped or failed for RBF (Replace-By-Fee) PSBTs. It decouples the signature extraction logic from "spendable wallet slices," allowing Caravan to import signatures for any valid transaction inputs—whether they are confirmed UTXOs or reconstructed inputs from pending transactions.

### The Problem

Previously, the `importPSBT` workflow skipped signature extraction if `hasPendingInputs` was true. This was a necessary workaround because the downstream function, `setSignaturesFromPsbt`, relied on the `selectAvailableInputsFromPSBT` selector .

That selector **only** queries "spendable" wallet slices (UTXOs visible in `listunspent`). Since RBF inputs are technically "consumed" by the pending transaction they are replacing, they do not appear in spendable slices. Consequently, attempting to extract signatures for RBF PSBTs would fail to find the necessary input metadata (scripts, paths), leading to errors.

### The Solution

This PR refactors `setSignaturesFromPsbt` to use **Redux state** as the source of truth for inputs, rather than re-querying selectors, thanks to the idea that @jbrauck-unchained provided :)

1. **Source Agnostic:** By the time `setSignaturesFromPsbt` is called, `importPSBT` has already dispatched `setInputs(resolvedInputs)`. The Redux state (`state.spend.transaction.inputs`) contains the fully resolved inputs, regardless of whether they came from the standard UTXO set or were reconstructed from pending RBF data.

2. **Removal of Guard Clause:** With the extraction logic now robust enough to handle RBF inputs, the `if (!hasPendingInputs)` check in `importPSBT` has been removed. Signature extraction now runs unconditionally for all imports.

### Verification**Scenario 1: Standard PSBT Import**

1. Import a partially signed PSBT using standard confirmed UTXOs.
2. Verify that signatures are extracted and populated in the UI.

Screenshot 2025-12-15 at 17 44 32

**Scenario 2: RBF PSBT Import**

1. Create a transaction and broadcast it.
2. Create a generic RBF replacement for that transaction (consuming the same inputs).
3. Partially sign the RBF PSBT externally.
4. Import the RBF PSBT into Caravan.
5. **Result:** Signatures should now be successfully extracted and displayed, whereas previously they were ignored.

Screenshot 2025-12-15 at 17 45 08

**Left : To test BCUR sig import**

### DIFF
--- a/.changeset/witty-kiwis-occur.md
+++ b/.changeset/witty-kiwis-occur.md
@@ -1,0 +1,5 @@
+---
+"caravan-coordinator": patch
+---
+
+fix importing of rbf psbts with signatures
--- a/apps/coordinator/src/selectors/transaction.ts
+++ b/apps/coordinator/src/selectors/transaction.ts
@@ -8,8 +8,6 @@ import {
   reverseTxidEndianness,
   getSequenceForInput,
   getInputIdentifiersFromPsbt,
-  mapSignaturesToImporters,
-  extractSignaturesFromPSBT,
   createInputIdentifier,
 } from "utils/psbtUtils";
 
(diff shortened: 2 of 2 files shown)


apps/coordinator/src/actions/transactionActions.js
bucko13: lovely cleanup

PR #425, opened 2025-12-19
Resolving multiple xpubs patch1

### LINKED ISSUE #389: Trezor suite confirmation of xpub export includes multiple xpubs
Confusingly, when an xpub export happens from trezor, two xpubs are shown to the user. This is documented in [this issue](https://github.com/trezor/trezor-suite/issues/21278) for trezor. The reason for this is how we had to previously handle requests for xpub export that needed to include the root fingerprint. For this, we added code to request a "bundle" export [here](https://github.com/caravan-bitcoin/caravan/blob/main/packages/caravan-wallets/src/trezor.ts#L550). This results in the multiple key request. The reason for this is to request the extended key at one depth below the root and then use the serialized parent fingerprint to extract [here](https://github.com/caravan-bitcoin/caravan/blob/main/packages/caravan-wallets/src/trezor.ts#L528). 

It turns out that newer trezor devices return a descriptor when exporting an extended public key, see the docs [here](https://connect.trezor.io/9/methods/bitcoin/getPublicKey/#get-public-key), which has the root fingerprint serialized. This is not supported on the trezor ones but would solve the issue for other devices. 

we could optionally check for the descriptor in the return object and parse/deserialize that descriptor and if it's not available fallback to the bad UX behavior we have currently.

TomLucasakaTGeek:
The additional code snippets provide a way to find descriptor carried by the payload, and based on the prescence of descriptor serialize the descriptor into => FingerPrint + Keys  

**Issue Number:**

Fixes #389 

**Snapshots Attached:**
image
image

Probably Not required

**Summary**

[Issue:](https://github.com/caravan-bitcoin/caravan/issues/389)

Probably Not

Not Applicable

## Checklist
- [x] I have verified that test coverage meets or exceeds 95% (if applicable). 

### DIFF
--- a/packages/caravan-wallets/src/trezor.ts
+++ b/packages/caravan-wallets/src/trezor.ts
@@ -592,17 +592,17 @@ export class TrezorExportPublicKey extends TrezorExportHDNode {
    * On newer firmware, prefer descriptor parsing.
    */
   parsePayload(payload) {
-    // Newer Trezor firmware: descriptor present 
-    if ( payload.descriptor ) {
-      const parsed = this.ParseDescriptor(payload.descriptor);
-      return {
-        rootFingerprint: parsed.rootFingerprint,
-        pubkey: parsed.xpub,
-      }
-    }
+    // ✅ Newer firmware: descriptor present 
+    if (payload.descriptor) { 
+      const parsed = this.parseDescriptor(payload.descriptor); 
+      return { 
+        rootFingerprint: parsed.rootFingerprint, 
+        publicKey: parsed.xpub, 
+      }; 
+    } 
 
-    // Older Trezor firmware: no descriptor
-    if ( this.includeXFP && Array.isArray(payload) ) {
+    // ❌ Older firmware: bundle response
+    if (this.includeXFP) {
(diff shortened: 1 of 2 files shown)


packages/caravan-wallets/src/trezor.ts
```diff
@@ -602,6 +614,30 @@ export class TrezorExportPublicKey extends TrezorExportHDNode {
...
+   */ 
+  parseDescriptor(descriptor: string) { 
+    let rootFingerprint = ""; let xpub = ""; 
+    // Extract fingerprint 
+    // (inside square brackets before the first slash) 
+    const fpMatch = descriptor.match(/ \[([0-9a-fA-F]+)\//); 
+    if (fpMatch) { rootFingerprint = fpMatch[1]; } 
+    
+    // Extract xpub (the base58 string starting with "xpub" or "tpub") 
+    const xpubMatch = descriptor.match(/(xpub[0-9A-Za-z]+|tpub[0-9A-Za-z]+)/);
```
bucko13: we have xpub parsing in @caravan/bitcoin in the keys.ts file that we can use to have more robust checking.

bucko13 (changes requested):

PR #435, opened 2025-12-30
docs: add sortedmulti public key ordering documentation

bensig:
## Summary

Clarifies that `generateMultisigFromPublicKeys` uses public keys in the order passed, and that callers must sort keys lexicographically for `sortedmulti` compatibility.

This is a common gotcha - if you don't sort pubkeys before calling this function, you'll generate a valid multisig address but it will be **different** from the address generated by hardware wallets (Ledger, Trezor) and other software expecting `sortedmulti`.

## Changes

- Added warning callout to README with example code
- Added JSDoc comment to `generateMultisigFromPublicKeys` function
- Fixed typo: `javasacript` → `javascript` in README code block

## Test plan

- [x] Ran `turbo run test --filter=@caravan/bitcoin` - all 633 tests pass
- Documentation-only change, no functional changes


bucko13 (approved):
thank you!

PR #441, opened 2026-01-01
CPFP Fee Calculation Bug Fix and UI Improvements

### LINKED ISSUE #424: Adjust language in CPFP flow
Please see attached images below, and ignore the red boxes.

My original parent transaction was ~5 sat/vb, and the first image suggests the current high fee rate is ~33 sat/vb. Therefore the child transaction needs to be ~68 sat/vb to result in an effective fee rate of ~33 sat/vb (I'm assuming this is the target) for both transactions.

That was my first interpretation.

However things got confusing on the second image. It says in two separate places that the "target combined fee rate" is ~68 sat/vb. That doesn't make sense, if ~68 sat/vb is what is being used for the child transaction only. Yet the math 23,445 sats ÷ 360 vb does check out. 

So I'm seeing a couple issues:
1) Why is the "target combined fee rate" (which I would suggest renaming to "target effective fee rate" to match verbiage on mempool.space) double the high fee rate of 33 sat/vb? I would think it should be barely above it, like 35 sat/vb or 40 sat/vb.

2) I think we should also be explicitly displaying the fee rate being used for the child transaction. This will, after all, be the fee rate someone sees after loading up the PSBT file that was created by this feature. For this transaction I calculate the child transaction used 147.5 sat/vb (23455 sats ÷ 159 vb).

Image

Image

Legend101Zz:
### Critical Bug Fixed

FIxes #424 

Fixed the `minimumCPFPFee` calculation bug identified by @tomunchained  that was causing users to overpay by alot ... 

**The Issue:**
- `cpfpFeeRate` correctly calculated the child transaction's required fee rate (e.g., 68 sat/vB)
- `minimumCPFPFee` then incorrectly multiplied this by `CPFPPackageSize` (parent + child) instead of just `estimatedCPFPChildSize`
- Result: Child paid fees for the entire package at target rate, doubling the effective fee rate
  - Expected: 33 sat/vB combined effective rate
  - Actual: 68 sat/vB combined effective rate

**The Fix:**
Changed `minimumCPFPFee` to multiply `cpfpFeeRate` by `estimatedCPFPChildSize` only, ensuring the child pays just the additional fee needed to reach the target package rate.

### UI Improvements

**Strategy Selection Screen:**
- Added child-specific labels for CPFP ("Child transaction fee" vs generic "Minimum fee required")
- Added tooltip explaining CPFP formula: `(TargetRate × TotalSize - ParentFee) / ChildSize`
- Shows target fee rate is set to current network medium rate
- Caption explains combined effective rate goal

**CPFP Configuration Form:**
- Slider now controls child transaction fee rate directly (not combined rate)
- Displays three distinct rates:
  - Parent fee rate (existing transaction)
  - Child fee rate (new transaction being created)
  - Combined effective rate (package rate miners evaluate)
- Updated slider label from "Target Combined Fee Rate" to "Child Transaction Fee Rate"
- Minimum slider value starts at calculated minimum child fee rate from `cpfp.feeRate`

**Credit:** Thanks to @tomunchained for identifying and documenting the bug, and for the detailed analysis and solution :)

Screenshot 2026-01-01 at 19 34 51
Screenshot 2026-01-01 at 19 35 17

Screenshot 2026-01-01 at 22 34 17

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   },
   "scripts": {
     "preinstall": "node scripts/check-node-versions.js",
-    "build": "turbo build",
-    "build:coordinator": "turbo build --filter=caravan-coordinator",
+    "build": "NODE_OPTIONS='--max-old-space-size=8192' turbo build",
+    "build:coordinator": "NODE_OPTIONS='--max-old-space-size=8192' turbo build --filter=caravan-coordinator",
     "changeset": "changeset",
     "ci": "turbo run build lint test",
     "clean": "rm -rf .turbo && rm -rf **/dist && rm -rf **/build && rm -rf **/.turbo",


apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/ConfigurationStep/CPFP/CPFPForm.tsx
```diff
@@ -246,20 +265,23 @@ export const CPFPForm: React.FC = () => {
...
-          minimumFeeRate={minimumFeeRate}
+          minimumFeeRate={minimumChildFeeRate}
           maxFeeRate={maxFeeRate}
           feeEstimates={feeEstimates || {}}
         />
 
         <Box sx={{ mt: 2 }}>
           <Typography variant="body2" color="text.secondary">
-            Parent fee rate: {originalFeeRate.toFixed(1)} sats/vB
+            Parent fee rate: {originalFeeRate.toFixed(2)} sats/vB
```
bucko13: was the `toFixed(1)` causing issues.

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/StrategySelectionStep/StrategyRadioGroup.tsx
```diff
@@ -173,24 +176,112 @@ export const StrategyRadioGroup: React.FC<StrategyRadioGroupProps> = ({
...
-                        </Typography>
-                      </Grid>
-                      <Grid item xs={6}>
-                        <Typography variant="body2" color="text.secondary">
-                          Suggested fee rate:
-                        </Typography>
-                        <Typography variant="body2">
-                          {config.suggestedFeeRate.toFixed(2)} sats/vB
+                    {config.strategy === "CPFP" ? (
+                      <>
```
bucko13: can we pull these out into their own component? I think it'll help make this a little more readable.

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/hooks.ts
```diff
@@ -129,6 +134,7 @@ export const useAnalyzeTransaction = (
           feeRate: analyzer.cpfpFeeRate,
           childSize: analyzer.estimatedCPFPChildSize,
           estimatedPackageSize: analyzer.CPFPPackageSize,
+          targetFeeRate: targetFeeRate,
```
bucko13:
```suggestion
          targetFeeRate,
```

.changeset/busy-poems-sit.md
```diff
@@ -0,0 +1,6 @@
+---
+"@caravan/transactions": patch
```
bucko13: I haven't published yet so no need to add this. (I really should though and publish it as a v1)

packages/transactions/src/transactionAnalyzer.ts
```diff
@@ -394,8 +396,10 @@ export class TransactionAnalyzer {
    *                     transaction's fee is already sufficient for the desired rate.
    */
   get minimumCPFPFee(): Satoshis {
+    // so we want to know the extra fees that child has to pay,
+    // remember when miner takes in this tx ( along with the parent) they are rewarded minimumCPFPFee + ParentFee
```
bucko13:
```suggestion
    // remember when miner takes in this tx (along with the parent) they are rewarded minimumCPFPFee + ParentFee
```

bucko13 (changes requested):
Nothing crazy in my comments. Mostly nits. Thanks for the fixes!

Legend101Zz:
> Lastly, super small nit, but the PSBT I downloaded has a fee that is off by one satoshi. 11584 instead of 11583. It might just be some type of rounding issue. Probably not worth spending much time on unless it can be identified and fixed super easliy.

Thanks a ton for testing it out ... I'll make the UI changes you suggested for the last ( Review Step) page as you suggested :)

> Lastly, super small nit, but the PSBT I downloaded has a fee that is off by one satoshi. 11584 instead of 11583. It might just be some type of rounding issue. Probably not worth spending much time on unless it can be identified and fixed super easliy.


Ahh, good catch ... 
This is actually a fairly hard thing to avoid, and in this case `11584` is technically more accurate than the `11583` shown in the UI.

What’s happening is that the UI performs a few rounding steps when displaying fee-related values, based on the data we get from the `caravan/transaction` package (especially around fee bumping). Those rounded values are what end up being surfaced to the user.

Under the hood though, what we ultimately pass to the `caravan/transaction` package is just the final `targetFeeRate` selected by the user. The actual fee calculation and PSBT construction then happen downstream with full precision, which is why you can occasionally see a ±1 sat difference.

So you’re right—it’s essentially a rounding / display artifact on the UI side  ... honestly let me check if I can have it fixed by using `BigNumber` lib operations instead of normal JS ops ( on the UI side(

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/StrategySelectionStep/StrategyRadioGroup.tsx
Legend101Zz: sure , done https://github.com/caravan-bitcoin/caravan/pull/441/commits/74e838a6cca476c3b220ed4dac1749008c991bfb :)

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/hooks.ts
```diff
@@ -111,9 +118,7 @@ export const useAnalyzeTransaction = (
       const analyzer = new TransactionAnalyzer({
         txHex,
         network: network as Network,
-        targetFeeRate:
-          feeEstimates[FeePriority.MEDIUM] ??
-          (transaction.vsize ? transaction.fee / transaction.vsize : 250),
+        targetFeeRate: targetFeeRate,
```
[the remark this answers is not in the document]
Legend101Zz: done https://github.com/caravan-bitcoin/caravan/pull/441/commits/daf47b5dbba843deb48b96e949a9515a6a9ca6d1

.changeset/busy-poems-sit.md
Legend101Zz: done https://github.com/caravan-bitcoin/caravan/pull/441/commits/07d6db75b8c68865fd4a1e9344f384f7b437600e

bucko13 (dismissed):
Code approved. Will wait to merge after @tomunchained gives a final thumbs up.

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/ReviewStep/TransactionComparison.tsx
```diff
@@ -17,7 +17,7 @@ import { FeeBumpStrategy } from "@caravan/transactions";
 import { useAccelerationModal } from "../../AccelerationModalContext";
 import { RBF_TYPES } from "../../../types";
 
-import { TransactionComparisonTable } from "./TransactionComparisonTable ";
```
Legend101Zz: seems like the file name had a space char init , so I renamed that file :)

bucko13:
I think [this](https://github.com/caravan-bitcoin/caravan/pull/441/files#diff-b44d0cad60eb562944a672746bb59f2a62fbe4a2d92c0c412a5bf7dd09598309R104) is where the confusion is coming from. The column name in code is `difference`, but the column title is `Package Total`. I think you probably just want to change [this line](https://github.com/caravan-bitcoin/caravan/pull/441/files#diff-b44d0cad60eb562944a672746bb59f2a62fbe4a2d92c0c412a5bf7dd09598309R268) so that it's `<strong>{isCPFP ?  : "Difference": "Package Total"}</strong>` and then have the `fees` row also show a difference instead of total OR you can flip it and show `Package Total`, remove `Total:` [here](https://github.com/caravan-bitcoin/caravan/pull/441/files#diff-b44d0cad60eb562944a672746bb59f2a62fbe4a2d92c0c412a5bf7dd09598309R112), and then recalculate the fee rate row value to calculate the total fee rate (which should be `(fee_A + fee_B) / (pk_A_size + pk_B_size)`

Legend101Zz:
> I think [this](https://github.com/caravan-bitcoin/caravan/pull/441/files#diff-b44d0cad60eb562944a672746bb59f2a62fbe4a2d92c0c412a5bf7dd09598309R104) is where the confusion is coming from. The column name in code is `difference`, but the column title is `Package Total`. I think you probably just want to change [this line](https://github.com/caravan-bitcoin/caravan/pull/441/files#diff-b44d0cad60eb562944a672746bb59f2a62fbe4a2d92c0c412a5bf7dd09598309R268) so that it's `\<strong\>{isCPFP ? : "Difference": "Package Total"}\</strong\>` and then have the `fees` row also show a difference instead of total OR you can flip it and show `Package Total`, remove `Total:` [here](https://github.com/caravan-bitcoin/caravan/pull/441/files#diff-b44d0cad60eb562944a672746bb59f2a62fbe4a2d92c0c412a5bf7dd09598309R112), and then recalculate the fee rate row value to calculate the total fee rate (which should be `(fee_A + fee_B) / (pk_A_size + pk_B_size)`

Yup calculation like this :) , let me try

apps/coordinator/src/components/Wallet/TransactionsTab/FeeBumping/components/FeeBumpSteps/ReviewStep/TransactionComparisonTable.tsx
```diff
@@ -124,7 +131,33 @@ export const TransactionComparisonTable: React.FC<TransactionComparisonTableProp
...
+
+                    const childFee = packageFee.minus(parentFee);
+                    const childFeeRate = new BigNumber(
+                      feeBumpResult.newFeeRate,
+                    );
+
+                    // size = fee / fee_rate
+                    const parentSize = parentFee.dividedBy(parentFeeRate);
+                    const childSize = childFee.dividedBy(childFeeRate);
+                    console.log(
```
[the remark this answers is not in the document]
Legend101Zz: changes done https://github.com/caravan-bitcoin/caravan/pull/441/commits/7d82c05735bb7e427bcbb52c800c454b094f8812 :)

PR #471, opened 2026-02-08
fix(coordinator): track pending transactions without change outputs u…

### LINKED ISSUE #407: Broken PendingTx Logic: getPendingTransactionIds fails to track transactions without change outputs (1:1, 2:2, etc.)
## Title
`getPendingTransactionIds` fails to track transactions without change outputs (1:1, 2:2, etc.)

---

## Problem Description

The wallet's pending transaction tracking is broken for transactions that fully consume their inputs without creating a change output. This happens commonly in scenarios like:
- Sending exact amounts (1:1 transactions)
- Consolidating UTXOs (2:2, 3:1 transactions)
- Any transaction where inputs are completely consumed

**Current behavior:**
- ✅ Pending transactions WITH change outputs are tracked correctly
- ✅ Transactions to self (within wallet) are tracked correctly
- ❌ Transactions without change outputs disappear from pending view immediately after broadcast

**Example:**
```
User broadcasts a transaction:
- Input: 1 BTC (from address A)
- Output: 1 BTC (to external address B)
- No change output

Result: Transaction vanishes from pending transactions tab
```

---

## Root Cause

The `getPendingTransactionIds` selector only looks for pending transactions by scanning **unconfirmed UTXOs** in the wallet:

```typescript
export const getPendingTransactionIds = createSelector(
  getWalletSlices,
  (slices: Slice[]): string[] => {
    const txids = new Set<string>();

    slices.forEach((slice: Slice) => {
      if (slice.utxos && slice.utxos.length > 0) {
        slice.utxos.forEach((utxo: UTXO) => {
          // Only finds txids from unconfirmed UTXOs
          if (utxo.txid && !utxo.confirmed) {
            txids.add(utxo.txid);
          }
        });
      }
    });

    return Array.from(txids);
  },
);
```

**The problem:**
When a transaction has no change output:
1. Input UTXOs are spent (removed from wallet state)
2. Output goes to external address (not in wallet)
3. **No unconfirmed UTXOs remain in wallet to track the transaction**
4. Transaction becomes "invisible" despite being pending

This only worked before by coincidence - when there was a change output, that unconfirmed change UTXO would mark the transaction as pending.

---

## Proposed Solution

We already have the infrastructure to solve this! The blockchain client methods (`getAddressTransactionHistory` and `getWalletTransactionHistory`) impleted by @ice-009 fetch **all transactions** - both confirmed and pending. We currently filter these by `"confirmed"` status, but we can create parallel hooks that filter by `"pending"` status instead.

### Implementation Plan

**1. Create new pending transaction hooks** (mirror the existing confirmed transaction hooks):

```typescript
// New hooks in your transactions file

export const usePublicClientPendingTransactions = () => {
  const blockchainClient = useGetClient();
  const walletAddresses = useSelector(getWalletAddresses);
  const spentSlices = useSelector(getSpentSlices);
  
  // ... same setup as confirmed transactions hook ...

  return useQuery({
    queryKey: transactionKeys.pendingHistory(),
    queryFn: async () => {
      const rawTransactions =
        await blockchainClient.getAddressTransactionHistory(
          currentAddresses,
          MAX_TRANSACTIONS_TO_FETCH,
          0,
        );

      // Key change: filter for "pending" instead of "confirmed"
      const pendingTx = selectProcessedTransactions(
        rawTransactions,
        walletAddresses,
        "pending", // 👈 This is the key difference
      );

      return pendingTx.map((tx) => ({
        ...tx,
        valueToWallet: calculateTransactionValue(tx, walletAddresses),
      }));
    },
    enabled: !!blockchainClient && clientType === "public" && currentAddresses.length > 0,
    staleTime: 10000, // Shorter stale time for pending (10 seconds)
    refetchInterval: 30000, // More frequent updates (30 seconds)
  });
};

export const usePrivateClientPendingTransactions = () => {
  // Similar implementation for private clients
  // Uses getWalletTransactionHistory instead
};

export const usePendingTransactions = () => {
  const clientType = useSelector((state: WalletState) => state.client.type);
  const privateQuery = usePrivateClientPendingTransactions();
  const publicQuery = usePublicClientPendingTransactions();
  return clientType === "private" ? privateQuery : publicQuery;
};
```

**2. Update transaction keys** to include pending history:

```typescript
const transactionKeys = {
  all: ["transactions"] as const,
  tx: (txid: string) => [...transactionKeys.all, txid] as const,
  pending: () => [...transactionKeys.all, "pending"] as const,
  pendingHistory: () => [...transactionKeys.all, "pending", "history"] as const,
  confirmedHistory: () => [...transactionKeys.all, "confirmed", "history"] as const,
  // ... other keys
};
```

## Implementation Notes

- The blockchain client already returns all transactions (confirmed + pending)
- The `selectProcessedTransactions` function already handles filtering by status
- We just need to create the parallel "pending" path alongside the existing "confirmed" path
- React Query will handle all the caching, refetching, and invalidation automatically

---

##  Testing

After implementation, verify:
- [ ] 1:1 transactions (exact amount, no change) appear in pending
- [ ] 2:2 transactions (consolidation) appear in pending
- [ ] Transactions with change outputs still work correctly
- [ ] Self-sends to wallet addresses are tracked
- [ ] Both public and private clients work correctly
- [ ] Transactions disappear from pending once confirmed

RIYAKUMARI001:
What kind of change does this PR introduce?

Bugfix

Issue Number:

Fixes #407 

Snapshots:

Screenshot 2026-02-08 222413
Screenshot 2026-02-08 222358
Screenshot 2026-02-08 222342
Screenshot 2026-02-08 224604
Screenshot 2026-02-08 222328

Summary
This PR resolves a critical bug where transactions without change outputs (e.g., "MAX" sends, consolidations, or 1:1 transactions) would disappear from the Caravan "Pending" transactions view immediately after broadcast.

The Fix:

Previously, the pending transaction logic relied primarily on detecting new "unconfirmed" outputs in the wallet's change/deposit nodes. For transactions that spent the entire balance to a single external or internal address without a change output, this detection would fail.
This PR refactors the transaction hooks to use historical address scanning for pending transactions.
Updated 
selectProcessedTransactions
 to support a explicit "pending" filter.
Implemented usePublicClientPendingTransactions and usePrivateClientPendingTransactions that fetch the mempool history of all relevant wallet addresses.
This ensures that any transaction involving the wallet's addresses is tracked in the Pending tab, regardless of whether a change output exists.
Motivation: Users performing consolidation or spending their full balance were left with no visual feedback in Caravan that their transaction was successfully broadcast and pending. This fix provides a reliable and consistent UI state.

Does this PR introduce a breaking change?

No.

Checklist
 I have tested my changes thoroughly.
 I have added or updated tests to cover my changes (if applicable).
 I have verified that test coverage meets or exceeds 95% (if applicable).
 I have run the test suite locally, and although some legacy environment issues exist, my changes do not introduce new regressions.
 I have written tests for all new changes/features
 I have followed the project's coding style and conventions (formatted with Prettier).
 I have created a changeset to document my changes (npm run changeset)
Other information Successfully verified on Bitcoin Testnet using a 2-of-2 multisig wallet integrated with Sparrow. A "MAX" send (consolidation) was performed, and the transaction was correctly tracked in Caravan's "Pending" tab until it reached its first confirmation.

Yes 


Legend101Zz:
Screenshot 2026-03-04 at 20 10 53

@RIYAKUMARI001 Sorry for the delay for getting back on this PR , the initial testing looks really good I was able to on a private node send a 1 to 1 tx and I was able to see that in my pending tx tab :)

apps/coordinator/src/clients/transactions.ts
```diff
@@ -1,26 +1,32 @@
...
 export const transactionKeys = {
   all: ["transactions"] as const,
   tx: (txid: string) => [...transactionKeys.all, txid] as const,
   pending: () => [...transactionKeys.all, "pending"] as const,
+  pendingHistory: () => [...transactionKeys.all, "pending", "history"] as const,
   txWithHex: (txid: string) =>
     [...transactionKeys.all, txid, "withHex"] as const,
   coins: (txid: string) => [...transactionKeys.all, txid, "coins"] as const,
-  confirmedHistory: () => [...transactionKeys.all, "confirmed"] as const,
+  confirmedHistory: () =>
```
Legend101Zz: don' change this

apps/coordinator/src/clients/transactions.ts
```diff
@@ -1,26 +1,32 @@
...
 import { bitcoinsToSatoshis } from "@caravan/bitcoin";
 
+const MAX_TRANSACTIONS_TO_FETCH = 500;
+
 // Centralized query key factory for all transaction-related queries
 export const transactionKeys = {
   all: ["transactions"] as const,
   tx: (txid: string) => [...transactionKeys.all, txid] as const,
   pending: () => [...transactionKeys.all, "pending"] as const,
+  pendingHistory: () => [...transactionKeys.all, "pending", "history"] as const,
```
Legend101Zz: so you can remove this too

apps/coordinator/src/clients/transactions.ts
```diff
@@ -44,48 +50,6 @@ export const useFetchTransactionDetails = (txid: string) => {
   });
 };
 
-// Hook for fetching pending transaction IDs and their details
-// Service function for fetching pending transaction fees
-const fetchPendingTransactionFee = async (
```
Legend101Zz, line 49: yes these can be removed as you did

apps/coordinator/src/clients/transactions.ts
```diff
@@ -106,58 +70,126 @@ export const useTransactionsWithHex = (txids: string[]) => {
 };
 
 // Basic hook for raw pending transactions (no processing)
-export const useRawPendingTransactions = () => {
+export const usePublicClientPendingTransactions = () => {
```
Legend101Zz: no need to add new hook

apps/coordinator/src/clients/transactions.ts
```diff
@@ -106,58 +70,126 @@ export const useTransactionsWithHex = (txids: string[]) => {
...
   });
+};
 
-  return {
-    transactions,
-    isLoading,
-    error,
-    refetch: () => {
-      transactionQueries.forEach((query) => query.refetch());
+export const usePrivateClientPendingTransactions = () => {
```
Legend101Zz: same no need to add new hook

apps/coordinator/src/clients/transactions.ts
```diff
@@ -106,58 +70,126 @@ export const useTransactionsWithHex = (txids: string[]) => {
...
+    refetchInterval: 30000,
+  });
 };
 
-// Hook for processed pending transactions - uses selector
 export const usePendingTransactions = () => {
-  const walletAddresses = useSelector(getWalletAddresses);
-  const rawPendingQuery = useRawPendingTransactions();
+  const clientType = useSelector((state: WalletState) => state.client.type);
+  const privateQuery = usePrivateClientPendingTransactions();
```
Legend101Zz: see instead in txHistory.ts we already have these hooks so add a filter parameter to the existing hooks in txHistory.ts:

like eg :

```ts
export const usePublicClientTransactions = (
  filter: "confirmed" | "pending" | "all" = "confirmed"
) => {
```

and then here in transcation.ts file 

```ts
export const usePendingTransactions = () => {
  const clientType = useSelector((state: WalletState) => state.client.type);
  const privateQuery = usePrivateClientTransactions("pending");
  const publicQuery = usePublicClientTransactions("pending");
  const query = clientType === "private" ? privateQuery : publicQuery;

  return {
    transactions: query.data || [],
    isLoading: query.isLoading,
    error: query.error,
    refetch: query.refetch,
  };
};
```

and delete the rest unnecessary and duplicate code from transactions.ts , let's follow the DRY ( do not repeat ) principle here :)

apps/coordinator/src/selectors/wallet.ts
```diff
@@ -543,14 +543,15 @@ export const selectWalletConfig = createSelector(
 export const selectProcessedTransactions = (
   transactions: WalletTransactionDetails[],
   walletAddresses: string[],
-  filter: "all" | "confirmed" | "unconfirmed" = "all",
+  filter: "all" | "confirmed" | "unconfirmed" | "pending" = "all",
```
Legend101Zz: why add a new status , we can add use "unconfirmed" itself , also in above too

apps/coordinator/src/clients/transactions.ts
```diff
@@ -1,26 +1,32 @@
...
   Slice,
+  SliceWithLastUsed,
+  WalletState,
   selectProcessedTransactions,
 } from "selectors/wallet";
 import { calculateTransactionValue } from "utils/transactionCalculations";
 import { useGetClient } from "hooks/client";
 import { bitcoinsToSatoshis } from "@caravan/bitcoin";
 
+const MAX_TRANSACTIONS_TO_FETCH = 500;
```
Legend101Zz: can be removed now

Legend101Zz (changes requested):
Hi @RIYAKUMARI001, thanks for the PR! I had a few comments while going through the changes.

Instead of introducing new hooks in `transactions.ts` (which would duplicate the public/private client logic we already have in `txHistory.ts`), I think we could extend the existing hooks — `usePublicClientTransactions` and `usePrivateClientTransactions` — by adding a `filter` parameter. We could default this to `"confirmed"` to preserve the current behavior.

With that in place, `usePendingTransactions` could simply call these hooks with `"pending"`, which avoids creating a separate hook tree. Since `selectProcessedTransactions` already supports filtering, the integration should be fairly straightforward.

One thing that might be worth discussing with @bucko13 , though: if we reuse the same hooks for both confirmed history and pending transactions, they may end up having different polling requirements. Pending transactions likely need a faster `refetchInterval` than confirmed ones. So we might want to consider whether that should be configurable via options passed to the hooks, or handled in another way.

Not a blocker, but probably good to align on the approach before we move forward.

Legend101Zz:
Hi @RIYAKUMARI001 gentle ping to check if you had the chance to work on the reviews :)

RIYAKUMARI001:
> Hi @RIYAKUMARI001 gentle ping to check if you had the chance to work on the reviews :)

Hi! Yes, I saw the requested changes. I’m planning to start working on them now. My exams were going on, and tomorrow is my last exam. After that, I’ll push the changes as soon as possible.

RIYAKUMARI001:
Screenshot 2026-04-02 163923
Screenshot 2026-04-02 160859
Hi @Legend101Zz , thank you for your patience. I’ve gone through all your comments and added a changeset. Please take a look when you can. I'm ready for another review. Thanks.

apps/coordinator/src/clients/transactions.ts
RIYAKUMARI001: Restored it!

apps/coordinator/src/clients/transactions.ts
RIYAKUMARI001: Done. Removed the unnecessary pendingHistory

apps/coordinator/src/clients/transactions.ts
RIYAKUMARI001: Understood. Removed the redundant hook and switched to using the existing ones with a filter.

apps/coordinator/src/clients/transactions.ts
RIYAKUMARI001: Removed this as well. Now utilizing the consolidated base hooks

apps/coordinator/src/clients/transactions.ts
RIYAKUMARI001: Done! I've moved the logic to txHistory.ts and refactored it to use the existing client hooks with a filter parameter as suggested. The duplicate code has been removed

apps/coordinator/src/selectors/wallet.ts
RIYAKUMARI001: I've reverted to using the existing 'unconfirmed '

apps/coordinator/src/clients/transactions.ts
RIYAKUMARI001: done!

RIYAKUMARI001:
> Hi @RIYAKUMARI001, thanks for the PR! I had a few comments while going through the changes.
> 
> Instead of introducing new hooks in `transactions.ts` (which would duplicate the public/private client logic we already have in `txHistory.ts`), I think we could extend the existing hooks — `usePublicClientTransactions` and `usePrivateClientTransactions` — by adding a `filter` parameter. We could default this to `"confirmed"` to preserve the current behavior.
> 
> With that in place, `usePendingTransactions` could simply call these hooks with `"pending"`, which avoids creating a separate hook tree. Since `selectProcessedTransactions` already supports filtering, the integration should be fairly straightforward.
> 
> One thing that might be worth discussing with @bucko13 , though: if we reuse the same hooks for both confirmed history and pending transactions, they may end up having different polling requirements. Pending transactions likely need a faster `refetchInterval` than confirmed ones. So we might want to consider whether that should be configurable via options passed to the hooks, or handled in another way.
> 
> Not a blocker, but probably good to align on the approach before we move forward.

Thank you for the detailed feedback. I changed the hooks in txHistory.ts to use a filter parameter and followed the DRY principle as you suggested. I also made the refetchInterval faster, now set at 30 seconds, for unconfirmed transactions to keep the UI updated.

Legend101Zz (approved):
Thanks for the changes they look solid to me now :) 

Image

Legend101Zz:
@bucko13 a gentle reminder for giving this a final review when you get the chance :)

bucko13 (approved):

bucko13:
Very nice cleanup. Thank you!

jpcummins (approved):

RIYAKUMARI001:
Thanks again for the reviews and approvals. It looks like all checks have passed now. When someone has a moment, could you please review the merge? Thank you!

PR #477, opened 2026-02-14
fix(cpfp): correct fee satisfaction loop and change recalculation logic

mansiverma897993:

**Issue Number:**

Fixes #

**Snapshots/Videos:**

**Summary**

## Checklist

**Other information**

### DIFF
--- a/packages/transactions/src/cpfp.ts
+++ b/packages/transactions/src/cpfp.ts
@@ -196,23 +196,28 @@ export const createCPFPTransaction = (options: CPFPOptions): string => {
    * is attractive for miners to include in a block.
    */
 
-  for (const utxo of availableInputs) {
-    if (
-      isCPFPFeeSatisfied(txAnalyzer, childTxTemplate) &&
-      childTxTemplate.needsChange
-    ) {
-      break; // Stop adding inputs once CPFP fee requirements and change requirements (to ensure we don't end up with dust Output) are met
-    }
-    // Skip if this UTXO is already added
-    if (
-      childTxTemplate.inputs.some(
-        (input) => input.txid === utxo.txid && input.vout === utxo.vout,
-      )
-    ) {
-      continue;
-    }
-    childTxTemplate.addInput(BtcTxInputTemplate.fromUTXO(utxo));
+ for (const utxo of availableInputs) {
+
+  // Skip if already added
+  if (
+    childTxTemplate.inputs.some(
+      (input) => input.txid === utxo.txid && input.vout === utxo.vout
+    )
+  ) {
+    continue;
   }
+
+  // Add input
+  childTxTemplate.addInput(BtcTxInputTemplate.fromUTXO(utxo));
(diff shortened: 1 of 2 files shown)


Legend101Zz:
@mansiverma897993 can you please describe what exactly does this PR solve ? maybe add the issue link it resolves or any small description on the issue would help :)

mansiverma897993:
@Legend101Zz Hi, This PR fixes an issue in the CPFP fee satisfaction loop where the fee recalculation logic could result in incorrect fee estimation during transaction confirmation.
Previously, the loop could stop prematurely or calculate an insufficient fee when adjusting the CPFP transaction. This change improves the recalculation logic so that the fee is reevaluated correctly until the required fee rate is satisfied.
Additionally, this PR adds support for Coldcard interaction in the confirm-address flow to ensure better compatibility with hardware wallets.
Overall, the goal of this PR is to ensure more reliable fee calculation and smoother confirmation flow when creating CPFP transactions.

packages/transactions/src/cpfp.ts
```diff
@@ -196,24 +196,29 @@ export const createCPFPTransaction = (options: CPFPOptions): string => {
    * is attractive for miners to include in a block.
    */
 
-  for (const utxo of availableInputs) {
-    if (
-      isCPFPFeeSatisfied(txAnalyzer, childTxTemplate) &&
-      childTxTemplate.needsChange
```
Legend101Zz, line 202: The original loop works correctly. The reason is that the change output starts at `amountSats: 0`, which means `currentFee` during the loop equals the full sum of all inputs — so `isCPFPFeeSatisfied` is trivially true from the very first iteration. The actual controlling condition was always `needsChange`, which correctly ensures we don't terminate with a dust or zero change output. The fee check in the original loop was a no-op in practice, not a bug.

This PR moves `adjustChangeOutput()` inside the loop (so `currentFee` now reflects real fees) and makes the fee check meaningful — that part is a fair cleanup conceptually. But dropping `needsChange` from the break condition means we can now exit the loop with a valid fee rate but a dust or zero change output. That scenario already has a hard throw downstream in strict mode, so this just trades a clean validation error for silent bad behavior.

Also, calling `adjustChangeOutput()` on every iteration is unnecessary work — it only needs to run once after the loop finishes, which is exactly what the original does.

Could you share what specific issue or failure this was triggered by? I don't see a linked issue and this is touching core CPFP logic in `@caravan/transactions`, so before merging we'd need a thorough regression pass across all CPFP scenarios — at minimum the happy path, the dust-change edge case, and the insufficient-inputs case. Without a concrete reproduction of the original bug, the risk/reward here doesn't look favorable.

Legend101Zz (changes requested):
@mansiverma897993 Thanks for the PR, but I'm not convinced this change is needed, and I think it introduces a regression, I have a detailed comment below on this below , please let me know what you think :)

PR #488, opened 2026-02-24
Fix ColdCard Q PSBT signature validation error

### LINKED ISSUE #377: Error uploading signature to caravan
Sharing a report from another user but should be investigated ASAP. 

When uploading his signed PSBT, using a micro SD, from his ColdCard Q to Caravan, he received the message. Signature for input 4 is invalid. CCQ firmware was up to date. His trezor was able to sign in caravan. 

He tried again with Sparrow, and it was successfully uploaded. 

EDIT (10/3):
I did not clearly describe the signed PSBT bit here. This was for signing with a coldcard which has an "upload psbt" step, but it wasn't with the import psbt for tx creation that would happen on the first step.

harrshita123:

Bugfix

**Issue Number:**

Fixes #377

**Snapshots/Videos:**

(Verified with E2E suite in Docker; 11/11 tests passing)

N/A

**Summary**

Fixed the signature validation error reported when uploading ColdCard Q PSBTs. 

The issue stems from [validateMultisigPsbtSignature](cci:1://file:///Users/apple/caravan/packages/caravan-psbt/src/psbtv0/psbt.ts:209:0-264:2) relying strictly on `bip32Derivation` metadata. ColdCard Q sometimes strips these derivations from the signed PSBT, causing the validator to return an error even if the signature itself is valid. 

Added a fallback to check the public keys directly from the `redeemScript`/`witnessScript` when derivation info is missing. I also updated [SignatureImporter](cci:2://file:///Users/apple/caravan/apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx:56:0-633:1) to provide more specific error messages to help distinguish between metadata issues and actual cryptographic failures.

No

## Checklist

**Other information**

Added a regression test in [psbt.test.ts](cci:7://file:///Users/apple/caravan/packages/caravan-bitcoin/src/psbt.test.ts:0:0-0:0) that verifies validation passes even after stripping derivation data from a PSBT.

Yes

image

### DIFF
--- a/apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
+++ b/apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
@@ -434,7 +434,7 @@ class SignatureImporter extends React.Component {
             inputs[inputIndex].amountSats,
           );
         } catch (e) {
-          errback(`Signature for input ${inputNumber} is invalid.`);
+          errback(`Signature for input ${inputNumber} is invalid: ${e.message}`);
           return;
         }
         if (publicKey) {
@@ -448,7 +448,7 @@ class SignatureImporter extends React.Component {
 
             if (
               finalizedSignatureImporter.signature[inputIndex] ===
-                inputSignature ||
+              inputSignature ||
               finalizedSignatureImporter.publicKeys[inputIndex] === publicKey
             ) {
               errback(
@@ -459,7 +459,9 @@ class SignatureImporter extends React.Component {
           }
           publicKeys.push(publicKey);
         } else {
-          errback(`Signature for input ${inputNumber} is invalid.`);
+          errback(
+            `Signature for input ${inputNumber} is invalid (no matching public key found).`,
+          );
           return;
         }
       }
@@ -541,7 +543,7 @@ class SignatureImporter extends React.Component {
               );
             } catch (e) {
               // eslint-disable-next-line no-console
-              console.error(e);
+              console.error(`Validation error for input ${inputIndex}:`, e);
             }
             // FIXME `&& publicKey is member of a particular root xpub/fingerprint` should be added here!
             if (publicKey) {
@@ -552,7 +554,9 @@ class SignatureImporter extends React.Component {
             }
           }
           if (!publicKey) {
-            errback(`No valid signature for input ${inputIndex} found.`);
+            errback(
+              `No valid signature for input ${inputIndex} found in the uploaded PSBT.`,
+            );
             return;
           }
         }
--- a/apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
+++ b/apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
@@ -434,7 +434,9 @@ class SignatureImporter extends React.Component {
             inputs[inputIndex].amountSats,
           );
         } catch (e) {
-          errback(`Signature for input ${inputNumber} is invalid: ${e.message}`);
+          errback(
+            `Signature for input ${inputNumber} is invalid: ${e.message}`,
+          );
           return;
         }
         if (publicKey) {
@@ -448,7 +450,7 @@ class SignatureImporter extends React.Component {
 
             if (
               finalizedSignatureImporter.signature[inputIndex] ===
-              inputSignature ||
+                inputSignature ||
               finalizedSignatureImporter.publicKeys[inputIndex] === publicKey
             ) {
               errback(
--- a/packages/caravan-psbt/src/psbtv0/psbt.test.ts
+++ b/packages/caravan-psbt/src/psbtv0/psbt.test.ts
@@ -14,6 +14,7 @@ import _ from "lodash";
 import { psbtArgsFromFixture } from "./utils";
 import assert from "assert";
 import { combineBip32Paths } from "@caravan/bip32";
+import { Psbt } from "bitcoinjs-lib-v6";
 
 describe("getUnsignedMultisigPsbtV0", () => {
   TEST_FIXTURES.transactions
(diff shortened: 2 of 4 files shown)


packages/caravan-psbt/src/psbtv0/psbt.ts
```diff
@@ -226,18 +226,41 @@ export const validateMultisigPsbtSignature = (
...
+        network: (psbt as any).opts.network,
+      });
+      if (multisig && multisig.pubkeys) {
+        for (const pubkey of multisig.pubkeys) {
+          if (isValidSignature(pubkey, msgHash, signatureBuffer)) {
+            return pubkey.toString("hex");
+          }
+        }
+      }
+    } catch (e) {
```
bucko13: let's not silently catch and swallow errors

bucko13 (changes requested):
thanks for adding this! interesting that the ColdCard Q "sometimes" behaves differently. 

We'll need to add a [changest](https://github.com/caravan-bitcoin/caravan?tab=readme-ov-file#changesets) for the package to bump the version as well as reverting the swallowing of the thrown error to make sure that gets surfaced. otherwise this should be good to go.

harrshita123:
Thanks for the feedback! I've removed the try/catch block to let the errors surface properly and added a changeset for the package bump. The CI should now be green.

bucko13 (dismissed):

harrshita123:
Fixed the CI issues. I've updated the PR ,please take another look when you get a chance.

harrshita123:
I fixed the CI issue . The failure was due to Prettier formatting in SignatureImporter.jsx after adding the new validation error handling. I formatted the file, verified the changed files with Prettier, checked coordinator lint, and confirmed the changeset status.

packages/caravan-psbt/src/psbtv0/psbt.test.ts
```diff
@@ -94,6 +95,22 @@ describe("validateMultisigSignaturePsbt", () => {
             expect(pubkey).toEqual(false);
           });
         });
+
+        it("validates signature even if bip32Derivation is missing from the PSBT", () => {
```
Legend101Zz, line 99: Small test suggestion: it would also be good to add a negative-path case where both `bip32Derivation` *and* `witnessScript` / `redeemScript` are stripped from the PSBT, and verify that validation correctly returns `false`.

That helps document the actual fallback contract here:

- `bip32Derivation` missing , still recoverable from scripts
- scripts missing too , no longer enough information to recover cosigner pubkeys

packages/caravan-psbt/src/psbtv0/psbt.ts
```diff
@@ -238,6 +238,22 @@ export const validateMultisigPsbtSignature = (
       return pubkey.toString("hex");
     }
   }
+
+  const script = input.witnessScript || input.redeemScript;
+  if (script) {
+    const multisig = payments.p2ms({
+      output: script,
+      network: (psbt as any).opts.network,
```
Legend101Zz, line 246: I think `(psbt as any).opts.network` in `psbt.ts` is probably an unnecessary type-escape hatch here.

`payments.p2ms` only uses `network` for address derivation/encoding paths — it doesn't need it to recover or populate the multisig pubkeys from the script itself.

So the simplest fix should just be:

```ts
const multisig = payments.p2ms({ output: script });

File: apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx [hunk 4]
```diff
@@ -541,7 +545,7 @@ class SignatureImporter extends React.Component {
               );
             } catch (e) {
               // eslint-disable-next-line no-console
-              console.error(e);
+              console.error(`Validation error for input ${inputIndex}:`, e);
             }
             // FIXME `&& publicKey is member of a particular root xpub/fingerprint` should be added here!
```
Legend101Zz, line 550: I think this concern applies a bit more strongly now with the fallback path.

Once `bip32Derivation` is missing, we no longer have derivation paths to cross-check recovered pubkeys against the configured xpubs — the validator is effectively trusting whatever keys are present in the `witnessScript` / `redeemScript`.

That’s probably acceptable for this PR since the goal is robustness/compatibility, but it does slightly weaken identity verification semantics.

Might be worth opening a follow-up issue to track this explicitly: in the fallback flow, a valid signature from *any* key present in the redeem/witness script would currently pass validation, even if that key is not one Caravan was originally expecting/configured for , @bucko13 what do you think ?

Legend101Zz (changes requested):
Hi @harrshita123 — thanks for the PR, and thanks @bucko13 for the detailed back-and-forth as well. I learned quite a bit reading through the discussion and tracing the flow end-to-end.

One small factual correction though: I went through the Coldcard side to verify the “Coldcard QR strips `bip32Derivation`” framing, and I don’t think that’s exactly what’s happening.

In the `psbtInputProxy.serialize()` path in the Coldcard firmware (`shared/psbt.py`)   https://github.com/Coldcard/firmware/blob/master/shared/psbt.py#L989 , Coldcard explicitly writes derivations back during serialization:

`for k in self.subpaths: wr(PSBT_IN_BIP32_DERIVATION, self.subpaths[k], k)`

`subpaths` itself is populated from the incoming PSBT during parsing, so Coldcard preserves derivation paths if they were present originally — it just doesn’t reconstruct metadata that was already missing upstream.

So I’d probably reframe the issue slightly:

The real bug is that Caravan’s validator assumes `bip32Derivation` exists on every signed PSBT input, and that assumption is fragile. BIP-174 does not require signers/intermediate tools to preserve those fields, and in practice there are multiple places where they can disappear during round-trips — QR/NFC transport (BBQr), SD-card workflows, or wallets like Sparrow/Nunchuk re-emitting the PSBT.

The fix in this PR is still absolutely the correct approach regardless of where the field disappeared in the pipeline: recovering cosigner pubkeys from `witnessScript` / `redeemScript`, which *are* required to remain according to BIP-174.

So IMO this is less interesting as a “Coldcard bug” and more interesting as a “Caravan correctness/robustness fix that Coldcard workflows helped expose.” Might be worth tightening the changelog wording a bit to reflect that framing :)

apps/coordinator/src/components/ScriptExplorer/SignatureImporter.jsx
bucko13: yeah, i think that makes sense to do a follow up. in general caravan coordinator doesn't do a good job of handling a config where the path isn't included and the UI will just kind of choke on that.

harrshita123:
@bucko13 reopen this pr , then i resolved copilot comments

PR #495, opened 2026-03-06
Verification  coldcard checkout

### LINKED ISSUE #476: Missing Address Verification for Coldcard?
I was surprised to see this when testing out the BCUR2 additions, but it looks like we don't have address verification for coldcards. probably want to add this. 

It would require:
- Enabling it in this dropdown
- Including a step/button to register if not done already
- added to the test suite

Until we have proper message signing interactions, this can be done by just giving instructions for the address explorer (after registration) 

Image

KartikBansal4976:
Fix Coldcard address verification flow and add config helper UI

Fixes #476

**ISSUE**
Coldcard address verification was not fully implemented in the confirmation flow.
In some cases the device could not be selected due to overly restrictive legacy path checks, and users were not guided through installing the multisig configuration required for manual verification.

This PR completes the Coldcard verification flow

**CHANGES**

1. Enable Coldcard in the Confirm on Device method selection
2. Relax legacy path validation that prevented Coldcard from appearing
3. Add Download Coldcard Config helper button during the confirmation flow
4. Improve confirmation messaging to include multisig configuration installation step

**TEST**

1. Added unit tests covering Coldcard interaction logic
2. Updated e2e tests to verify the config download helper
3. Fixed a BCUR2 interaction test issue blocking the test suite

Local Verification 
npx turbo test --filter=@caravan/wallets
npx vitest run (apps/coordinator)
npx vitest run --coverage

### DIFF
--- a/.changeset/add-coldcard-address-verification.md
+++ b/.changeset/add-coldcard-address-verification.md
@@ -1,0 +1,18 @@
+---
+"caravan-coordinator": minor
+"@caravan/wallets": minor
+---
+
+Add address verification support for Coldcard hardware wallets
+
+Implements ColdcardConfirmMultisigAddress interaction for manual address verification via address explorer. Users can now verify multisig addresses on Coldcard devices by:
+1. Selecting Coldcard from the address verification dropdown (now enabled)
+2. Receiving clear instructions for manual verification using popular block explorers
+3. Manually confirming the address matches their device
+4. Completing the verification workflow
+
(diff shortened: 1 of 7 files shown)


Legend101Zz:
Hi @KartikBansal4976 thanks for the PR , the e2e changes you did , actually we had a major revision of our e2e structure so can you please pull in the latest changes and change it according to that :)

PR #496, opened 2026-03-24
Feat/sp sending

### LINKED ISSUE #467: Silent payment support
While working on [bitcoinjs silent payment support](https://github.com/bitcoinjs/bitcoinjs-lib/pull/2269 ) [bip352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki)  I stumbled upon this repo as it already implements psbtv2.
Might be interesting to also implement it here for devs that already depend on this stack.

Some example are already out there:
- https://github.com/BlueWallet/SilentPayments/blob/master/src/index.ts
- Blind bit with PSBT: https://github.com/setavenger/blindbit-wallet-cli/blob/dc555fb7ce2d6cb79f38a8444bfd518e91e35c70/pkg/wallet/send.go#L316
- Sparrow: https://github.com/sparrowwallet/drongo/tree/master/src/main/java/com/sparrowwallet/drongo/silentpayments
- https://github.com/cygnet3/rust-silentpayments
-Bip375 example implementation : https://github.com/macgyver13/bips/tree/bip375-reference-testvectors/bip-0375

I was working on something here https://github.com/bitcoinjs/bitcoinjs-lib/pull/2269 but it needs bip370, 375 and a merge in https://github.com/rust-bitcoin/rust-secp256k1/pull/876 and https://github.com/bitcoinjs/tiny-secp256k1 to become useful

My suggestion would be to split it up in:
- Creating SP addresses
- Sending to SP addresses
- Receiving on SP address (due to different choices in scanning models eg tweak data, taproot data, full scan etc)
- Parallel you could implement [bip353](https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki) using https://github.com/TheBlueMatt/dnssec-prover/tree/master/wasmpack

jvgelder:

feature

**Issue Number:**

Fixes #467 

**Summary**

Implements sending to silent payment [Bip352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki)  address via psbtv2 [bip375](https://github.com/bitcoin/bips/blob/master/bip-0375.mediawiki). 

Still work in progress but gives some insights in how I plan to integrate.

TODO:
- [ ] Improve test coverage

Should not introduce any breaking changes

## Checklist

**Other information**

Yes


jvgelder:
@Legend101Zz could you have a look ?

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
```diff
@@ -224,23 +224,26 @@ export class PsbtConversionMaps extends PsbtV2Maps {
...
+      const script = this.outputMaps[i].get(KeyType.PSBT_OUT_SCRIPT);
+
       if (
-        !this.outputMaps[i].has(KeyType.PSBT_OUT_SCRIPT) ||
+        (!spInfo && !script) ||
         !this.outputMaps[i].has(KeyType.PSBT_OUT_AMOUNT)
       ) {
         console.warn(
-          `Output ${i} is missing previous out script or amount. Skipping.`,
+          `Output ${i} is missing SP_INFO, script or amount. Skipping.`,
```
jvgelder: should have added SP_INFO instead of replaced previous out

Legend101Zz:
> @Legend101Zz could you have a look ?

Hi @jvgelder — really sorry for the delay. I hadn’t had a chance to review this closely yet, but I’ll start going through it today or tomorrow. Thanks a lot for the PR!

jvgelder:
> > @Legend101Zz could you have a look ?
> 
> Hi @jvgelder — really sorry for the delay. I hadn’t had a chance to review this closely yet, but I’ll start going through it today or tomorrow. Thanks a lot for the PR!

Thanks note this is the first part after this we should add the DLEQ proofs which I think should be part of this PR.

After that we need to add the actual usage of this code to the wallet? to allow users to specify the silent payment address they want to send to (and thus address decoding too)

Legend101Zz:
> > > @Legend101Zz could you have a look ?
> > 
> > 
> > Hi @jvgelder — really sorry for the delay. I hadn’t had a chance to review this closely yet, but I’ll start going through it today or tomorrow. Thanks a lot for the PR!
> 
> Thanks note this is the first part after this we should add the DLEQ proofs which I think should be part of this PR.
> 
> After that we need to add the actual usage of this code to the wallet? to allow users to specify the silent payment address they want to send to (and thus address decoding too)

Hi @jvgelder thanks for the heads up ... let me get back to you once I go through the PR , finally got the chance to review it ... thanks for your patience :)

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,8 +841,541 @@ export class PsbtV2 extends PsbtV2Maps {
         throw Error("PsbtV2 input hight locktime is gte 500000000.");
       }
     }
+    this.validateSilentPayments();
```
Legend101Zz, line 854: I think this needs one more BIP375 guard before signatures are accepted on a
silent-payment PSBT. `validateSilentPayments()` rejects explicit
`PSBT_IN_SIGHASH_TYPE` values that are not `SIGHASH_ALL`, but `addPartialSig()`
can still add a partial signature whose trailing sighash byte is `SIGHASH_NONE`
or `SIGHASH_SINGLE` when no `PSBT_IN_SIGHASH_TYPE` field is present.

Like , BIP375 says that if silent payment outputs are present, a signer must fail when
the sighash type is not `SIGHASH_ALL` , if no sighash type is provided, it must
sign using `SIGHASH_ALL`. The current code only uses `handleSighashType()` to
update modifiable flags after inserting the signature, so a non-ALL signature
can be serialized and reparsed successfully.

So should we reject this in `addPartialSig()` or in `handleSighashType()` when
`this.hasSilentPaymentOutputs` is true? A regression test should cover a silent
payment PSBT where `PSBT_OUT_SCRIPT` is already computed and `addPartialSig()`
is called with a signature ending in `0x02` or `0x03`.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,8 +841,541 @@ export class PsbtV2 extends PsbtV2Maps {
...
+      }
+
+      const witnessUtxo = this.inputMaps[i].get(KeyType.PSBT_IN_WITNESS_UTXO);
+      let pubkeyBytes: Buffer | null = null;
+
+      if (witnessUtxo) {
+        pubkeyBytes = getTaprootOutputKeyFromWitnessUtxo(witnessUtxo);
+      }
+
+      if (!pubkeyBytes) {
```
Legend101Zz: So for P2WPKH, P2SH-P2WPKH, and P2PKH inputs, this uses the first
`PSBT_IN_BIP32_DERIVATION` key as the input public key for DLEQ verification
and `input_hash` right, but I do not see a check that this pubkey actually matches
the prevout script.

That leaves a dangerous case: like a PSBT can spend a P2WPKH prevout whose script
hash commits to pubkey A, but include a BIP32 derivation key for pubkey B. The
DLEQ proof can be valid for B, and the SP output script can be computed from B,
while the final transaction is signed with A. The transaction can still be
consensus-valid, but the receiver scans using the actual input pubkey A and
will not find the output.


This may not affect Caravan's current classic multisig coordinator path directly,
because P2WSH/P2SH-P2WSH multisig inputs are not BIP352 shared-secret inputs.
But since this PR is adding generic BIP375 support in `@caravan/psbt`, I think
the P2WPKH/P2SH-P2WPKH/P2PKH public-key path should still verify that the
`PSBT_IN_BIP32_DERIVATION` pubkey matches the prevout script before using it for
DLEQ verification and `input_hash`.

Otherwise, a PSBT can provide a derivation pubkey B, generate a valid DLEQ proof
and SP output for B, while the actual input spends a prevout committed to pubkey
A. The final transaction can still be valid, but receivers scan using the actual
input pubkey revealed on-chain, so they would derive with A and miss the output.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,8 +841,541 @@ export class PsbtV2 extends PsbtV2Maps {
...
+          throw Error(
+            `PsbtV2 input ${i} uses non-SIGHASH_ALL (${sighashType}) with silent payments.`,
+          );
+        }
+      }
+    }
+
+    // Only non-labeled SP outputs can be recomputed from public PSBT data.
+    // Labeled outputs require recipient-side label state / scan private key.
+    const computedSPScanKeys = new Set<string>();
```
Legend101Zz: I think labeled outputs need to be either fully verified or explicitly rejected
for now. Right now validation skips outputs that have `PSBT_OUT_SP_V0_LABEL`
when deciding which scan keys need ECDH coverage and output-script validation.
That means a labeled SP output with an arbitrary `PSBT_OUT_SCRIPT` can pass
validation once `PSBT_GLOBAL_TX_MODIFIABLE` is zeroed, even with no ECDH share
or DLEQ proof for that scan key.

This is especially risky because BIP375 uses labels for SP change verification.
If this implementation cannot derive or verify labeled outputs yet, I think the
safe behavior is to fail validation when a labeled SP output has a script, or to
keep it non-final/in-progress until label-aware verification exists. Otherwise
the PSBT layer can bless an output script that is not tied to the stored silent
payment data, what do you think ?

Legend101Zz (changes requested):
@jvgelder Thanks for the patience here, and thanks for pushing this PR forward. I finally got time to do an initial deeper review.

Overall, this looks like a solid PSBT-layer foundation for BIP375/Silent Payments. The current version is much stronger than the first pass I looked at :) 

The main things I think still need attention before this  is safe to merge are:

- enforce `SIGHASH_ALL` when adding partial signatures to a PSBT with Silent Payment outputs
- verify that BIP32 derivation pubkeys used for P2WPKH/P2SH-P2WPKH/P2PKH Silent Payment derivation actually match the prevout script
- either fully support labeled Silent Payment outputs or reject/keep them non-final until label-aware verification exists


After this PSBT layer is correct, I think the next PRs should tackle the usable Caravan path: Silent Payment address decoding, coordinator PSBTv2 construction, clear input eligibility checks, signer/wallet handling for ECDH shares and DLEQ proofs, and a precise Taproot-support story, do you plan to have all those in this PR itself ?

jvgelder:
> @jvgelder Thanks for the patience here, and thanks for pushing this PR forward. I finally got time to do an initial deeper review.
> 
> Overall, this looks like a solid PSBT-layer foundation for BIP375/Silent Payments. The current version is much stronger than the first pass I looked at :)
> 
Thanks a lot for taking the time to review it, this gives me some more energy to move it forward.

> The main things I think still need attention before this is safe to merge are:
> 
>     * enforce `SIGHASH_ALL` when adding partial signatures to a PSBT with Silent Payment outputs

Good one, wasn't sure if I should alter those functions but a simple assert (if silent output && sighash != All ) should suffice I guess.

>     * verify that BIP32 derivation pubkeys used for P2WPKH/P2SH-P2WPKH/P2PKH Silent Payment derivation actually match the prevout script
> 
Il have a look.

>     * either fully support labeled Silent Payment outputs or reject/keep them non-final until label-aware verification exists
>
While we are at it we should support it in full :)
> 
> After this PSBT layer is correct, I think the next PRs should tackle the usable Caravan path: Silent Payment address decoding, coordinator PSBTv2 construction, clear input eligibility checks, signer/wallet handling for ECDH shares and DLEQ proofs, and a precise Taproot-support story, do you plan to have all those in this PR itself ?

I Might but just to manage the expectations a bit I am also working on various other projects, if we can get more people excited, I would direct more of my energy (and tokens) here ^^.
Would need some pointers on where to implement address en/decoding and how to change the UI so that it can handle silent payment output addresses.

packages/caravan-psbt/src/psbtv2/dleq.ts
```diff
@@ -0,0 +1,243 @@
...
+    Buffer.concat([
+      ABytes,
+      BBytes,
+      CBytes,
+      Buffer.from(G.toRawBytes(true)),
+      R1Bytes,
+      R2Bytes,
+      m,
+    ]),
+  );
```
macgyver13: tidy: function for dleqChallenge could be reused in verify too

packages/caravan-psbt/src/psbtv2/dleq.test.ts
```diff
@@ -0,0 +1,209 @@
+import { describe, expect, it } from "vitest";
+import { secp256k1 } from "@noble/curves/secp256k1";
+
+import {
+  generateDLEQProof,
+  multiplyCompressedPoint,
+  verifyDLEQProof,
+} from "./dleq";
+import { verifySilentPaymentDLEQProof } from "./silentpayment";
```
macgyver13: nit: import not used

packages/caravan-psbt/src/psbtv2/silentpayment.ts
```diff
@@ -0,0 +1,552 @@
...
+ * @returns 32-byte hash buffer.
+ */
+export function bip352TaggedHash(tag: string, data: Buffer): Buffer {
+  const tagHash = createHash("sha256").update(tag).digest();
+  return createHash("sha256")
+    .update(tagHash)
+    .update(tagHash)
+    .update(data)
+    .digest();
+}
```
macgyver13: duplicate taggedHash implementation in dleq.ts

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: see new vectors : https://github.com/caravan-bitcoin/caravan/pull/496/changes#diff-bc1ee7a599cdf1acc66b16e760460b1be919fd420a1484423afc522a5510d3ffR679-R713
thanks to [macgyver13](https://github.com/macgyver13)

if you have more vectors you can think of :)  https://github.com/macgyver13/bip375-test-generator/commits/bip32-more-cases/

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: Just talked with @macgyver13 and the PSBT_OUT_SP_V0_LABEL is just there as a hint for the hardware wallet: hey the output address i sent you is actually one of our labeled addresses when you verify you need to derive it properly to arrive at the same address.

To clarify the scan will most likely always has to live in the coordinator otherwise it will be close to impossible to scan for your incoming payments. Hence the coordinator can actually also derive labeled address from the scan key.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -424,17 +475,12 @@ export class PsbtV2 extends PsbtV2Maps {
...
+    return this.outputMaps.map((map) => {
       const value = map.get(KeyType.PSBT_OUT_SCRIPT);
-      if (!value) {
-        // This should never happen, but it can't be gracefully handled.
-        throw Error("PSBT_OUT_SCRIPT not set for an output");
-      }
-      indices.push(value.toString("hex"));
-    }
-    return indices;
+      if (!value) return "";
```
jvgelder: return "" and checking on empty string seems like a bad idea

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: Added a check https://github.com/caravan-bitcoin/caravan/pull/496/changes#diff-cefa8ce7b34d9637fc1bfd4f54edc413efd6e74068ebf1e97e4ab8cad063ee2fR1683-R1699

packages/caravan-psbt/src/psbtv2/psbtv2.ts
macgyver13: would it be possible to maintain the throw when PSBT_OUT_SP_V0_INFO is not present in the output map.

macgyver13 (commented):
Thanks for getting the ball rolling on the silent payments PSBT integration!

packages/caravan-psbt/src/psbtv2/silentpayment.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/dleq.test.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/dleq.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: yes changed it back

packages/caravan-psbt/src/psbtv2/dleq.test.ts
```diff
@@ -0,0 +1,208 @@
+import { describe, expect, it } from "vitest";
+import { secp256k1 } from "@noble/curves/secp256k1";
+
+import {
+  generateDLEQProof,
+  multiplyCompressedPoint,
+  verifyDLEQProof,
+} from "./dleq";
+function hex(s: string): Buffer {
```
macgyver13, line 10: nit: new line should be added here

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
```diff
@@ -224,23 +224,30 @@ export class PsbtConversionMaps extends PsbtV2Maps {
...
-        numberAmount,
-      );
+
+      // BIP375 unique identification uses v0 || Bscan || Bspend when
+      // PSBT_OUT_SP_V0_INFO is present, so the txid is stable before and after
+      // the real output script has been computed.
+      const outputScript = spInfo
+        ? Buffer.concat([Buffer.from([0x00]), spInfo])
+        : (script as Buffer);
+      tx.addOutput(outputScript, numberAmount);
```
macgyver13, line 250: This change to handle BIP375 unique id is not compatible with convertToV0
```
  public convertToV0 = () => {
    this.globalMap.set(
      V0KeyTypes.PSBT_GLOBAL_UNSIGNED_TX,
      this.buildUnsignedTx(),
    );
```

Have you considered creating a similar buildIdentifierTx() and using that in getTransactionId()?

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,6 +848,434 @@ export class PsbtV2 extends PsbtV2Maps {
...
+    return this.inputMaps.flatMap((map) => {
+      const txid = map.get(KeyType.PSBT_IN_PREVIOUS_TXID);
+      const vout = map.get(KeyType.PSBT_IN_OUTPUT_INDEX);
+      return txid && vout ? [Buffer.concat([txid, vout])] : [];
+    });
+  }
+
+  private getSilentPaymentOutputGroups(
+    onlyWithScripts = false,
+    skipLabeled = false,
```
macgyver13: why is skipLabeled necessary? 

L1227 is the only location where skipLabeled = true, all tests pass with false, maybe this was a condition for an older issue.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,6 +848,434 @@ export class PsbtV2 extends PsbtV2Maps {
...
+
+    for (const i of contributing) {
+      const pubkeyBytes = getSilentPaymentPubkeyFromInputDescriptor(
+        descriptors[i],
+        this.getSilentPaymentCandidatePubkeys(i),
+      );
+
+      if (!pubkeyBytes) {
+        continue;
+      }
```
macgyver13, line 975: returning null here is safer than continuing to avoid masking an invalid aggregate key

packages/caravan-psbt/src/psbtv2/silentpayment.ts
```diff
@@ -0,0 +1,615 @@
...
+  return bscan;
+}
+
+export function scanKeyHexFromBIP375Key(
+  key: string,
+  keyType: string,
+  context: string,
+): string {
+  return parseSilentPaymentScanKeyData(key, keyType, context).toString("hex");
+}
```
macgyver13: dead code?

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -1513,4 +2212,192 @@ export class PsbtV2 extends PsbtV2Maps {
...
+   * `validate()` already guarantees they have a script.
+   *
+   * Per BIP375: the Signer must not add a signature until this is true.
+   */
+  get hasAllSPOutputScripts(): boolean {
+    return this.outputMaps.every((map) => {
+      if (!map.has(KeyType.PSBT_OUT_SP_V0_INFO)) return true;
+
+      const script = map.get(KeyType.PSBT_OUT_SCRIPT);
+      return !!script && script.length > 0;
```
macgyver13: reuse hasNonEmtpyOutputScript?
```
  return this.outputMaps.every(
    (map, i) =>
      !map.has(KeyType.PSBT_OUT_SP_V0_INFO) || this.hasNonEmptyOutputScript(i),
  );
```

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
```diff
@@ -277,6 +284,8 @@ export class PsbtConversionMaps extends PsbtV2Maps {
     this.v0delete(this.globalMap, KeyType.PSBT_GLOBAL_INPUT_COUNT);
     this.v0delete(this.globalMap, KeyType.PSBT_GLOBAL_OUTPUT_COUNT);
     this.v0delete(this.globalMap, KeyType.PSBT_GLOBAL_TX_MODIFIABLE);
+    this.v0delete(this.globalMap, KeyType.PSBT_GLOBAL_SP_ECDH_SHARE);
```
macgyver13, line 299: v0delete() performs exact key match so this will not delete keys with key_data.  same issue for all `NonUniqueKeyTypeValue[]`

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,6 +848,434 @@ export class PsbtV2 extends PsbtV2Maps {
...
+            "PSBT_OUT_SP_V0_INFO. An output must have one or both per BIP375.",
+        );
+      }
+
+      if (spInfo && script && script.length === 0) {
+        throw Error(`PsbtV2 output ${i} has empty PSBT_OUT_SCRIPT.`);
+      }
+
+      if (spInfo && script && script.length > 0) {
+        if (this.PSBT_GLOBAL_TX_MODIFIABLE.length > 0) {
```
macgyver13: This could collapse to:
`if (this.hasNonEmptyOutputScript(i) && this.PSBT_GLOBAL_TX_MODIFIABLE.length > 0)`

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -1513,4 +2212,192 @@ export class PsbtV2 extends PsbtV2Maps {
...
+      (map) =>
+        map.has(KeyType.PSBT_OUT_SP_V0_INFO) &&
+        map.has(KeyType.PSBT_OUT_SP_V0_LABEL),
+    );
+
+    if (hasLabeledSilentPaymentOutput) {
+      throw new Error(
+        "Cannot compute labeled silent payment output scripts without label tweak derivation.",
+      );
+    }
```
macgyver13: What is the reason for this check? PSBT_OUT_SP_V0_INFO contains the labeled spend public key

packages/caravan-psbt/src/psbtv2/silentpayment.ts
```diff
@@ -0,0 +1,615 @@
...
+
+  assertValidDLEQProof(proof);
+
+  return {
+    ecdhShare,
+    proof,
+  };
+}
+
+export function verifySilentPaymentDLEQProof({
```
macgyver13, line 617: nit: add documentation or flag as internal to avoid direct consumers

packages/caravan-psbt/src/psbtv2/silentpayment.ts
```diff
@@ -0,0 +1,615 @@
...
+  const Pk = secp256k1.ProjectivePoint.fromHex(bspend).add(
+    secp256k1.ProjectivePoint.BASE.multiply(tweakScalar),
+  );
+
+  // P2TR output script: OP_1 <0x20> <x-only pubkey>
+  const xOnlyPk = Pk.toRawBytes(true).subarray(1);
+  return Buffer.concat([Buffer.from([OPS.OP_1, 0x20]), xOnlyPk]);
+}
+
+export function generateSilentPaymentDLEQProof({
```
macgyver13, line 579: nit: add documentation or flag as internal to avoid direct consumers

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -534,6 +600,11 @@ export class PsbtV2 extends PsbtV2Maps {
       return false;
     }
 
+    // Per BIP375: if SP outputs are present, all must have a computed
+    // PSBT_OUT_SCRIPT before signing is permitted.
+    if (this.hasSilentPaymentOutputs && !this.hasAllSPOutputScripts) {
+      return false;
+    }
     return true;
```
macgyver13: These two checks are overlapping, This yields the same result:

`return this.hasAllSPOutputScripts;`

leading whitespace is missing on these lines as well

packages/caravan-psbt/src/psbtv2/silentpayment.ts
```diff
@@ -0,0 +1,615 @@
...
+export function assertValidLabel(m: number, outputIndex: number): void {
+  if (!Number.isInteger(m) || m < 0 || m > 0xffffffff) {
+    throw new Error(
+      `Silent payment label m is invalid at output ${outputIndex}: ${m}`,
+    );
+  }
+}
+
+export function assertValidOutputK(k: number, outputIndex: number): void {
+  if (!Number.isInteger(k) || k < 0 || k > 2323) {
```
macgyver13: BIP352 includes 2323 as failure

`if (!Number.isInteger(k) || k < 0 || k >= 2323) {`

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,6 +848,434 @@ export class PsbtV2 extends PsbtV2Maps {
...
+      if (!proof) {
+        throw Error(
+          `Global ECDH share missing PSBT_GLOBAL_SP_DLEQ for scan key ${bscanHex}.`,
+        );
+      }
+
+      const aggregatePubkey =
+        this.tryGetInputAggregatePubkey(contributingInputs);
+      if (!aggregatePubkey) {
+        return null;
```
macgyver13: You might consider handling error for global aggregate pub key in one place matching input handling. This would unify the two `const ecdhShare = this.resolveAndVerifyScanKeyEcdhShare` cases.

```
     if (!aggregatePubkey) {
        throw Error(`Global ECDH aggregate key missing for scan key ${bscanHex}`);
     }
```

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,6 +848,434 @@ export class PsbtV2 extends PsbtV2Maps {
...
+          `Invalid input ${i} silent payment DLEQ proof for scan key ${bscanHex}.`,
+        );
+      }
+
+      perInputBuffers.push(ecdhShare);
+    }
+
+    if (perInputBuffers.length === 0) {
+      return null;
+    }
```
macgyver13: nit: probably unnecessary guard

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -773,6 +848,434 @@ export class PsbtV2 extends PsbtV2Maps {
...
+      if (!spInfo && !script) {
+        throw Error(
+          `PsbtV2 output ${i} is missing PSBT_OUT_SCRIPT and has no ` +
+            "PSBT_OUT_SP_V0_INFO. An output must have one or both per BIP375.",
+        );
+      }
+
+      if (spInfo && script && script.length === 0) {
+        throw Error(`PsbtV2 output ${i} has empty PSBT_OUT_SCRIPT.`);
+      }
```
macgyver13, line 1146: nit: reuse PSBT_OUT_SCRIPT helpers

```
      if (spInfo) {
        assertValidSPV0Info(spInfo, i);
      } else {
        continue;
      }

      if (outputMap.has(KeyType.PSBT_OUT_SCRIPT) && !this.hasNonEmptyOutputScript(i)) {
        throw Error(`PsbtV2 output ${i} has empty PSBT_OUT_SCRIPT.`);
      }
```

packages/caravan-psbt/src/psbtv2/psbtv2.ts
```diff
@@ -1016,12 +1525,45 @@ export class PsbtV2 extends PsbtV2Maps {
...
+
+      if (label !== undefined) {
+        if (!Number.isInteger(label) || label < 0 || label > 0xffffffff) {
+          throw Error(`label must be a uint32 (0–${0xffffffff}), got ${label}`);
+        }
+        const labelBw = new BufferWriter();
+        labelBw.writeU32(label);
+        map.set(KeyType.PSBT_OUT_SP_V0_LABEL, labelBw.render());
+      }
+    }
```
macgyver13, line 1574: deduplication: replace with setOutputSP helpers

```
    if (silentPayment) {
      const { bscan, bspend, label } = silentPayment;
      const outputIndex = this.outputMaps.length;
      this.setOutputSPInfo(map, bscan, bspend);

      if (label !== undefined) {
        this.setOutputSPLabel(map, outputIndex, label);
      }
    }
...
  private setOutputSPInfo(
    outputMap: Map<Key, Value>,
    bscan: Buffer,
    bspend: Buffer,
  ): void {
    assertValidBscan(bscan);
    assertValidBspend(bspend);
    const bw = new BufferWriter();
    bw.writeBytes(bscan);
    bw.writeBytes(bspend);
    outputMap.set(KeyType.PSBT_OUT_SP_V0_INFO, bw.render());
  }
...
  public addOutputSPInfo(
    outputIndex: number,
    bscan: Buffer,
    bspend: Buffer,
  ): void {
    if (outputIndex < 0 || outputIndex >= this.outputMaps.length) {
      throw new Error(
        `outputIndex ${outputIndex} out of range (${this.outputMaps.length} outputs)`,
      );
    }
    this.setOutputSPInfo(
      this.outputMaps[outputIndex],
      bscan,
      bspend,
    );
  }
```

macgyver13 (changes requested):
Appreciate you making the last round of changes.

Most comments are to improve readability you can decide if you want to apply them.

The two comments in psbtv2map are most important.

packages/caravan-psbt/src/psbtv2/dleq.test.ts
jvgelder: i blame prettier but fixed

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/silentpayment.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: yes great thanks replaced it

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: yep thanks fixed

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: I assumed PSBT_OUT_SP_V0_INFO was the base ? but i guess i am wrong here ?

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/silentpayment.ts
jvgelder: I stopped reading at:
```
_If any of the groups exceed the limit of Kmax (=2323) silent payment addresses, fail.[[18](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki#cite_note-why_limit_k_18)]_
```
but its explained futher indeed:

```
_Kmax = 2323 represents the maximum number of P2TR outputs that can fit into a 100kvB transaction, meaning a transaction that adheres to the current standardness rules is guaranteed to be within the limit. This ensures flexibility and also mitigates potential fingerprinting issues._
```

The log was correct however :P 

```js
  throw new Error(
      `Silent payment recipient index should be 0 < k < 2323 (Kmax) but was at output ${outputIndex}: ${k}`,
    );
```

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: its just so i can do a simple if later to check if i got any results instead of checking the length == 0 since `[]` is true in javascript.

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: added and fixed

packages/caravan-psbt/src/psbtv2/psbtv2.ts
macgyver13: > I assumed PSBT_OUT_SP_V0_INFO was the base ? but i guess i am wrong here ?

Pretty confident SP_V0_INFO must carry labeled spend public key - opened a BIP test vector improvement - this change is included in the vectors you used for this PR.
https://github.com/bitcoin/bips/pull/2207

jvgelder:
planning to fix the open points in the coming days

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
jvgelder: this should be addressed now let me know if the `buildPsbtIdentifierTx` and `getTransactionId` in `psbtv2Maps` what you meant

packages/caravan-psbt/src/psbtv2/psbtv2.ts
jvgelder: fixed

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
jvgelder: fixed

jvgelder:
were there  any remarks left besides a few nits ? @macgyver13 ?

PR #459, opened 2026-01-26
e2e: Add test for testing wallet name editing

Ferryx349:
**What kind of change does this PR introduce?**

- Test improvement / E2E test coverage

**Issue Number:**

Fixes #442 

**If relevant, did you update the documentation?**

- Not applicable (test-only change)

**Summary**

This PR adds an end-to-end test to cover the wallet name editing functionality in the Caravan coordinator.

Previously, editing and saving a wallet name was not covered by E2E tests. This change ensures that:
- A loaded wallet can enter edit mode
- The wallet name can be updated and saved
- The updated name is correctly reflected in the UI

The test follows the existing E2E test structure and error-handling conventions and runs as part of the full coordinator E2E suite.

**Does this PR introduce a breaking change?**

- No

## Checklist
- [x] I have tested my changes thoroughly.
- [x] I have added or updated tests to cover my changes (if applicable).
- [ ] I have verified that test coverage meets or exceeds 95% (if applicable).
- [x] I have run the test suite locally, and all tests pass.
- [x] I have written tests for all new changes/features
- [x] I have followed the project's coding style and conventions.
- [ ] I have created a changeset to document my changes (`npm run changeset`).

**Other information**

- The test handles Playwright strict mode by scoping locators where multiple elements share the same `data-cy` attributes.
- All coordinator E2E tests pass locally using Docker-based setup.

**Have you read the [contributing guide](https://github.com/caravan-bitcoin/caravan/blob/main/apps/coordinator/CONTRIBUTING.md)?**

- Yes

PR #508, opened 2026-05-19
test(e2e): cover wallet name editing

bucko13:
Adds a scoped e2e check for editing and saving the wallet name from the wallet info card.

Supersedes #459.


.changeset/purple-llamas-hang.md
jpcummins, line 1: ?

jpcummins (approved):

PR #511, opened 2026-05-21
feat(wallets): per-cosigner message signing across keystores

bucko13:
Adds `SignMessage({keystore, network?, bip32Path, message, expectedPubkey})` across Ledger v1+v2, Trezor, Jade, BitBox, and Coldcard. All keystores return a normalized `Entry = {bip32Path, signature, expectedPubkey}` and a single `verifyMessageSignature` (loose-mode bip322-js) handles the recovery side. 

This only implements BIP137 support as most devices still don't have any BIP322 support. I might put up a follow up that covers BIP322 for coldcard which recently added support. 

Also wires the new flow into `/#/test` as a Message Signing category over `TEST_FIXTURES.multisigs`.

Note: `@caravan/wallets` bumps to 1.0.0 due to the backwards incompatible changes to the existing interactions normalizing the return string. 

## What's left before merging

Need to validate in the test suite:
- [x] Ledger
- [x] Trezor
- [x] Jade
~~- [x] Bitbox~~ not actually supported
- [x] Coldcard

Not included:
- BCUR2 since it doesn't look like BCUR defines a registry type for message signing and implementation might require more research and bigger scope, I punted for now
- BIP322 support: there's not a lot of support yet so didn't seem like a priority for first pass
- Bitbox: only seems to support message signing for non multisig paths. would need to talk to bitbox team about being more flexible for message signing paths so that you can validate keys from a multisig wallet using the BIP137 standard (which is only P2WPKH)


packages/caravan-wallets/package.json
bucko13, line 1: need to confirm these packages don't need to be cleaned up.

apps/coordinator/src/tests/messageSigning.jsx
```diff
@@ -0,0 +1,133 @@
...
+class MessageSigningTest extends Test {
+  name() {
+    return `Sign message — ${this.params.network} ${this.params.addressType} (open_source cosigner)`;
+  }
+
+  description() {
+    return (
+      <Box>
+        <p>
+          Sign a fixed UTF-8 message with the <strong>open_source</strong>{" "}
```
bucko13: too verbose. don't need this much explanation for the test suite

packages/caravan-bitcoin/src/fixtures.ts
```diff
@@ -93,6 +93,12 @@ const BIP39_PHRASE = [
 
 export const ROOT_FINGERPRINT = "f57ec65d";
 
+// Common message body for the pre-computed BIP-137 / BIP-322 signatures
```
bucko13: don't need to mention BIP-322 here. BIP322 can do more complex message signing and we might do something different for those tests (like proof of reserves)

packages/caravan-wallets/src/bitbox.test.ts
```diff
@@ -0,0 +1,128 @@
+import { Network } from "@caravan/bitcoin";
+import { MessageSigningError } from "@caravan/messages";
+import { describe, it, expect, vi } from "vitest";
```
bucko13: these should be available as globals.

packages/caravan-wallets/src/ledger.ts
```diff
@@ -37,6 +37,7 @@ import {
   PsbtV2,
   ExtendedPublicKey,
 } from "@caravan/bitcoin";
+import type { Entry } from "@caravan/messages";
```
bucko13: assuming this will need to be made available to consumers of the wallets package, we probably want to export this from wallets too (since messages is a private package). Maybe title it something a little more obvious like `MessageSignatureResult` or something (probably could workshop that)

packages/caravan-wallets/src/jade.ts
```diff
@@ -464,32 +470,120 @@ export class JadeSignMultisigTransaction extends JadeInteraction {
...
-  async run() {
+
+  async run(): Promise<Entry> {
     return await this.withDevice(async (jade: IJade) => {
       const path = bip32PathToSequence(this.bip32Path);
-      return await jade.signMessage(path, this.message);
+      let rawSig: Uint8Array;
+      try {
+        // We do not pass useAeSignatures, so the SDK returns a plain
+        // Uint8Array (not the [sig, hostCommitment] tuple used for anti-exfil).
```
bucko13: should we useAeSignatures?

packages/caravan-wallets/src/bitbox.test.ts
```diff
@@ -0,0 +1,140 @@
...
+const PATH = FIXTURE.bip32Path;
+const MESSAGE = FIXTURE.signedMessages.message;
+// Real BIP-137 sig (88-char base64) decoded to its 65-byte wire form,
+// which is exactly what BitBox returns as `electrumSig65`.
+const REAL_ELECTRUM_SIG65 = new Uint8Array(
+  Buffer.from(FIXTURE.signedMessages.bip137, "base64"),
+);
+const REAL_SIG_BASE64 = FIXTURE.signedMessages.bip137;
+
+function makeInteraction(opts: Partial<{
```
bucko13: if these will always stay sing message specific, then move these into the scope of those tests. if these are meant to be more general, then make it more of a factory style

bucko13 (commented):
some things to follow up on.

bucko13:
Running manual tests, Ledger: 
Screenshot 2026-05-21 at 3 41 35 PM

bucko13:
Jade passed (after some fixes):
Screenshot 2026-05-21 at 3 54 02 PM

bucko13:
Trezor (mostly) passed. Note that P2SH signing fails, but it also fails on normal transaction signing due to a forbidden key path error. 
Screenshot 2026-05-21 at 4 08 58 PM

bucko13:
And confirmed with coldcard (also needed a fix), this was MK3 only: 

Screenshot 2026-05-21 at 4 34 54 PM

PR #532, opened 2026-09-08
fix(wallets): handle Jade USB timeouts and multipart sign_psbt replies

jbrauck-unchained:
## Summary
- `jadets` times out `getReceiveAddress` after 5 seconds. Official `jadepy` waits 120s. Classic devices often need the extra time to confirm a change address.
- Multipart `sign_psbt` replies are 1-based. `jadets` skips the last fragment and omits `origid` / `orig` / `seqlen` on `get_extended_data`, so Jade stays on Processing and the host never gets a signature.
- After opening serial, clear DTR/RTS the same way `jadepy` does. Leaving them set can reboot Jade Classic.

Caravan now wraps serial + RPC instead of forking `jadets`. Unchained will pick this up after `@caravan/wallets` is published and bumped.

## Test plan
- [ ] Confirm a receive/change address on Jade Classic over USB. It should not time out at 5 seconds.
- [ ] Sign a PSBT over USB on Jade Classic. Jade should leave Processing and return a signed PSBT.
- [ ] Open the serial connection and confirm the device does not reboot.
- [ ] Run `packages/caravan-wallets` Jade unit tests (`jadeRpc`, `jadeSerial`, existing `jade` tests).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/jade-usb-protocol.md
+++ b/.changeset/jade-usb-protocol.md
@@ -1,0 +1,5 @@
+---
+"@caravan/wallets": patch
+---
+
+Fix Jade USB address confirmation timeouts, multipart sign_psbt replies, and serial DTR/RTS handling so Classic devices can finish signing without hanging.
--- a/.changeset/tired-knives-own.md
+++ b/.changeset/tired-knives-own.md
@@ -1,0 +1,5 @@
+---
+"@caravan/wallets": patch
+---
+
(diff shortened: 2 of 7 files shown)


jbrauck-unchained:
## Before
Screenshot 2026-09-08 at 4 39 23 PM

## After
Screenshot 2026-09-08 at 4 39 06 PM

jbrauck-unchained:
Moving this patch over to https://github.com/Austin-Fulbright/jadets/pull/5

PR #533, opened 2026-09-09
fix: encode PSBTv2 counts as CompactSize

jpcummins:

Bug fix and security hardening.

**Snapshots/Videos:**

Not applicable. This change affects PSBT parsing and serialization only.

Not applicable. No public API or documentation changes are required.

**Summary**

This PR fixes incorrect handling of PSBTv2 input and output counts. BIP-370 defines these fields as CompactSize integers, but they were previously read and written as single-byte values. Transactions with 256 or more outputs could therefore be serialized with a truncated output count, causing undeclared output maps to be discarded during parsing and their value to become transaction fees.

No public API changes are introduced.

Malformed, noncanonical PSBTv2 files produced by the previous implementation with truncated counts will now be rejected rather than silently interpreted as transactions with fewer inputs or outputs.

## Checklist

**Other information**

All tests pass

Yes


packages/caravan-psbt/src/psbtv2/functions.ts
```diff
@@ -98,19 +98,29 @@ export function parseDerivationPathNodesToBytes(path: string): Buffer {
...
 
-  const keyLen = nextByte.readUInt8(0);
   const key = br.readBytes(keyLen);
   const value = br.readVarBytes();
 
   map.set(key.toString("hex"), value);
   readAndSetKeyPairs(map, br);
 }
 
+export function readCompactSize(value: Buffer): number {
```
Shadouts, line 113: This function is duplicated in the `bitcoin` package. It can just be imported from there.

packages/caravan-psbt/src/psbtv2/functions.ts
```diff
@@ -98,19 +98,29 @@ export function parseDerivationPathNodesToBytes(path: string): Buffer {
...
+  const br = new BufferReader(value);
+  const result = br.readVarint();
```
Shadouts, line 115: I was not aware that this method was compatible with Core's definition of CompactSize. Convenient.

packages/caravan-bitcoin/src/psbtv2.ts
Shadouts, line 1: I'm concerned with why this module exists at all. I think it might have been a mistake during the migration to our new repo, or it was left for backward compatibility reasons.

This is an older version of our current `PsbtV2` class, but they no longer match because this has not been maintained with new features.

I think it's fine to add this particular fix, but we need a note added to the top comment indicating that this module is deprecated in favor of the `@caravan/psbt` package.

packages/caravan-psbt/src/psbtv2/psbtv2.test.ts
```diff
@@ -903,6 +903,62 @@ describe("PsbtV2", () => {
     expect(psbt.PSBT_GLOBAL_OUTPUT_COUNT).toBe(0);
     expect(psbt.PSBT_GLOBAL_FALLBACK_LOCKTIME).toBe(0);
   });
+
+  test.each([252, 253, 256])(
+    "Round trips a PSBTv2 with %i outputs",
```
Shadouts, line 908: This is a confusing description for the test.

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
```diff
@@ -199,18 +205,13 @@ export class PsbtConversionMaps extends PsbtV2Maps {
...
         .get(KeyType.PSBT_GLOBAL_FALLBACK_LOCKTIME)
         ?.readUInt32LE() || 0;
 
-    const numInputs =
-      this.globalMap.get(KeyType.PSBT_GLOBAL_INPUT_COUNT)?.readUInt8() || 0;
-    const numOutputs =
-      this.globalMap.get(KeyType.PSBT_GLOBAL_OUTPUT_COUNT)?.readUInt8() || 0;
-
-    for (let i = 0; i < numInputs; i++) {
+    for (let i = 0; i < this.inputMaps.length; i++) {
```
Shadouts, line 214: This and the one below concerns me as it's a paradigm shift in how to handle a mismatch of mapped values to defined number. There was another piece of code in this change that also deals with mismatch.

Shadouts (changes requested):
Primary request: Let's make `readCompactSize` a first-class utility exported by the `@caravan/bitcoin` package in the `utils` module. Then it can be imported from everywhere.

Secondarily, I'm concerned with the shift in how a mismatch in `PSBT_GLOBAL_X_COUNT` from the actual number of map items is handled. Previously, if global count value differed, then extra mapped items were dropped during unsigned txn creation. Now, global count value is ignored and mapped items iterated.

I see the global count value as truth because the input and output maps are tolerant of arbitrary data. If it's not truth, what's its purpose? I know there was also a check to try and ensure mismatches don't happen, but I'm not sure of how that will interact when combining PSBTs.

I think this topic deserves further discussion, but can that part of the change be omitted or kept separate from simply adding the readCompactSize functionality?

packages/caravan-psbt/src/psbtv2/psbtv2maps.ts
jpcummins: Good point. Reverted this so that `buildUnsignedTx()` keeps the global counts as authoritative.

jack-unchained (dismissed):
.

Shadouts (dismissed):
Approved.

To get the ci workflows to pass, you can wait for https://github.com/caravan-bitcoin/caravan/pull/534 to merge or perform a like fix on this branch.

package.json
```diff
@@ -38,6 +38,7 @@
     "@typescript-eslint/eslint-plugin": "^7.18.0",
     "@vitest/coverage-istanbul": "^3.2.4",
     "@vitest/ui": "^3.2.4",
+    "bitcoinjs-lib": "5.2.0",
```
jack-unchained: It's strange that this is in devDependencies, but it seems to work to unblock CI and I don't understand why. Because I don't understand it, it concerns me about the downstream effect and how it might cause some debugging confusion. We already juggle some bitcoinjs-lib overrides between v5 and v6.

If the goal is to prevent bumps to v5.2.1, I don't think it would prevent someone from upgrading the lockfile in one of the packages like caravan-wallets which uses bitcoinjs-lib v5. My recommendation is to pin the specific dependency entries per package and see if that works. Packages using v5 are:

- coordinator
- caravan-bitcoin
- caravan-wallets

Shadouts (approved):

package.json
jpcummins: Updating the root dev dependency is the cleanest way to fix CI dependency issues without changing workspace dependency resolution.

